PR #1, opened 2017-11-13
[clarify] Fill BUILDING_FILES with cmake fs rules

dalerank:
+ we can fill BUILDING_FILES with using cmake


bvschaik:
The cmake folks recommend against this:

https://cmake.org/cmake/help/v3.0/command/file.html
"We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate."

dalerank:
yep, project have a 100500 files in project ))) less code, less erros
close it if need

PR #2, opened 2017-11-13
another realization for namespace alternate

dalerank:


bvschaik:
Closing this: the code as-is doesn't compile, though I will use this approach for the GUI parts later on.

PR #3, opened 2017-11-13
Update CityInfo.c

dalerank:


bvschaik:
I assume this has to be combined with PR #2 to make it actually work, right?

dalerank:
I do it from mobile ;)

bvschaik:
Closing this: the code as-is doesn't compile, though I will use this approach for the GUI parts later on.

PR #5, opened 2017-12-03
better implement game.messages subsystem using struct for functions i…

sleepstranger:
…ncapsulating

### DIFF
--- a/src/Building.c
+++ b/src/Building.c
@@ -384,7 +384,7 @@ void Building_collapseLastPlaced()
 		}
 	}
 	if (buildingId) {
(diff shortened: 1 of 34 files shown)


bvschaik:
Thanks for your interest in Julius. However, I'm currently already in a major refactor of the codebase, and I'm not going to introduce a third way of working.

devnoname120:
Would it make more sense to switch to C++ than use function pointers to simulate OOP?

bvschaik:
Yes. But only once I'm through the current phase of splitting stuff up in subdirectories and getting rid of the global variables in src/Data.

PR #11, opened 2017-12-13
Notify on console if game saving failed

nwtour:
If savefile has no write rights or unable free space on disk, game looks like work, but real save broken.


bvschaik:
Please make clean pull requests, with only the necessary commits, rather than everything including merges from my repo.

PR #16, opened 2017-12-29
Windows support through MinGW

devnoname120:
All the tests run fine.

The only issue I encountered in my limited testing is that Julius Caesar's voice (the one you get when you run out of funds) is sped up and higher pitch. Is that a Windows-only bug? Edit: Yes, I don't have this bug under Linux.


devnoname120:
It would be nice to add MinGW as well to Travis, tests wouldn't be able to run though.

bvschaik:
I've created an AppVeyor account to be able to build on Windows as well. Building under Visual Studio is going to be another challenge...

The audio: is that only during the out-of-funds video or all videos? Is normal audio OK, for example the mission briefings? It might be related to the smacker library I'm using. On Linux and macOS the audio runs OK.

devnoname120:
Yes, mission briefings are OK.

PR #21, opened 2018-02-12
Visual C++ Compile Fix

polaretto:
I found this when trying to compile following your instructions. 

I still get a LNK1107	invalid or corrupt file: cannot read at 0x368	C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CMakeLists.txt	...path...to...sdl2...\SDL2\lib\x64\SDL2.dll	1

Do you have any idea about where the error could be?

### DIFF
--- a/test/sav/run.c
+++ b/test/sav/run.c
@@ -23,19 +23,19 @@
 static void handler(int sig)
 {
 #if defined(__GNUC__) && !defined(__MINGW32__)
-	void *array[100];
-	size_t size;
+    void *array[100];
+    size_t size;
 
-	// get void*'s for all entries on the stack
-	size = backtrace(array, 100);
+    // get void*'s for all entries on the stack
+    size = backtrace(array, 100);
 
-	// print out all the frames to stderr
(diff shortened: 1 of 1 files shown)


bvschaik:
Ah yes, I only tried the julius.exe target, not the others.

Are you compiling for x86 or x64? So far I have only tried x86.

polaretto:
Ok, Julius works in x86. The problem seems to appear for the others and also for Julius in x64.

polaretto:
Thanks, I am used to get alerts when this happens, don't know why nothing showed up. Should be ok now, let me know if there are any problems :)

bvschaik:
Probably because I haven't set up anything to check the code... Do you have an example of a github project where you were alerted of style issues?

polaretto:
No that's not your fault! Apparently, this was done by a everyone-has extension in VS by Microsoft, which I didn't install since I switched to VS2017 ([Productivity Power Tools](https://marketplace.visualstudio.com/items?itemName=VisualStudioProductTeam.ProductivityPowerPack2017#overview)).

PR #24, opened 2018-02-13
Fixed bridge/fort demolition bug

polaretto:
Fixed a bug that made user earn money by changing his mind about demolishing a bridge (or fort).

### DIFF
--- a/src/building/construction.c
+++ b/src/building/construction.c
@@ -607,13 +607,13 @@ void building_construction_place(int orientation)
 
     int placement_cost = model_get_building(type)->cost;
     if (type == BUILDING_CLEAR_LAND) {
-        // BUG: if confirmation has to be asked (bridge/fort), the previous cost is deducted from treasury
-        // and if user chooses 'no', they still pay for removal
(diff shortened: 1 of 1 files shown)


bvschaik:
Could you explain what happened, how the user could earn money and what you did to fix it?

And could you place the `else` on the same line as the closing brace for the `if`?

polaretto:
Sure!
The problem was that items_placed stores as value the number of items deleted by the "Clear Land" button.
Some buildings (Forts and Bridges) ask for confirmation of the removal action. Unfortunately, since building_construction_clear_land() returns -1 if the user doesn't confirm, the original items_placed variable became negative, resulting in a change of sign in the amount to pay.
Since the default amount to pay for removing a bridge is 2, it resulted in a payment of -2 -> the player earned 2 Dn.

polaretto:
Updated as requested :)

bvschaik:
With your changes, the user doesn't pay at all for the removal of a bridge/fort.

The bug was in the `if >= 0` if statement, that should be `if < 0`. Good catch though!

PR #42, opened 2019-01-09
Change the way window invalidation works. Fixes bug with victory screen not displaying.

crudelios:
There are now two types of window invalidation:
- An immediate invalidation, which prevents any further ticks from occuring until the screen is refreshed;
- An invalidation "on draw", which allows the remaining game ticks to be processed before redrawing the screen (the way window invalidation worked before this commit).

Reasoning for the change:

With the old code, there was a bug with window drawing when game speed was set to >100%.
Specifically, the victory screen did not appear and instead the victory video showed immediately (with no chance to continue governing).

The bug was due to the fact that when victory conditions were met, a value that the victory popup was shown was set to true (`city_data.mission.victory_message_shown` was set to `1`).
After that, the screen was invalidated to draw the victory dialog.
However, since when game speed is set to >100%, the game processes more than one tick per frame, it kept running the city and went back to the victory condiction routine (`city_victory_check` in `victory.c`) and, as the popup was set to already having been shown, the game loaded the victory video without ever loading the victory popup.

By setting two different possible timings to a redraw request in `game_run` (either after the current tick was processed or after all the remaning ticks for the frame have been processed), the game is able to immediately show popup windows and prevent the city to keep running in the background until the next frame was shown.

Also, certain elements, like tooltips and the sidebar scrolling, do not cause the game to pause like most windows do, therefore they do not need to interrupt the ticks when speed is over 100%. Doing so would reduce the speed to 1 tick per frame (effectively 100% speed) when a tooltip was shown or the sidebar scrolled.
Therefore, those elements are only redrawn after all ticks have been processed, behaving like all the windows did before this commit.

This may also prevent all sorts of different bugs with game speed over 100%, for example if two message dialogs arrived in different ticks of the same frame, one of them might be ignored.

### DIFF
--- a/src/game/game.c
+++ b/src/game/game.c
@@ -124,6 +124,10 @@ void game_run(void)
     for (int i = 0; i < num_ticks; i++) {
         game_tick_run();
         game_file_write_mission_saved_game();
+
+        if (window_must_refresh()) {
+            break;
+        }
     }
 }
 
--- a/src/graphics/tooltip.c
+++ b/src/graphics/tooltip.c
@@ -44,7 +44,7 @@ static void reset_tooltip(tooltip_context *c)
 {
     if (c->type != TOOLTIP_NONE) {
         c->type = TOOLTIP_NONE;
-        window_invalidate();
+        window_request_refresh_on_draw();
     }
 }
 
--- a/src/graphics/window.c
+++ b/src/graphics/window.c
@@ -9,7 +9,8 @@
 static window_type window_queue[MAX_QUEUE];
 static int queue_index = 0;
 static window_type *current_window = 0;
-static int refresh_requested;
+static int refresh_immediate;
+static int refresh_on_draw;
 
 static void noop(void)
 {
@@ -36,7 +37,18 @@ static void decrease_queue_index(void)
 
 void window_invalidate(void)
 {
-    refresh_requested = 1;
+    refresh_immediate = 1;
+    refresh_on_draw = 1;
+}
+
+int window_must_refresh(void)
+{
+    return refresh_immediate;
+}
+
+void window_request_refresh_on_draw(void)
+{
+    refresh_on_draw = 1;
 }
 
 int window_is(window_id id)
@@ -87,9 +99,10 @@ static void update_mouse_after(void)
 void window_draw(int force)
 {
     update_mouse_before();
-    if (force || refresh_requested) {
+    if (force || refresh_on_draw) {
         current_window->draw_background();
-        refresh_requested = 0;
+        refresh_on_draw = 0;
+        refresh_immediate = 0;
     }
     current_window->draw_foreground();
 
--- a/src/graphics/window.h
+++ b/src/graphics/window.h
@@ -57,6 +57,10 @@ typedef struct {
 
 void window_invalidate(void);
 
+void window_request_refresh_on_draw(void);
+
+int window_must_refresh(void);
+
 void window_draw(int force);
 
 int window_is(window_id id);
--- a/src/widget/sidebar.c
+++ b/src/widget/sidebar.c
@@ -428,7 +428,7 @@ static void update_progress(void)
 
 static void draw_sliding_foreground(void)
 {
-    window_invalidate();
+    window_request_refresh_on_draw();
     update_progress();
     if (data.progress >= 47) {
         city_view_toggle_sidebar();
--- a/test/stub/ui.c
+++ b/test/stub/ui.c
@@ -29,6 +29,11 @@ window_id window_get_id(void)
     return WINDOW_CITY;
 }
 
+int window_must_refresh(void)
+{
+    return 0;
+}
+
 void window_draw(int force)
 {}
 


AzureDragon-1:
Hey.Can anyone tell me exactly how to make changes to my directory respective file so that I also can resolve this problem. Where to download the fixed file from ? Where to put the downloaded file ?

bvschaik:
@AzureDragon-1 I gave you instructions in your other comment. If you won the map and the victory *video* played, that's this bug. If you think you won the map but the game just played on, that's something else (and please report it if you think it's a bug with the game).

AzureDragon-1:
@bvschaik I met the criteria for victory AND there was NO VICTORY MESSAGE. I mean that message which gives you the option to accept promotion or continue governing. Instead of this message all of a sudden when I met the victory criteria ON A RANDOM MAP I was playing, the victory video started ! That's why I think it's not a bug from the original game-it;'s the problem with the mod.
By the way, @bvschaik I still don't understand something : when you or crudelios make a changer. or a fix to the mod, threads say that you MERGE the files or whatever got changed. What does this exactly mean ? According to my understanding you make a change in the Julius-windows 1.0.0 zip file-is that correct ? OR NOT ? Does it mean that when a fix is done, all I need to do is to download the zip file from the page you send me link to and unzip it again in the game folder choosing these files to be overwirtten ? 
Are you going anytime soon to make a modifies version of the mod in an exe file with ALL these changes and fixes implemented in it ?

AzureDragon-1:
@bvschaik 
 I met the criteria for victory AND there was NO VICTORY MESSAGE. I mean
 that message which gives you the option to accept promotion or continue
 governing. Instead of this message all of a sudden when I met the 
victory criteria ON A RANDOM MAP I was playing, the victory video 
started ! That's why I think it's not a bug from the original game-it;'s
 the problem with the mod.

By the way,  @bvschaik 
 I still don't understand something : when you or crudelios make a 
change or a fix to the mod, threads say that you MERGE the files or 
whatever got changed. What does this exactly mean ? According to my 
understanding you make a change in the Julius-windows 1.0.0 zip file-is 
that correct ? OR NOT ? Does it mean that when a fix is done, all I need
 to do is to download the zip file from the page you send me link to and
 unzip it again in the game folder choosing these files to be 
overwirtten ?
 
Are you going anytime soon to make a modifies version of the mod in an 
exe file with ALL these changes and fixes implemented in it ?  
  I just can't understand the interface of the site you are using. It's not very friendly to me. I still do not know which page to download the updated files from ?
 







 >-------- Оригинално писмо --------

 >От: Bianca van Schaik notifications@github.com

 >Относно: Re: [bvschaik/julius] Change the way window invalidation works. Fixes
 bug with victory screen not displaying. (#42)

 >До: bvschaik/julius  

 >Изпратено на: 28.01.2019 22:03



 
 
  
Merged  #42  into master.

crudelios:
@AzureDragon-1:

The bug you had is what this change fixes.

Regarding your questions about how to get the most recent version of the program and this site not being intuitive, you need to understand that this site is intended primarily for developing, that's why it's not user friendly. I could arguably say that GitHub isn't even easy at first for developers, because there are many rules to follow in order to do everything properly.

Let me try to explain how GitHub works: This site deals with the code and not with the actual created programs. Taking that into account, the expression "merged" means that the code changes made by me in this pull request (the fix for the victory screen) were, as to say, "inserted" to the main source code available on this site.

Now, even though the "main" source code has been updated, that doesn't mean the zip file you can download from this site has the new changes. That zip file is from a specific time, when the code didn't have the new changes.

There are however ways to get the most recent version of the game, that includes all the "merged" changes in the past few days, because there is a separate site that automatically creates the program when code changes are detected. That site is Appveyor, and I can say it's even less intuitive for end users than GitHub.

So here are the steps you can take to get the most recent version of Julius:

- Go to the followng link: https://ci.appveyor.com/project/bvschaik/julius
- There is a link that says "Jobs" with a table below it. Click the bottommost row of that table;
- Ignore all the text that appears and select "Artifacts". The link is right over the wall of text;
- Select "julius.zip";
- Now extract all the files to the Caesar 3 directory, just like you did before!

AzureDragon-1:
Heyq thank you very much for the detailed explanation. I really appreciated your time. That was exactly what I needed. Everything works fine now ! 
  I have a question : Does it always have to be the bottommost row of the table ?  When I clicked on the link there were 3 rows. The first and the second were more recently updated. Should I download them or ONLY the last one is the right for me ? 
 






 >-------- Оригинално писмо --------

 >От: crudelios notifications@github.com

 >Относно: Re: [bvschaik/julius] Change the way window invalidation works. Fixes
 bug with victory screen not displaying. (#42)

 >До: bvschaik/julius  

 >Изпратено на: 29.01.2019 00:38



 
 
  
 @AzureDragon-1 :  
  
The bug you had is what this change fixes.  
  
Regarding your questions about how to get the most recent version of the program and this site not being intuitive, you need to understand that this site is intended primarily for developing, that's why it's not user friendly. I could arguably say that GitHub isn't even easy at first for developers, because there are many rules to follow in order to do everything properly.  
  
Let me try to explain how GitHub works: This site deals with the code and not with the actual created programs. Taking that into account, the expression "merged" means that the code changes made by me in this pull request (the fix for the victory screen) were, as to say, "inserted" to the main source code available on this site.  
  
Now, even though the "main" source code has been updated, that doesn't mean the zip file you can download from this site has the new changes. That zip file is from a specific time, when the code didn't have the new changes.  
  
There are however ways to get the most recent version of the game, that includes all the "merged" changes in the past few days, because there is a separate site that automatically creates the program when code changes are detected. That site is Appveyor, and I can say it's even less intuitive for end users than GitHub.  
  
So here are the steps you can take to get the most recent version of Julius:  
    
    Go to the followng link:  https://ci.appveyor.com/project/bvschaik/julius   
    There is a link that says "Jobs" with a table below it. Click the bottommost row of that table;  
    Ignore all the text that appears and select "Artifacts". The link is right over the wall of text;  
    Select "julius.zip";  
    Now extract all the files to the Caesar 3 directory, just like you did before!

crudelios:
You can download the other ones, but the those require the updated MSVC runtimes.
You probably already have those installed, so there is no harm in trying. At worst the game will complain about a missing dll and not run, which you can fix by downloading from the last row or by installing the MSVC runtimes (google it if needed, the first results will point you where you need to go).
An important warning though, if you get a build from a different row, always copy all three files (julius, SDL2 and SDL_mixer), not only the julius executable!

AzureDragon-1:
OK. But does it still mean that the FINAL version of the FIX will always be in the bottom file ? 
  So after a few weeks, if I want to get the latest update, I should only download the file at the bottom and it would be enough, right ? 
 







 >-------- Оригинално писмо --------

 >От: crudelios notifications@github.com

 >Относно: Re: [bvschaik/julius] Change the way window invalidation works. Fixes
 bug with victory screen not displaying. (#42)

 >До: bvschaik/julius  

 >Изпратено на: 29.01.2019 17:47



 
 
  
You can download the other ones, but the those require the updated MSVC runtimes.
 You probably already have those installed, so there is no harm in trying. At worst the game will complain about a missing dll and not run, which you can fix by downloading from the last row or by installing the MSVC runtimes (google it if needed, the first results will point you where you need to go).
 An important warning though, if you get a build from a different row, always copy all three files (julius, SDL2 and SDL_mixer), not only the julius executable!

crudelios:
If all goes as planned a link will soon be added to the main page of the Julius project that downloads the latest update without having to go look for it at Appveyor. It will be listed under "Downloading latest Windows build".

AzureDragon-1:
OK. Thank you. We shall wait for these regular updates of the mod. 







 >-------- Оригинално писмо --------

 >От: crudelios notifications@github.com

 >Относно: Re: [bvschaik/julius] Change the way window invalidation works. Fixes
 bug with victory screen not displaying. (#42)

 >До: bvschaik/julius  

 >Изпратено на: 29.01.2019 19:54



 
 
  
If all goes as planned a link will soon be added to the main page of the Julius project that downloads the latest update without having to go look for it at Appveyor. It will be listed under "Downloading latest Windows build".

PR #44, opened 2019-01-10
Fixed a typo in the code which caused Grand Insulaes to not draw properly.

crudelios:
This happened only when they evolved or devolved from Large Insulaes/Small Villas, but not when they were loaded from a savegame.

Playing through the missions is helping me find all these bugs. More fixes incoming! (3x3 merging is causing many bugs as well)


crudelios:
Fixed the problem with the 3x3 merging, it was a typo as well!

PR #45, opened 2019-01-18
Improve game save/load/delete dialog.

crudelios:
- Fix a crash when long (>100 chars) file names are used;
- Prevent file name's sizes from overflowing the file list window (the text input with the selected file still overflows, though).

### DIFF
--- a/src/graphics/text.c
+++ b/src/graphics/text.c
@@ -5,6 +5,8 @@
 #include "graphics/graphics.h"
 #include "graphics/image.h"
 
+#include <string.h>
+
 static uint8_t tmp_line[200];
 
 static struct {
@@ -18,6 +20,19 @@ static struct {
     int x_offset;
     int y_offset;
 } input_cursor;
+
+static struct {
+    const char * string;
+    int width[FONT_TYPES_MAX];
+} ellipsis = { "...", { 0 } };
+
+static int get_ellipsis_width(font_t font)
+{
+    if (!ellipsis.width[font]) {
+        ellipsis.width[font] = text_get_width(string_to_ascii(ellipsis.string), font);
+    }
+    return ellipsis.width[font];
+}
 
 void text_capture_cursor(int cursor_position)
 {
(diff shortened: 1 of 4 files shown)


src/graphics/text.c
```diff
@@ -1,3 +1,5 @@
+#include <string.h>
```
bvschaik: Please add system includes below the project includes.

src/graphics/text.c
```diff
@@ -84,6 +86,40 @@ int text_get_width(const uint8_t *str, font_t font)
     return width;
 }
 
+void text_fit_to_width(char *str, font_t font, int requested_width, const char *end_replacement)
```
bvschaik: Maybe rename this to "text_ellipsize" and skip the `end_replacement` parameter?

src/window/file_dialog.c
```diff
@@ -47,6 +47,8 @@ static generic_button file_buttons[] = {
 
 #define NOT_EXIST_MESSAGE_TIMEOUT 500
 
+static const int MAX_FILE_WINDOW_TEXT_WIDTH = 18 * 16;
```
bvschaik: `static const int` vs. `#define`: which is better?

src/graphics/text.c
```diff
@@ -84,6 +86,40 @@ int text_get_width(const uint8_t *str, font_t font)
...
+{
+    char *orig_str = str;
+    const font_definition *def = font_definition_for(font);
+    int end_replacement_width = (end_replacement) ? text_get_width((const uint8_t *)end_replacement, font) : 0;
```
bvschaik: Use `string_from_ascii()` instead of a cast.

src/graphics/text.c
```diff
@@ -84,6 +86,40 @@ int text_get_width(const uint8_t *str, font_t font)
...
+            length_with_replacement++;
+        }
+        if (width > requested_width) {
+            break;
+        }
+        str++;
+        maxlen--;
+    }
+    if (10000 - maxlen < strlen(orig_str))
+    {
```
bvschaik: Curly brace on the same line as the `if`, please :)

src/window/file_dialog.c
crudelios: I'm used to C++ and was brainwashed to believe a `static const` is always better!

I will change to `define` if you prefer though.

src/graphics/text.c
crudelios: Good idea, that will also improve performance by only having to check the elipsis' width once per font type.

src/graphics/text.c
crudelios: Will do!

src/window/file_dialog.c
bvschaik: I don't have a clear preference (too little in-depth experience with C), but seeing a define and a static const next to each other is a bit weird. I'm leaning towards the `static const` though, because of type safety.

src/graphics/text.c
crudelios: Old habits die hard!!! Your coding style is more densely packed than mine and it takes actual effort to keep things consistant.

src/window/file_dialog.c
crudelios: That's my reasoning, `static const` has type safety whereas `define` doesn't. But I get your point, I'll change both to `static const` to prevent OCD-induced seizures.

src/graphics/text.c
```diff
@@ -21,6 +21,19 @@ static struct {
     int y_offset;
 } input_cursor;
 
+static struct {
+    const char * string;
+    int width[FONT_TYPES_MAX];
+} ellipsis = { "...", { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 } };
```
bvschaik: You can omit the array of zeroes: when you initialize `string`, the rest of the struct's variables is set to 0.

crudelios: I didn't know about that.

For readability's sake, how about `ellipsis = { "...", { 0 } };` ?

crudelios: Changed as requested, also rebased to prevent many commits. Kept `{ "...", { 0 } }` just to make things clearer.

Issue #49, opened 2019-01-26
Bridge reconstruction


iulianiliescu:
You can't place a new bridge where one has been deleted already.

1. Build a bridge
2. Delete it
3. Attempt to build a new bridge in the same location
4. The square where the game will attempt to put the other end of the bridge will be the middle of the water.

Between the docks and the reservoir, near the lumber industry, you will see one graden on each side of the river, that is the place of a former bridge, which I can't rebuild anymore.

[savegame.zip](https://github.com/bvschaik/julius/files/2798813/savegame.zip)

iulianiliescu:
Also the fact when you demolish a bridge, it only demolishes the portion you clicked on, it's kind of annoying.

crudelios:
It's indeed a bug, which happens only when the bridge itself (excluding its ramps) is one tile wide.

I'll try to fix it when I have time.

siredmar:
Fixed the bug. See pull request #50

crudelios:
Siredmar I'm afraid I don't agree  with the proposed bugfix. Treating all road tiles with sprites as bridges doesn't seem correct to me.

Contrary to what I said before, the bug happens on all ship bridges, regardless of its length. Also, it doesn't happen on the original Caesar 3.

Upon some investigation, I realized the bug happens because some tiles of the bridge are being assigned sprite id 0 (which means no sprite). More specifically, in lines 114 to 118 of `map\bridge.c`, in function `map_bridge_get_sprite_id`, there is the following code:

```
if (is_ship_bridge) {
    if (index == 1 || index == length - 2) {
        // platform after ramp
        return 0;
}
```

This happens because the first and last tiles of a ship bridge are set to the "bridge platform" sprite, and its image occupies two tiles. Therefore, no sprite needs to be drawn on the tiles `bridge start + 1` and `bridge end - 1`.

However, function `map_is_bridge` only returns true if the tile in question is a water tile and there is a sprite in that tile (meaning, the sprite id is other than 0). Because there are no sprites in those tiles, the function returns false and other functions that depend on it (such as the bridge demolition routine) get sent incorrect bridge length sizes.

[...]

PR #50, opened 2019-01-26
Fixed issue: Bridge reconstruction #49

siredmar:
Fixed bug mentioned in:
https://github.com/bvschaik/julius/issues/49


crudelios:
Siredmar I unfortunately don't think you got it right for this fix. Please check my comment for issue #49 for an explanation.

PR #51, opened 2019-01-28
Feature: capture mouse on full screen with multiple displays.

crudelios:
When there is more than one display and the user sets the game to fullscreen in one of the displays, prevent the mouse from leaving that display.

### DIFF
--- a/src/platform/screen.c
+++ b/src/platform/screen.c
@@ -54,7 +54,11 @@ int platform_screen_create(const char *title)
             return 0;
         }
     }
-
+#if !defined(__APPLE__)
+    if (fullscreen && SDL_GetNumVideoDisplays() > 1) {
+        SDL_SetWindowGrab(SDL.window, SDL_TRUE);
+    }
+#endif
     return platform_screen_resize(width, height);
 }
 
@@ -108,6 +112,12 @@ void platform_screen_set_fullscreen(void)
         return;
     }
     SDL_SetWindowDisplayMode(SDL.window, &mode);
+
+#if !defined(__APPLE__)
+    if (SDL_GetNumVideoDisplays() > 1) {
+        SDL_SetWindowGrab(SDL.window, SDL_TRUE);
+    }
+#endif
     setting_set_display(1, mode.w, mode.h);
 }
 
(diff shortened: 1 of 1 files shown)


jdmillard:
I don't have any helpful insight about this, but I'm excited for this fix!

bvschaik:
I tried this out on my MacBook (the only OS I have at home that I can run with two screens), and the experience was... not so nice. The mouse cursor tried to move to the other screen but was forced back to the fullscreen window, like an elastic band pulling it back.

I'm not sure about this... personally I like to have reference material and/or notes open on the 2nd screen when I'm playing the game, and on macOS you cannot use these when the mouse is locked to the game, since cmd-tabbing pulls the game to the background. How does that work for Windows and Linux?

crudelios:
Windows works well, but I can't test on Linux as I only have it installed via VirtualBox, which ignores any sort of mouse capture.

That behavior in macOS seems to be a known problem with SDL (check https://forums.libsdl.org/viewtopic.php?p=38517), which apparently can be fixed/mitigated by compiling SDL with `SDL_MAC_NO_SANDBOX` defined. That is a major hassle, and I'm not sure it would be worth the trouble.

Still, I think it's common nowadays that fullscreen games do capture the mouse.
My suggestion would be to add a setting with that option to make everyone happy, but I know implementing new settings isn't easy (Should unused bytes in c3.inf be used? Should there be a new settings file? Which format to use? Could those settings be changed in the game's GUI? Etc).

anon569:
I tested it on Linux (Ubuntu 18.10), and it's working fine for me.

jdmillard:
This change is awesome and I agree that the standard is for games to capture the mouse. I did a test and I can add a second (or third?) witness that this works great on Windows.

crudelios:
So only Mac OS is giving trouble...

I'll try to figure something out.

crudelios:
I haven't had time to look into this due to lack of time.

A temporary hacky solution would be to disable this feature for Apple until I can investigate this issue further.

@bvschaik if you agree with that temporary solution I can implement the changes and eventually find a fix for macOS, otherwise I'll just close the pull request for now.

bvschaik:
Agreed.

crudelios:
Done. Apple support removed for now.

Issue #48, opened 2019-01-24
Crash on startup


NotMe12:
Program crashes and fails to respond on startup screen stating 'Click to start'. 

![image](https://user-images.githubusercontent.com/47008386/51701055-7e29b300-2008-11e9-9615-becf6e4b3a8f.png)

This is with the English GOG version freshly reinstalled. Monitor resolution is 1920 x 1080. After restarting computer it allowed me to play for roughly 10 minutes before it crashed in the same fashion, but actually in game this time.

Since then it will not let me proceed past the screen in the screenshot before it crashes again, as it did when first installed. 

Readout from julius-log.txt reads as follows:

INFO: Initializing SDL

INFO: SDL initialized

INFO: Loading game from working directory

INFO: Creating screen 1920 x 1080, fullscreen? 1

INFO: Texture created (1920 x 1080)

INFO: model loaded

Thanks for your time!

crudelios:
When the game works for 10 minutes and then hangs, does a mouse click cause the hang?

NotMe12:
I believe it may do? I have certainly been in the process of doing something so whether it was just timed rightly I cannot say, but I believe I did click the mouse just before.

crudelios:
This seems to be an SDL event issue, however I don't know if it's a problem on your end or with the game as I can't reproduce your issue!

NotMe12:
Is there anything I can do or try in order to solve it?

crudelios:
You can try loading the game in a different computer to check if the problem persists. Also, what is your version of Windows?

NotMe12:
Latest version, I believe. I updated it earlier today and it made no difference. Be interesting if anyone else has this problem and whether they can sort it out or not.

crudelios:
I'm at a loss with your problem... Have you tried redownloading the game + SDL libraries?

NotMe12:
So I've downloaded the zip without the .exe in that you have to do the console commands with. Unfortunately I don't fully understand how to do it to create the .exe. What does "After cloning the repo" mean? And then do I just execute the commands in the cmd prompt whilst inside that directory?

Also I don't know whether you think I should try the method of "GOG version with InnoExtract" instead and see if that changes anything?

Thanks!

crudelios:
@NotMe12 there is now a link at the main page to grab the latest version, so you don't need to compile it yourself.

Please try a newer build, maybe one of them will somehow fix your issue.

NotMe12:
[...]

PR #52, opened 2019-01-28
Added version information upon startup

siredmar:
Added some versioning information upon startup to help with issues like #48 
The output does look like this:
`$ ./julius`
`INFO: Built with git commit hash: v1.0.0-9-g7868516950-dirty`
`INFO: Initializing SDL`
`INFO: SDL initialized  `

There will be a generated version.h in src/platform/version.h by cmake which.


src/platform/julius.c
```diff
@@ -325,9 +326,11 @@ static void teardown(void)
 
 int main(int argc, char **argv)
 {
+    SDL_Log("Built with git commit: %s\n", GIT_COMMIT_HASH);
+    
     const char *custom_data_dir = (argc > 1 && argv[1]) ? argv[1] : NULL;
     setup(custom_data_dir);
-
+
```
crudelios: Picky picky! Get rid of this extra space :P

siredmar: removed newline

src/platform/julius.c
```diff
@@ -325,9 +326,10 @@ static void teardown(void)
 
 int main(int argc, char **argv)
 {
+    SDL_Log("Built with git commit: %s\n", GIT_COMMIT_HASH);
+    
     const char *custom_data_dir = (argc > 1 && argv[1]) ? argv[1] : NULL;
     setup(custom_data_dir);
-
```
bvschaik: Can I have the newline (excluding the spaces ;) ) back please?

siredmar: Done. You got the newline back.

jdmillard:
I pulled `master` after this was merged. My console displays:
```
INFO: Built with git commit:
```

My "clean build process" on Windows is:
```
cd build
cmake.exe .. -G "MinGW Makefiles"
mingw32-make.exe -j8
julius.exe "path\to\GOG Games\Caesar 3"
```

...

OKAY, I took a closer look and this. This feature depends on cmake executing a git command which breaks for the following git setup: If "Git for Windows" is used to clone, then generic CMD is used to compile. Building this way might sound strange, but there are real reasons for some people to build this way on Windows machines (too hard to explain here).

I suggest to leave the new CMake logic as-is. However, if no commit hash string was set during CMakeLists.txt, the `INFO` print statement should be avoided altogether. Your thoughts?

PR #54, opened 2019-01-29
No eternal fires or reversible earthquakes

### LINKED ISSUE #53: Earthquake and Eternal Fires
Credit of finding this bug goes to my brother. I just helped confirm and reproduce the issue.

When an earthquake happens, buildings are destroyed and set on fire. Everything behaves fine. Then as the fires die out on their own and turn to ash tiles, they convert neighboring quake tiles into ash. Which can then be cleared.

Also, as all the fires die down, a few of them are left burning forever - which is not seen when buildings burn down under normal conditions. The number of tiles that become eternal fires seems nondeterministic - load the same .sav many times. Sometimes there were 8 fires and sometimes there were 9 (maybe dependent on the order that things catch fire?).

Load the attached saved game (which is moments before the earthquake) and you'll see this behavior. Load the same .sav in the original GOG game and see the expected comparison.

[eq test2.sav](https://github.com/bvschaik/julius/files/2805284/eq.test2.zip)

image

crudelios:
Prevents buildings from never stopping burning and prevents burning ruins replaced by earthquake tiles from being able to "stop burning", thus reconverting the earthquake to rubble.

Fixes #53.

### DIFF
--- a/src/building/maintenance.c
+++ b/src/building/maintenance.c
@@ -18,11 +18,13 @@
 #include "map/building.h"
 #include "map/building_tiles.h"
 #include "map/grid.h"
+#include "map/property.h"
 #include "map/random.h"
 #include "map/road_access.h"
 #include "map/road_network.h"
 #include "map/routing.h"
 #include "map/routing_terrain.h"
+#include "map/terrain.h"
 #include "map/tiles.h"
 #include "scenario/property.h"
 #include "sound/effect.h"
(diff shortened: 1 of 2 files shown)


bvschaik:
Could you put "Fixes #53" in your commit message (instead of "Fixes issue #53")? That way the issue automatically gets closed when this PR is merged.

crudelios:
The things I learn!

Done.

bvschaik:
I've investigated the issue, and the error is actually in `src/scenario/earthquake.c` where the original building is being set to DELETED_BY_GAME instead of the (new) ruin tile in the earthquake's path.

See commit c068da3879fd07007e29df157d8accab02a788f3.

crudelios:
Makes sense and is simpler!

PR #55, opened 2019-01-29
Add a link to the latest build of the game in the readme and make the instructions easier to understand.

crudelios:
I think this will help support julius, allowing everyone easy access to the latest bug fixes and improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,11 @@ UI enhancements include:
 
 Julius requires the original assets (graphics, sounds, etc) from Caesar 3 to run.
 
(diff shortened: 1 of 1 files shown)


jmillard7:
Very nice, this will let me get updates without setting up to compile it myself or having my brother send me his compiled .exe

crudelios:
I'm not sure I understood, you must copy both the exe and the DLLs at least the first time ypu download from that link.

If you don't, then it might not work due to incompatibiliities with the DLLs.

PR #57, opened 2019-01-30
Dpi awareness and window events overhaul.

crudelios:
This pull request was created at first to fix a bug with high DPI screens on Windows and Linux, but I went on to change screen events a lot.

When playing julius on high DPI displays, even though Windows scaled the game, the mouse cursor wasn't scaled so it became too tiny/big on custom DPI screens. In addition, the mouse wouldn't correctly scroll the map at the screen border, ony 1 or 2 pixels before the border (even on fullscreen!).  Also, Linux didn't support DPI scaling at all.

Because of that, I decided to implement DPI awareness. What started as a small project to get regular-sized mouse cursors became... a mess. SDL only supports DPI when it wants to, so it was a hinderance to check that every OS and screen event was working properly.

Fortunately after a long time of grinding I managed to eliminate nearly all of the hacks needed to get everything working.

DPI Support status:
- Windows is fully supported. Julius properly detects the correct DPI, responds to system DPI changes and, when moved across monitors with different DPI's, rescales accordingly.
- Linux/X11 is nearly as supported as Windows, except there is no per-monitor DPI scaling, only general system scaling. Other than that, the app also responds to system-wide DPI changes.
- Mac OS X support for DPI changes is disabled for now. Enabling DPI support for Mac would require an app bundle, which seems to be more trouble than it's worth. Also, strangely enough the mouse cursor scales properly in Mac OS even with DPI awareness disabled.

Besides the DPI awareness, there are three other important changes in this code:
- Major improvements to window positioning, the game no longer defaults to placing a window at coordinates [0, 0], but instead at the center of the screen. If the window size is changed via settings or hotkey, a custom position will be set based on the old one. That position is remembered even if the user goes to fullscreen and returns to windowed mode in the same session.
- Implemented minimum window size. Caesar 3 never could be set to anyrhing less than 640x480, and neither should Julius.
- Enabled threaded resize and move events. Allows the window to be redrawn while being resized, resulting in a strangely satisfying pattern when resizing the window in the normal city view.

Since this is a big, invasive change there are bound to be bugs so this PR needs some testing. I hope there aren't many though. I tested things as much as I could.


crudelios:
Not a single build passed! Great, and I also thought I had squashed the commits, but no, guess I was wrong.

If it's okay with you I'll close the pull request, fix the build failures, squash the commits and reopen the pull request when it's ready.

I believe one of the build failures in Travis is because the SDL version is old and doesn't support some functions.

bvschaik:
That's fine with me :)

crudelios:
Closed for maintenace.

crudelios:
@bvschaik I think I'll need some help here. The SDL lib used by travis is very old (2.0.2) and my changes require at least SDL 2.0.4. It seems it is possible to manually compile and build SDL, but as it implies, it's an extra build that takes even longer than julius. So I'm not sure what to do - the code compiles locally, but will fail on the automated builds.

bvschaik:
Can we introduce sdl version checks? I'd like to keep the code compatible with the lowest SDL2 version possible, because users might run an older version.

crudelios:
Sure, as a fallback the DPI will not be checked and that removes any need for the more recent SDL API calls. I think this is best done only at compile-time.

I still think #68 is useful, though.

crudelios:
Problems fixed, more bugs squashed.

Please comment and test.

pcn:
Built on a linux laptop (ubuntu 18.10, 3840x2160 screen running gnome) and it seems to work fine, scaling the game to what appears to be the right size.

src/input/mouse.c
```diff
@@ -18,6 +19,13 @@ const mouse *mouse_get(void)
 
 void mouse_set_position(int x, int y)
 {
+    double dpi_scale = platform_dpi_get_scale();
+
+    if (dpi_scale != 1.0f) {
+        x = (int)(x * dpi_scale);
+        y = (int)(y * dpi_scale);
+    }
+
```
crudelios: Performing the mouse scaling should probably go to `platform\julius.c`.

src/platform/screen.c
```diff
@@ -1,22 +1,141 @@
...
+static void platform_screen_position_window(int width, int height)
+{
+    int title_height = platform_dpi_get_title_height();
+
+    /* No position information, center window */
+    if ((!window_pos.x && !window_pos.y) || !window_pos.w || !window_pos.h) {
+        int display = SDL_GetWindowDisplayIndex(SDL.window);
+        SDL_SetWindowPosition(SDL.window, SDL_WINDOWPOS_CENTERED_DISPLAY(display), SDL_WINDOWPOS_CENTERED_DISPLAY(display));
+        SDL_GetWindowPosition(SDL.window, &window_pos.x, &window_pos.y);
+    } else if (width != window_pos.w || height != window_pos.h) {
```
crudelios: For my own reference: this code must go to `platform_screen_center_window`, and we call that function from here instead. Otherwise when a user presses `F5` to center whe window, the window positions won't be updated.

src/platform/screen.c
```diff
@@ -1,22 +1,141 @@
...
+    SDL_GetDisplayBounds(SDL_GetWindowDisplayIndex(SDL.window), &monitor_coordinates);
+    int minimum_y = monitor_coordinates.y + title_height;
+    int minimum_x = 0;
+    int total_displays = SDL_GetNumVideoDisplays();
+    for (int display_index = 0; display_index < total_displays; display_index++) {
+        SDL_GetDisplayBounds(display_index, &monitor_coordinates);
+        minimum_x = (monitor_coordinates.x < minimum_x) ? monitor_coordinates.x : minimum_x;
+    }
+
+    window_pos.x = (window_pos.x < minimum_x) ? minimum_x : window_pos.x;
```
crudelios: For my own reference: Should there also be a check for `maximum_x` and `maximum_y`? I guess it wouldn't hurt...

crudelios:
This PR now has some merge conflicts and will have even more once #114 and #107 are merged. Once those changes are in, I'll rebase this and fix the conflicts so it can be merged again.

crudelios:
I'm closing this for now as it's pretty incompatible with the recent merges and once touch support is merged it will need even more changes.

PR #67, opened 2019-01-31
Vertical scroll rate of original game

### LINKED ISSUE #65: Scroll Speed (City View)
When compared to the original game, it's obvious that the screen edge mouse scrolling moves the map at a slower speed. This is admittedly a nitpick, but I suspect it's super easy to fix. EDIT: there's an option for this in-game, see comments below for more context.

Perhaps
```
    int dx = 1;
    int dy = 2;
```
in `view.c`?

jdmillard:
Fixes #65 

Confirmation that this matches the original game is best confirmed by loading a big map, starting in one corner, then scrolling diagonally while watching the minimap. Then repeat on GOG and Julius versions.

### DIFF
--- a/src/city/view.c
+++ b/src/city/view.c
@@ -158,7 +158,7 @@ int city_view_scroll(int direction)
         return 0;
     }
     int dx = 1;
-    int dy = 2;
+    int dy = 4;
     switch (direction) {
         case DIR_0_TOP:
             data.camera.y -= dy;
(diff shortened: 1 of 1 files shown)


jdmillard:
Per #65 , closing this to prevent issue clutter.

If anyone wants to come back to this, a fancy PR isn't required. Just bump that value to match the original game.

PR #68, opened 2019-01-31
Download, compile and build the latest SDL and SDL_mixer sources when building on Travis.

crudelios:
This ensures the latest version of SDL libraries is used when builing, allowing the code to use all the current features of SDL without the build failing.

This needs to be done because Travis has a rather old version of SDL (2.0.2 or 2.0.4 depending on the distribution, while SDL is at version 2.0.9) - my dpi awareness patch requires SDL 2.0.5, and in the future something else may also require a newer SDL version.

In order to improve build times, the compiled main SDL library is cached so it doesn't need to be built every time. This saves upwards of 90 seconds from the build time.
> Unfortunately SDL_mixer cannot be cached as, for some strange reason I could not discern, the clang job will fail with error `clang not found` during SDL_mixer `make install` if using the previously compiled version from the cache. However, as building SDL_mixer only takes about 15 seconds, that seems a reasonable tradeoff.

This has been fixed and now SDL_mixer is cached, reducing its build time from 15 to 2 seconds.

Combined build times increased from 3 to 6 minutes when building SDL from source, but actually decreased to 2:30 when using the cached SDL libraries that were previously compiled.


bvschaik:
Actually, I think the current (outdated) Travis build can be useful for the DPI build, because this build can check if it still compiles on older versions of SDL. Not everyone is running an up-to-date system.

crudelios:
You have a point there. Then how about creating jobs for both older and more recent SDL versions?

bvschaik:
That's a good idea.

crudelios:
I've implemented Travis configurations with the most recent versions of SDL (2.0.9) and SDL_mixer (2.0.4), and with versions 2.0.0 of both libraries. Also, the SDL versions to be installed can easily be changed via an environment variable.

There's also the possibility of testing in Mac OS. Mac OS can use both gcc and clang as compilers and any version of the SDL libraries (I have tested macOS builds and they seem to work, even when using cached versions of SDL). However, Mac OS takes a longer time to start the build than Linux. Also, checking SDL and Clang on Linux and Mac OS with the original 2.0.0 versions or the most recent builds of SDL makes a total of 8 builds, which seems a bit overkill for me.

So my suggestion of possible Travis builds is:
- Linux, gcc, new SDL libs;
- Linux, gcc, old SDL libs;
- Linux, clang, new SDL libs;
- Mac OS, clang, new SDL libs.

What do you think?

bvschaik:
Sounds good to me. 8 builds is indeed overkill.

PR #70, opened 2019-01-31
Commits

CyranoDeVergerac:


bvschaik:
This is not CaesarIA, and it's not written in C#. Dalerank's branch is based on a very old version of Julius, and is nowhere near mergable. Please do not make pull requests with other people's code.

PR #71, opened 2019-02-01
increases max walkers, buildings, and routes 4x

jdmillard:
It is well-known within the community that there is a hard cap on the number of builders, walkers, and routes:

* [Heaven Games Thread 1](http://caesar3.heavengames.com/cgi-bin/forums/display.cgi?action=ct&f=2,3608,2175,all)
* [Heaven Games Thread 2](http://caesar3.heavengames.com/cgi-bin/forums/display.cgi?action=ct&f=2,6184,1250,all&st=recent)
* [Steam Community](https://steamcommunity.com/app/517790/discussions/0/350540974011956646/)
* [Reddit Community](https://www.reddit.com/r/impressionsgames/comments/1qq5vw/caesar_iii_recap_on_building_and_walker_limit/)

These limits seem to have been motivated by ancient PC hardware that was round in the year 1998. Load this saved game and try to build a few hundred new houses. This PR increases those limits 4x.

[- data limit test.zip](https://github.com/bvschaik/julius/files/2819629/-.data.limit.test.zip)

I understand that this is a deviation from the behavior of the original game, so feel free to reject this PR and I'll harbor no ill will!

### DIFF
--- a/src/building/building.h
+++ b/src/building/building.h
@@ -4,7 +4,7 @@
 #include "building/type.h"
 #include "core/buffer.h"
 
-#define MAX_BUILDINGS 2000
+#define MAX_BUILDINGS 8000
 
 typedef struct {
     int id;
--- a/src/figure/figure.h
+++ b/src/figure/figure.h
@@ -6,7 +6,7 @@
 #include "figure/action.h"
 #include "figure/type.h"
 
-#define MAX_FIGURES 1000
+#define MAX_FIGURES 4000
 
 typedef struct {
     int id;
--- a/src/figure/route.c
+++ b/src/figure/route.c
@@ -4,7 +4,7 @@
 #include "map/routing_path.h"
 
 #define MAX_PATH_LENGTH 500
-#define MAX_ROUTES 600
+#define MAX_ROUTES 2400
 
 static struct {
     int figure_ids[MAX_ROUTES];


crudelios:
The only (major) problem with this is that it makes savegames incompatible! I believe the limits are hardcoded to the savegame format as well...

bvschaik:
Yes, that's also why the automatic builds are failing, since there's a test with a city close to the figure limit. Save-game compatibility is required for the 'vanilla' Julius.

jdmillard:
I figured this would get rejected, but I wanted the PR in the repo history because this is a feature that die-hard C3 fans care about. If they know it's possible it will boost support, hence the links (there are many more out there too).

Also, can you articulate the saved game incompatibility? I don't think the limit number is being saved in any .sav. What I've seen is that with this patch, I can build past the limit without error, save, then load in original C3 without issues. The limit seems to prevent **additional** buildings only, meaning that load/save capability appears to be identical.

What is the nature of the test that is failing?

Cheers!

bvschaik:
The limit number is implicitly saved by the number of bytes in the figure/building/route parts of the saved game. Original C3 only reads the original number of figures/etc, so it's missing data if you go over the limit.

It may seem to work correctly at the moment, but C3 does not do any checks on the IDs of buildings/figures. This means that a figure loaded in C3 may refer to a route which is beyond the limit. This means that the game will read beyond the data allocated to it, which will likely be "garbage" for the routing algorithm. The same goes for example for fighting figures: they have a reference to each other, and if one is beyond the limit, garbage will be read _and written_ to random memory which happens to be just after the figure list, corrupting and possibly crashing the game.

The test that is failing checks that figures are not appearing (and other weird things are happening) when the figure limit is reached.
Every test is created like this: load starting save in C3, run it for a while, write the reference save. Then, to run the test, I load the starting save in a stripped-down Julius, run it for the same number of ticks, save it, and compare it with the reference save. If there are any relevant differences (which basically means everything except timing/animation related stuff), it fails.

jdmillard:
Fascinating! Thanks.

PR #74, opened 2019-02-01
Change Travis configuration to include both recent and old SDL libraries and to test Mac OS.

crudelios:
Instead of downloading the SDL and SDL_mixer libs from apt, their sources are now directly downloaded from libsdl.org. After being compiled, the resulting binaries are cached for faster build times in subsequent builds.

Two versions of both libs are compiled in two separate jobs: The most current version and the original 2.0.0 version. This allows SDL API calls that are not compatible with older versions of SDL to be detected.

Also add a Mac OS X build using clang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,57 @@
 language: c
 
-compiler:
-  - gcc
-  - clang
+matrix:
+  include:
+  - os: linux
+    compiler: gcc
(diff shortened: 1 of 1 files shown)


crudelios:
(Due to my relative inexpirience with git, I messed up and ended up making the last pull request unusable. Sorry for that).

PR #78, opened 2019-02-02
fixed soldier blur issue

### LINKED ISSUE #69: Earthquake bug and Forts' Animation
I was playing a random map of Caesar III. not one from the campaign when at some point an earthquake happened. After the quake was over, instead of the unpassible chasms that are usually left behind after an earthquake, they were replaced by lines of rocks and bushes. Near them there were several unstoppable forever burning houses. Even an iron mine could be put near the newly formed rocks in the places where the quake passed by. 
Again in this mission I noticed that sometimes soldiers from the forts I put down were trembling without any reason especially when the game speed is being increased. I destroyed the fort and even rebuild it several time, but this problem persists. Here is the save file of the map.
[Augusta Trevorum.zip](https://github.com/bvschaik/julius/files/2818044/Augusta.Trevorum.zip)

jdmillard:
Fixes #69 

Please inspect the load/save logic with the changed data type. My manual tests passed, but I don't fully understand the buffer read/write logic. If there is an issue, instruct me on what changes I can add to meet your standards.

Cheers.

### DIFF
--- a/src/figure/figure.h
+++ b/src/figure/figure.h
@@ -41,8 +41,8 @@ typedef struct {
     short destination_grid_offset; // only used for soldiers
     unsigned char source_x;
     unsigned char source_y;
-    signed char formation_position_x;
-    signed char formation_position_y;
+    short formation_position_x;
+    short formation_position_y;
     short __unused_24;
     short wait_ticks;
     unsigned char action_state;
--- a/src/figure/figure.h
+++ b/src/figure/figure.h
@@ -41,8 +41,8 @@ typedef struct {
     short destination_grid_offset; // only used for soldiers
     unsigned char source_x;
     unsigned char source_y;
-    short formation_position_x;
-    short formation_position_y;
+    unsigned char formation_position_x;
+    unsigned char formation_position_y;
     short __unused_24;
     short wait_ticks;
     unsigned char action_state;


jdmillard:
Also, something to think about: are there any other map position (x,y) integers that are represented as `unsigned char` that are no good beyond 127 tiles? This may be worth some investigation.

bvschaik:
Can you elaborate what went wrong and why this fixes the issue?

jdmillard:
I put a lot of details in the ticket. #69

bvschaik:
@dalerank Setting `formation_position_x`:
[Soldier](https://github.com/bvschaik/julius/blob/159a5084a4a5e899d87c9ce89b7d330ae2141dc9/src/figuretype/soldier.c#L244) -> formation position plus soldier's offset within formation layout
[Enemy](https://github.com/bvschaik/julius/blob/159a5084a4a5e899d87c9ce89b7d330ae2141dc9/src/figuretype/enemy.c#L166) -> just the soldier's offset within formation layout

jdmillard:
@bvschaik , yes exactly.

@dalerank :

The bug is manifest in `soldier.c`:

```
    f->formation_position_x = m->x + formation_layout_position_x(layout, f->index_in_formation);
    f->formation_position_y = m->y + formation_layout_position_y(layout, f->index_in_formation);
    
    switch (f->action_state) {
        case FIGURE_ACTION_150_ATTACK:
            figure_combat_handle_attack(f);
            break;
        case FIGURE_ACTION_149_CORPSE:
            figure_combat_handle_corpse(f);
            break;
        case FIGURE_ACTION_80_SOLDIER_AT_REST:
            map_figure_update(f);
            f->wait_ticks = 0;
            f->formation_at_rest = 1;
            f->image_offset = 0;
            if (f->x != f->formation_position_x || f->y != f->formation_position_y) {
                f->action_state = FIGURE_ACTION_81_SOLDIER_GOING_TO_FORT;
            }
            break;
```
Specifically at line 259.

In my test (there's a download in the ticket #69 ), I was seeing:
f->x: 60
f->formation_position_x: 60
f->y: 136
f->formation_position_y: -120 !!!!!!!!!!!!!!!!!!
these are the numbers for the single javelin soldier in the save game attached (once he arrives at fort)

The state machine is broken and switches back and forth.

However, the cause comes down to the type used for f->formation_position_y. It is being functionally treated as `int` but the type is `signed char`. Which can only reliably cast to and from `int` up to 127.

@bvschaik 

Switching to `unsigned char` would open up an extra bit, which I think increases the valid range to 255, thus opening up functionality for maps that are size 160 tiles. We might want someone else to verify that this is correct first. We can do that if you feel it would more closely mirror the original game.

@dalerank ,

your suggestion of `__unused_24` is exactly what I was referring to in regard to the buffer read/write logic. - which I have not wrapped my head around. If we switched to `unsigned char` we could leave this as-is, correct?

Thank you all for your input as you are far more familiar with the codebase than I.

jdmillard:
Based on a discussion with @bvschaik, using unsigned char might be causing problems with enemy soldier logic. An incomplete test using 'autopilot' showed some differences. Maybe a signed int or short is the way to go. But then we need to make sure we're not breaking the load/save buffering.

This needs some investigation and maybe a dedicated ticket and since the original one is now focused on earthquake rendering.

bvschaik:
Actually, I checked your test yesterday and the `unsigned char` works for enemy soldiers as well. This is pure coincidence though, since the resulting variable is cast to an `unsigned char` again, and any wrapping from negative values is undone during that cast.

I'll make a more future-proof version, I will close this PR once I have it ready.

jdmillard:
Sounds good to me.

bvschaik:
Fixed in commit 75ef294.

Issue #64, opened 2019-01-31
Cutscene video volume control


jdmillard:
I tend to play with the sound settings at 50% so that I have some control over the game volume relative to other apps running simultaneously on my PC. There is no way to control the volume of the videos that play (Caesar requests, Barbarians attack, etc).

Reproduce this by setting all the volume settings to 5% (or even 0), then wait for a Caesar request and you'll hear it at full volume. To be fair, the original game had the same issue, but it's really annoying. Is it possible to make the video volume into one of the existing categories (Music, Speech, Sound Effects, City Sounds)?

crudelios:
I agree with this feature, that loud music can be annoying.

Also I would use the "Music volume" setting to determine the video volume.

bvschaik:
Good idea. Right now the video's sound is not played if Sound Effects are off (same as C3). Maybe we should use that setting?

I briefly looked into this yesterday: video sound is played using a custom music player in SDL. To change the volume, we need to change the volume of the music data somewhere in [platform/sound_device.c:next_audio_frame()](https://github.com/bvschaik/julius/blob/258ae61f3095c98262ad430ddb4e27eaa351ec4f/src/platform/sound_device.c#L156). It looks like we can use [SDL_MixAudioFormat()](https://wiki.libsdl.org/SDL_MixAudioFormat) for that, but I'm not too familiar with SDL and audio, so if anyone wants to pick this up I'd be grateful.

crudelios:
Since I already messed around with libsmacker I'll try to do it.

siredmar:
@crudelios: I didn't see that you are working on that because i am as well :)

My investigations so far: Like @bvschaik already pointed out that the custom player is used. 
I focused on video.c where next_audio_frame() extracts the next blob that gets played. I tried to find out what format the audio is within the smk files and played with SDL_MixAudioFormat(). Turned out that no AudioFormat prepared by SDL is working on this stream. However the audio stream seems to be simple 8 bit that can be scaled with any factor to reduce the volume.

Currently i'm working on a pull request for this.

bvschaik:
@siredmar Wouldn't it be better to let SDL_MixAudioFormat do its job in sound_device.c where the audio is converted to SDL format, rather than manually downscaling 8-bit audio?

crudelios:
@siredmar I hadn't started yet so no work duplicated :)

jdmillard:
[...]

PR #81, opened 2019-02-02
Videos now play with volume settings of 'sound effects'

siredmar:
The SMK raw stream gets converted to the SDL format with the given AUDIO_FORMAT (=AUDIO_S16MSB). 
Before the converted audio chunk gets passed its volume is being mixed with the settings for 'sound effects'.

### DIFF
--- a/src/platform/sound_device.c
+++ b/src/platform/sound_device.c
@@ -1,5 +1,6 @@
 #include "core/log.h"
 #include "sound/device.h"
+#include "game/settings.h"
 #include "SDL.h"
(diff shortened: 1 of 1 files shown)


jdmillard:
Fixes #64 .

(I don't know if that phase has to be in the PR description for it to automatically close that issue, or if putting it here will work? ... probably not?)

bvschaik:
Great!
@jdmillard It has to be in the commit message for the merge to automatically close the issue, though posting it as comment does link the issue to this PR on Github.

Issue #79, opened 2019-02-02
Fine-tune printing git commit hash


bvschaik:
Printing the git commit hash on startup can use some fine-tuning (and explanation). For me it says right now:
`INFO: Built with git commit: v1.0.0-22-g258ae61f30-dirty`
Where does the "v1.0.0-22-g" come from?
What about the dirty flag? Is that always set because the build process creates files in the working directory? (need to check AppVeyor/Travis build scripts)
What does it print for release versions?

Originally proposed change by @jdmillard:

---
I pulled `master` after this was merged. My console displays:
```
INFO: Built with git commit:
```

My "clean build process" on Windows is:
```
cd build
cmake.exe .. -G "MinGW Makefiles"
mingw32-make.exe -j8
julius.exe "path\to\GOG Games\Caesar 3"
```

...

OKAY, I took a closer look and this. This feature depends on cmake executing a git command which breaks for the following git setup: If "Git for Windows" is used to clone, then generic CMD is used to compile. Building this way might sound strange, but there are real reasons for some people to build this way on Windows machines (too hard to explain here).

I suggest to leave the new CMake logic as-is. However, if no commit hash string was set during CMakeLists.txt, the `INFO` print statement should be avoided altogether. Your thoughts?

Here's my horrible pseudo code suggestion:
```
if (GIT_COMMIT_STRING_WAS_SET_BY_CMAKE)
{
    print_stuff_about_the_commit();
}
```

_Originally posted by @jdmillard in https://github.com/bvschaik/julius/pull/52#issuecomment-459549167_

siredmar:
To explain what the string itself is:
v1.0.0-22-g258ae61f30-dirty -> **a**-**b**-**c**[-**d**]
**a**: latest git tag (e.g. v1.0.0)
**b**: commits above latest tag (22)
**c**: commit hash (10 digits) referencing the current build (e.g. g258ae61f30)
**d**: if the local working tree has modifications based on the hash seen in c (e.g. g258ae61f30) there will be a dirty flag set. This flag is optional as it won't be printed if the working tree is exactly the git commit without modifications

But where does that come from? 
[...]

PR #82, opened 2019-02-02
Version information fine tuning (#79)

siredmar:
Version information containing git version if cmake could extract that information. If not a fallback string is being printed

Example git can be found by cmake: 
``./julius                          ``
``INFO: Built with git commit: v1.0.0-27-g86939162e8-dirty``

Example git cannot be found by cmake:
``./julius                                 ``
``INFO: Built with git commit: unknown (please install git and/or add it to PATH to get build information)``


bvschaik:
One problem I see with this: release versions. Packagers download the .zip without any git information and build the package from there. How should we deal with that?

siredmar:
If you do a release compile without any git information, you need to maintain a extra version information within the codebase e.g. version_release.h that contains the release version.
If this is the way you want to go, its a minor thing to implement. But keep in mind, that you have to maintain it manually.

I would either then print the release_version information or the git related version information on startup.
Current release version is 1.0.0, right?

One thing to ask: Do the automatically build windows versions contain the git information?
If so, standard users who don't compile it manually won't be able to tell with which git commit the software was built.

siredmar:
@bvschaik please review the proposal i've made. However, I have to consider that you have to maintain the version number manually. :-/

crudelios:
Strangely enoguh, for me the built packages from appveyor don't show anything at all, not even the `Built with version` part. I made sure to download the artifact from the proper pull request.

shirakaba:
Sorry that I don't have time to read/discuss details at the moment, but if I'm understanding the situation correctly (CMake shouldn't depend on git) there may be a way to solve this by using a pre-commit hook to save a commit name to a text file.

```sh
# Here's an example script of how to get the latest tag
git describe --always --tag --abbrev=0 > git_tag.txt
```

That code

jdmillard:
Manually maintaining the version in a 'version.h' for a single project really isn't that hard, especially when there is a sole repo owner as is our case.

You might ask : what about folks using nightly builds and how to communicate version numbers when bugs are found? I think in that case a nightly build is identified by its date and/or a checksum (as well as the version.h which could be printed to console).

siredmar:
> Sorry that I don't have time to read/discuss details at the moment, but if I'm understanding the situation correctly (CMake shouldn't depend on git) there may be a way to solve this by using a pre- or post-commit hook to save a commit name to a text file.
> 
> ```shell
> # Here's an example script of how to get the latest tag
> git describe --always --tag --abbrev=0 > git_tag.txt
> 
> # A similar script could be used to write in the commit hash.
> ```

I have tried that right now and finished with my tests.
What i did was: put post hooks in merge, checkout and commit that generates a version.h file.
If developers all have those hooks installed this might work just fine. 
Problem is however that version.h file gets generated after the commit/merging/checkout is done. This means you have a dirty tree. One needs to check in that modified file in an extra commit generating a new version.h and so on.
Putting this version.h in the .gitignore file is working when getting the source via git. But when downloading the zip file, this file is missing obviosly.

So i guess not the way to go? @shirakaba maybe i have done something wrong or did not use the correct hook. Maybe you can share any further information.

bvschaik:
Actually, there is already a version in the source code for the windows build: [res/version.rc](res/version.rc).

I don't like the idea of a version.h that gets updated on every commit, even if it's automated.

What I'd like to have:
A single file where I can add the version number and a boolean flag whether it's an official release or not. Ideally, I'd like to generate both `res/version.rc` and some sort of `version.h` from that file.
Releasing a version then consists of 3 easy steps:
- Update version/release flag on master
- Create a tag for that version
- Update release flag to false on master

Now, what to print:
1. For release versions, Julius can print something simple like "Julius release v1.0.1"
1. For non-release versions without git directory: "Julius unknown development version (please install git)"
1. For non-release versions with git directory present: "Julius development v1.0.1-$git_commit" (I don't think we need the "tag + _n_ commits" part, perhaps the dirty flag is useful here)

bvschaik:
@crudelios The Windows build doesn't write anything to the `julius-log.txt` because the log message is written before log output is redirected to that file. That needs to be fixed.

@siredmar Please no actual code in `.h` files: if you include the `.h` file in multiple source files, you're toast. And please keep the formatting consistent with the rest of the project (whitespace, curly braces position, naming of include guards, etc).

shirakaba:
@siredmar 

> Problem is however that version.h file gets generated after the commit/merging/checkout is done. This means you have a dirty tree. One needs to check in that modified file in an extra commit generating a new version.h and so on.
> Putting this version.h in the .gitignore file is working when getting the source via git. But when downloading the zip file, this file is missing obviosly.
> 
> So i guess not the way to go? @shirakaba maybe i have done something wrong or did not use the correct hook. Maybe you can share any further information.

I suppose it makes sense that the change doesn't get committed. Okay, assuming a pre-commit hook, maybe the hook just needs to also include a procedure for adding the updated file into the commit (could be as simple as `git add git_tag.txt`..?). I'm no expert on git hooks, unfortunately.

shirakaba:
> What I'd like to have:
> A single file where I can add the version number and a boolean flag whether it's an official release or not. Ideally, I'd like to generate both `res/version.rc` and some sort of `version.h` from that file.
> Releasing a version then consists of 3 easy steps:
> 
> * Update version/release flag on master
> * Create a tag for that version
> * Update release flag to false on master

Is this currently a manual operation? Would a release script be helpful?

I use this script to help with CI on some projects. It includes my process of writing a tag to a file. Of course that part could be removed, and the required toggling of the release flag could be introduced.

For inspiration:

```sh
# Example of tagging a commit as "v1.1.1"
./tag.sh "v1.1.1"
``` 

```sh
#!/usr/bin/env bash

echo $1 > ./git_tag.txt
git add ./git_tag.txt
git commit -am "Tag as $1"
git push origin master
git tag $1 master
git push origin $1
```

bvschaik:
> Is this currently a manual operation? Would a release script be helpful?

Currently I have made only one release (which was a bit of trial and error), so I have no real release procedure as of yet... perhaps your script will be useful for that.

PR #86, opened 2019-02-06
Fix bug where some rocks in the range of a reservoir would not redraw on map rotate.

### LINKED ISSUE #84: Some rocks are not being redrawn on map rotate when they are in range of a reservoir.
If you place a reservoir next to a group of rocks and then rotate the map, some of the rocks won't redraw.

This doesn't happen when there is no reservoir in range of the rocks.

Here are some images of the bug (tested in the campaign map Syracusae):

rock-reservoir-bug-1
rock-reservoir-bug-2

I haven't properly investigated the cause of the bug, but it must be related to `line 102` of `src/map/tiles.c`. Here's lines 100 to 102:

```
return map_terrain_is(grid_offset, TERRAIN_ROCK) &&
           !map_property_is_plaza_or_earthquake(grid_offset) &&
           !map_terrain_is(grid_offset, TERRAIN_RESERVOIR_RANGE | TERRAIN_ELEVATION | TERRAIN_ACCESS_RAMP);
```

Note that the code is inside the function `is_updatable_rock` that was just added in the (as of now) latest commit `695b5ee4b34e967c896b7b560813a1be4487f1ba`, but I had this issue before and couldn't pinpoint its cause. The latest commit changes made me realize that it could be because of reservoirs being in range, and a quick test proved that to be the case.

That `TERRAIN_RESERVOIR_RANGE` is somehow preventing some rocks that should be drawn to redraw, but since it doesn't happen on all the rocks I can't really tell if it is the cause of if the problem lies in the rock drawing routine. I can investigate further when I have some time.

Also, it might be unrelated, but all the problematic rocks were at the edge of the map.

This is definitely a drawing issue as loading the game in C3 and then rotating the map fixes the rocks.

crudelios:
Just in case this is correct...

Fixed #84.

### DIFF
--- a/src/map/tiles.c
+++ b/src/map/tiles.c
@@ -99,7 +99,7 @@ static int is_updatable_rock(int grid_offset)
 {
     return map_terrain_is(grid_offset, TERRAIN_ROCK) &&
            !map_property_is_plaza_or_earthquake(grid_offset) &&
-           !map_terrain_is(grid_offset, TERRAIN_RESERVOIR_RANGE | TERRAIN_ELEVATION | TERRAIN_ACCESS_RAMP);
+           !map_terrain_is(grid_offset, TERRAIN_ELEVATION | TERRAIN_ACCESS_RAMP);
 }
 
 static void clear_rock_image(int x, int y, int grid_offset)


bvschaik:
It is. I don't know how I translated `0x600` to something with RESERVOIR_RANGE...

PR #87, opened 2019-02-07
Change the information window position when a city tile is right-clicked.

crudelios:
This is a small cosmetic change, but one I had already implemented in my Pharaoh widescreen mod and I think makes the GUI a little bit better.

With this change, the information window's position is set to the mouse posiition (instead being centered either at the top or at the bottom of the screen), except if it would not fit in the window. In terms of positioning, it behaves very much like a right-click context menu.

The "special orders" menus have also been changed to appear depending on the information window position (something I couldn't implement in Pharaoh widescreen because assembly).

### DIFF
--- a/src/window/building/common.c
+++ b/src/window/building/common.c
@@ -4,10 +4,54 @@
 #include "building/model.h"
 #include "city/labor.h"
 #include "city/population.h"
+#include "city/view.h"
 #include "graphics/image.h"
 #include "graphics/lang_text.h"
+#include "graphics/screen.h"
 #include "graphics/text.h"
 #include "sound/speech.h"
+
+void window_building_set_possible_position(int *x_offset, int *y_offset, int width_blocks, int height_blocks)
+{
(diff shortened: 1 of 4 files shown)


bvschaik:
C3 has two positions for the dialog: top of the screen and bottom of the screen. I think the idea of the original game was to keep the right-clicked building in view when you see the details. For warehouses I can imagine this can be useful when you want to adjust the resource controls, so you can visually see how full the warehouse is. Now with widescreen (actually "high-screen"), the distance between the dialog and the building is so far it's getting ugly.

Anyway, good improvement, but I have found one small glitch: when the window is padded with black area at the bottom, the dialog can be placed on top of that.
overflow

crudelios:
Yes, and I kept the old behavior in the 640x480 resolution (it can be increased to 800x600 or 1024x768 if you think it's better).

Oh and good catch, I'll account for that black area then.

PR #94, opened 2019-02-13
Slightly optimize video drawing and line drawing.

crudelios:
Avoid the constant calling of graphics_draw_pixel and draw directly to graphics_canvas.

To prevent clipping errors, clipping for the whole area to be drawn is checked beforehand.

Should marginally improve performance.

### DIFF
--- a/src/graphics/graphics.c
+++ b/src/graphics/graphics.c
@@ -189,54 +189,60 @@ void graphics_clear_screen(void)
     memset(canvas.pixels, 0, sizeof(color_t) * canvas.width * canvas.height);
 }
 
-void graphics_draw_pixel(int x, int y, color_t color)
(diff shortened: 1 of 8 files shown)


crudelios:
Good question, I didn't check that and assumed it was used somewhere else! I'll confirm that.

Also, what do you think about separating `graphics_draw_line` into `graphics_draw_horizontal_line` and `graphics_draw_vertical_line`? That way the `if(x1 == x2)` and `if(y1 == y2)` would not be needed, saving two conditionals and potential CPU pipeline flushes.

bvschaik:
Good suggestion, yes, please do.

crudelios:
@bvschaik Changed and tested, seems to be working fine.

PR #101, opened 2019-02-15
Cache the minimap to improve performance on window invalidation.

crudelios:
Should improve tooltip performance.

As a consequence of this change, two bugs were found and fixed.

When the minimap was invalidated (`data.minimap_redraw_requested` set to `1` in `sidebar.c`), it was never set back to `0` so the minimap was being redrawn every single frame. This should improve overall performance.

`graphics_save_to_buffer` and `graphics_draw_from_buffer` were using the height size, and not the width size, to find the starting pixel, which resulted in the wrong pixels being copied if the width and height were different.

### DIFF
--- a/src/city/view.c
+++ b/src/city/view.c
@@ -3,6 +3,7 @@
 #include "core/direction.h"
 #include "map/grid.h"
 #include "map/image.h"
+#include "widget/sidebar.h"
 
 #define MENUBAR_HEIGHT 24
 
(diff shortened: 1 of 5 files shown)


crudelios:
This code can still be improved. It is caching the map during map scroll, which makes no sense giving that map scroll invalidates the minimap.

In fact, if the minimap offsets don't change (i.e. the part of the city that is visible in the minimap), it is possible to use the cached minimap and draw the viewport area (the yellow rectangle) over it.

Issue #103, opened 2019-02-15
Visual feedback for the land clearing tool


devnoname120:
When clearing some land using the tool, there is no visual feedback. Only when left click is pressed and mouse is dragged does the user know what will be destroyed.

bvschaik:
Good one. Maybe highlight the building/tile in red when you hover over it?

crudelios:
This is a visual change from C3, but I agree it would be an useful one!

> Maybe highlight the building/tile in red when you hover over it?

Seems like a good idea :)

PR #107, opened 2019-02-22
Touch support.

crudelios:
This is a touch support branch.

It is feature-complete, but it may still have bugs.

### 1. Touch usage:

A single finger is equivalent to mouse usage in most menus. You can tap any menu with one finger and it will  behave as a left mouse click.

- Touch actions override mouse actions. If you are touching the screen and moving the mouse, the mouse cursor will not move.
- In the city window, if nothing is being built, you can drag the finger to scroll the map. This also works for the empire view.
- When building something, there are two types of buildings with different behaviors: normal buildings such as schools and prefectures, and special buildings like roads, aqueducts, gardens, plazas, walls, houses and the clear tool. These buildings are special because then can be dragged to place multiple copies of them.
- If a single building is being built, you can drag the finger to show the shadow of the building's construction. Releasing the finger will place the building at the desired location. To cancel building construction, tap anywhere on the map with a second finger. To scroll the map when attempting to build something, get close to the city edges (NOT the window edges).
- If a road, houses, walls, gardens, plazas, reservoirs are being built or land is being cleared, you can select and drag that building with a finger. Releasing the finger creates the buildings in the selected area. The map can be scrolled the same way as when building a normal building. To cancel this build, tap once with a second finger. To restart the build after cancelling it, release and hold the first finger in the desired spot.
- If a legion is selected, you can scroll the map by touching and dragging a finger like you would if no construction was active. However, if you tap anywhere on the map, the legion will move there/attack. To not use the legion, press and hold with a finger and thap anywhere on the map with another finger. Please note that as of now there is no visual feedback when a legion is selected, as the only feedback on the original C3 was the mouse cursor, which for obvious reasons doesn't appear when using touch.
- To view a building information, make sure no construction is active. Place a finger in the tile you want to inspect, and tap anywhere in the city with the second finger.
-  To right-click in a menu, tap with two fingers. For better control you can press and hold with the first finger on the position to right-click, and then tap anywhere with a second finger.
- To view tooltips, touch and hold where you want to check.
- You can scroll text or the file dialog windows by touching and dragging up/down with two fingers at the same time.

### 2. Pending changes:

- [x] Empire map scrolling.
- [x] Right click emulation.
- [x] Scrolling the map during a build when reaching the map edges (*not* the window edges as when using the mouse). The closer to the map edge, the faster the scroll (maximum speed will be the scroll speed set in the settings).
- [x] Scroll decay when touch scrolling the map.
- [x] Mouse wheel emulation.
- [x] Improve military touch handling.

### 3. Technical stuff:

- This code makes heavy usage of the recently created `map_point` structure. I'm not sure I'm using it properly, but it gets the job done and reduces the number of function parameters.
- The `touch` structure is being passed by value. It was a small `struct` when I started, but it's getting bigger fast. I still like to directly copy it by value, but if it gets too big or if you think it's better, I'll start using `const touch *`.
- This code makes it possible to cancel a pending construction such as a row of houses or a road. That feature does not exist Caesar 3, but then again the game doesn't have touch support! However if you want it is now easy to add a "cancel construction" feature by right-clicking with the mouse if that feels appropriate.
- Given how the code evolved, it is now possible to merge `touch` and `mouse` structure to a single `pointer_input` structure (the name may be subject to change, maybe `pointing_device`?). Here is how I would suggest such a merge:
To distinguish between touch and mouse events, a `pointer_input` struct member named `is_touch` would be used.
Mouse buttons `left` and `right` would be replaced with the names `first` (representing the left button or the first active touch) and `second` (representing the right button or the second active touch) respectively. `went_down`, `is_down` and `went_up` would be replaced with `was_pressed`, `is_pressed` and `was_released`.
Mouse event handling would be changed to provide as much information to the `pointer_input` structure as the `touch` struct currently does.
Meta-information would be available via the functions `pointer_input_was_tap(button)`, `pointer_input_was_double_tap(button)` and `pointer_input_was_drag(button)`.
All current references to `mouse` would be changed to `pointer_input`. For example, `const mouse *m = mouse_get()` would become `const pointer_input *p = pointer_input_get()`.
And more things I as I remember or find useful.
I'll probably do this in a separate PR, as the necessary code changes will be substancial.

### 4. Suggestions?

- Even though using military units works, right now there is no visual feedback to show whether a legion is selected or not as there is no mouse cursor when touch is used. What could be a good replacement?
- The same problem exists with the clear tool. That may be fixed by some solution from #103 though.

Apart from bugfixing and code improvements, this is ready for merging after rebasing everything to a single commit.

Suggestions are still welcome!

### DIFF
--- a/src/city/view.c
+++ b/src/city/view.c
@@ -220,7 +220,7 @@ void city_view_get_selected_tile_pixels(int *x_pixels, int *y_pixels)
     *y_pixels = data.selected_tile.y_pixels;
 }
 
-int city_view_pixels_to_grid_offset(int x_pixels, int y_pixels)
+int city_view_pixels_to_view_tile(int x_pixels, int y_pixels, view_tile *tile)
 {
     if (x_pixels < data.viewport.x ||
             x_pixels >= data.viewport.x + data.viewport.width_pixels ||
@@ -255,9 +255,19 @@ int city_view_pixels_to_grid_offset(int x_pixels, int y_pixels)
         data.selected_tile.x_pixels -= 30;
     }
     data.selected_tile.y_pixels = data.viewport.y + 15 * y_view_offset - 15; // TODO why -1?
-    int x_view = data.camera.x + x_view_offset;
-    int y_view = data.camera.y + y_view_offset;
-    int grid_offset = view_to_grid_offset_lookup[x_view][y_view];
+
+    tile->x = data.camera.x + x_view_offset;
+    tile->y = data.camera.y + y_view_offset;
+    return 1;
+}
+
+int city_view_pixels_to_grid_offset(int x_pixels, int y_pixels)
+{
+    view_tile view;
+    if (!city_view_pixels_to_view_tile(x_pixels, y_pixels, &view)) {
+        return 0;
+    }
+    int grid_offset = view_to_grid_offset_lookup[view.x][view.y];
     return grid_offset < 0 ? 0 : grid_offset;
 }
 
--- a/src/graphics/arrow_button.c
+++ b/src/graphics/arrow_button.c
@@ -9,6 +9,9 @@ static const int REPEATS[] = {
     1, 0, 1, 0, 1, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 0
 };
 
+static const time_millis REPEAT_MILLIS = 30;
+static const unsigned int BUTTON_PRESSED_FRAMES = 3;
+
 void arrow_buttons_draw(int x, int y, arrow_button *buttons, int num_buttons)
 {
     for (int i = 0; i < num_buttons; i++) {
@@ -39,7 +42,7 @@ int arrow_buttons_handle_mouse(const mouse *m, int x, int y, arrow_button *butto
 
     time_millis curr_time = time_get_millis();
     int should_repeat = 0;
-    if (curr_time - last_time >= 30) {
+    if (curr_time - last_time >= REPEAT_MILLIS) {
         should_repeat = 1;
         last_time = curr_time;
     }
@@ -60,13 +63,13 @@ int arrow_buttons_handle_mouse(const mouse *m, int x, int y, arrow_button *butto
     }
     arrow_button *btn = &buttons[button_id -1];
     if (m->left.went_down) {
-        btn->pressed = 3;
+        btn->pressed = BUTTON_PRESSED_FRAMES;
         btn->repeats = 0;
         btn->left_click_handler(btn->parameter1, btn->parameter2);
         return button_id;
     }
     if (m->left.is_down) {
-        btn->pressed = 3;
+        btn->pressed = BUTTON_PRESSED_FRAMES;
         if (should_repeat) {
             btn->repeats++;
             if (btn->repeats < 48) {
--- a/src/graphics/tooltip.c
+++ b/src/graphics/tooltip.c
@@ -16,7 +16,8 @@
 
 #include <stdlib.h>
 
-#define DEFAULT_TEXT_GROUP 68
+static const int DEFAULT_TEXT_GROUP = 68;
+static const time_millis TOOLTIP_DELAY_MILLIS = 150;
 
 static time_millis last_update = 0;
 static uint8_t overlay_string[1000];
@@ -45,7 +46,7 @@ static int should_draw_tooltip(tooltip_context* c)
         reset_timer();
         return 0;
     }
-    if (time_get_millis() - last_update < 150) { // delay drawing by 150 ms
+    if (time_get_millis() - last_update < TOOLTIP_DELAY_MILLIS) { // delay drawing tooltip
         return 0;
     }
     return 1;
@@ -250,6 +251,10 @@ void tooltip_invalidate(void)
 
 void tooltip_handle(const mouse *m, void (*func)(tooltip_context *))
 {
+    if (m->is_touch && !m->left.is_down) {
+        reset_timer();
+        return;
+    }
     tooltip_context context = {m->x, m->y, 0, 0, 0, 0, 0, 0};
     context.text_group = DEFAULT_TEXT_GROUP;
     if (setting_tooltips() && func) {
--- a/src/input/scroll.c
+++ b/src/input/scroll.c
@@ -5,7 +5,12 @@
 #include "game/settings.h"
 #include "graphics/screen.h"
 
-#define SCROLL_BORDER 5
+#include <math.h>
+
+static const int SCROLL_BORDER = 5;
+static const int SCROLL_DECAY_MULTIPLIER = 250;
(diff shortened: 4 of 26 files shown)


mirh:
@devnoname120 you might be interested

crudelios:
This PR is now feature complete.

I would still welcome suggestions, especially for city construction.

crudelios:
Apart from some eventual bugs, this is now ready to merge.

src/platform/julius.c
```diff
@@ -324,6 +343,9 @@ static int init_sdl(void)
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Could not initialize SDL: %s", SDL_GetError());
         return 0;
     }
+#if SDL_VERSION_ATLEAST(2, 0, 3)
```
bvschaik: Does not compile on 2.0.3: min sdl version required is 2.0.4

bvschaik:
I tried this out on my mac with touchpad and it's definitely not compatible: the mouse position is off by a random amount vertically, clicks are going bonkers, map drags around when I just move the pointer.
I can guess why: touches are interpreted both as touches and as mouse clicks. Next up: how to solve this?

I'll review the code lightly, but we need to get this touchpad issue fixed first.

src/graphics/arrow_button.c
```diff
@@ -9,6 +9,9 @@ static const int REPEATS[] = {
     1, 0, 1, 0, 1, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 0
 };
 
+static const time_millis REPEAT_TIME = 30;
```
bvschaik: Minor: rename to `REPEAT_MILLIS` to indicate which time unit we use

src/graphics/tooltip.c
```diff
@@ -16,7 +16,8 @@
 
 #include <stdlib.h>
 
-#define DEFAULT_TEXT_GROUP 68
+static const int DEFAULT_TEXT_GROUP = 68;
+static const time_millis TOOLTIP_DELAY = 150;
```
bvschaik: Minor: rename to `TOOLTIP_DELAY_MILLIS`

src/input/scroll.c
```diff
@@ -113,6 +97,190 @@ int scroll_get_direction(const mouse *m)
...
+    data.speed.last_position = position;
+    data.speed.decaying = 0;
+}
+
+static void find_scroll_speed(map_point position)
+{
+    int current_time = time_get_millis();
+    int time_delta = current_time - data.speed.last_time;
+
+    int position_delta;
```
bvschaik: Variable definition can be moved to assignment, below the if statement (we're not C89 anymore :D )

src/input/scroll.h
```diff
@@ -1,12 +1,23 @@
 #ifndef INPUT_SCROLL_H
 #define INPUT_SCROLL_H
 
+#include "input/touch.h"
```
bvschaik: Alphabetical order please.

src/window/empire.c
```diff
@@ -407,13 +409,16 @@ static void draw_foreground(void)
...
+}
+
 static void determine_selected_object(const mouse *m)
 {
-    if (!m->left.went_down) {
-        return;
-    }
-    if (m->x < data.x_min + 16 || m->x >= data.x_max - 16 ||
-        m->y < data.y_min + 16 || m->y >= data.y_max - 120) {
+    if (!((!m->is_touch && m->left.went_down ) || (m->is_touch && m->left.went_up && !data.finished_scroll)) || is_outside_map(m->x, m->y)) {
```
bvschaik: Just split this out to multiple if-statements or make it a small function. Multiple parentheses and not-operators make me nervous ;)

crudelios:
> I tried this out on my mac with touchpad and it's definitely not compatible: the mouse position is off by a random amount vertically, clicks are going bonkers, map drags around when I just move the pointer.
> I can guess why: touches are interpreted both as touches and as mouse clicks. Next up: how to solve this?
> 
> I'll review the code lightly, but we need to get this touchpad issue fixed first.

Someone reported the same problem in a forum somewhere and the provided solution was simple: disable touch for mac, as there is no mac with touch devices.

What do you think?

src/platform/julius.c
crudelios: Oops, I misinterpreted the SDL manual.

crudelios:
@bvschaik I fixed all the reported issues and attempted to sort out the trackpad problem. I'm not sure it's a proper solution but SDL should make a distinction between Apple's trackpad and a proper touch device!

src/city/view.c
```diff
@@ -220,13 +220,14 @@ void city_view_get_selected_tile_pixels(int *x_pixels, int *y_pixels)
     *y_pixels = data.selected_tile.y_pixels;
 }
 
-int city_view_pixels_to_grid_offset(int x_pixels, int y_pixels)
+map_point city_view_pixels_to_map_point(int x_pixels, int y_pixels)
```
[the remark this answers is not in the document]
crudelios: > I prefer passing return structs as out parameter. Returning it copies the entire struct onto the stack, which may be fine for small things like this map_point, but I'd rather have one approach.

If that's the case, I'm guessing I should pass `const touch *` instead of the `touch` structure in touch-related functions, right?

src/platform/touch.c
```diff
@@ -0,0 +1,54 @@
...
+#include "game/settings.h"
+#include "input/touch.h"
+
+static SDL_FingerID touch_id[MAX_ACTIVE_TOUCHES];
+
+/* TODO perhaps this could be reused in other files? */
+static SDL_Rect get_screen_size(void)
+{
+    SDL_Rect screen_size = { 0 };
+    if (setting_fullscreen()) {
```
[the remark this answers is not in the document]
crudelios: Duh!

src/window/empire.c
crudelios: This was an unneeded mess.

crudelios:
Apparently the next version of SDL will have a function called `SDL_GetTouchDeviceType` which will allow us to distinguish between touchscreens and trackpads.

The code has already been commited ([mercurial](https://hg.libsdl.org/SDL/rev/eb60e952b13fcd5fd00bafdaf1f209a5d3fb9d73)).

Until the new SDL version rolls out, I see no other option but to either disable touch events alltogether for macOS, or disable events for the first touch device, hoping it will be the trackpad. Also, this will always  be a problem for older SDL versions.

bvschaik:
The issue with the Mac trackpad acting as touch device is solved 🎉

crudelios:
Just saw this now. I will apply all the requested changes.

> The issue with the Mac trackpad acting as touch device is solved 🎉

Nice! The changes were hackish but I'm glad they worked!

crudelios:
All the requested changes were implemented.

PR #110, opened 2019-03-02
Playstation Vita port

devnoname120:
Here are changes allowing to run Julius on a [Playstation Vita device](https://en.wikipedia.org/wiki/PlayStation_Vita).

Screenshots:

Notes:
- Lots of commits, because I tried different approaches (e.g. SDL for rendering, then switched to a native Vita library for performance reasons). I can squash them once this pull request is reviewed.
- @rsn887 did a pull request on top of my changes in order to improve input handling (https://github.com/devnoname120/julius/pull/1). This pull request will be done on https://github.com/bvschaik/julius instead when mine is merged.
- Performances are pretty good, the game is playable even in big cities. The videos are laggy however.
- I had to create a wrapper on top of `fopen()` because Julius relies a lot on `chdir()`/`getcwd()` which are not supported on the Vita.
- It should be possible to automatically compile using Travis with the [vitasdk](https://vitasdk.org), there is also a [Docker container](https://hub.docker.com/r/gnuton/vitasdk-docker). However, it will **not** be possible to run automated tests since there is no reliable PS Vita emulator yet. A physical PS Vita running a special program receiving unit tests could theoretically be done but it's a lot of work and a maintenance burden.


src/game/settings.c
```diff
@@ -40,9 +40,9 @@ static struct {
 static void load_default_settings(void)
 {
     data.fullscreen = 1;
-    data.window_width = 800;
-    data.window_height = 600;
-    
+    data.window_width = 960;
```
bvschaik: I don't like changing the default resolution for everyone. Can we make this vita-specific?

src/graphics/video.c
```diff
@@ -73,16 +74,19 @@ static const unsigned char *next_audio_frame(int *out_len)
 
 static int load_smk_video(const char *filename)
 {
-    data.video.s = smk_open_file(filename, SMK_MODE_DISK);
+    FILE *fp = file_open(filename, "rb");
+    data.video.s = smk_open_filepointer(fp, SMK_MODE_DISK);
     if (!data.video.s) {
+        if (fp != NULL)
```
bvschaik: Add curly braces please, same for further on in the file.

src/game/settings.c
devnoname120: Good catch! It was a temporary thing while porting Julius and I forgot to remove it.

Issue #100, opened 2019-02-13
Bundle as macOS app


shirakaba:
It sounds like bundling as a .app is necessary for DPI awareness: #57 #98. It would also be nice for user-friendliness (i.e. it's the necessary format that a 'release' build would take).

Just gathering together some research on how we might approach bundling this.

# Produce an Xcode project

This is what I'm familiar with (when making Cocoa + Swift apps for macOS). However, the fact that we have a build process (based on `cmake`, no less) means I'm out of my comfort zone here; I don't know how I'd configure an Xcode project to read header/include paths from a CMakeLists file (although I know of examples in the React Native core).

* [Setting up SDL on mac OS X Lion with Xcode 4](https://stackoverflow.com/questions/8221151/setting-up-sdl-on-mac-os-x-lion-with-xcode-4)
* [The SDL official docs – see bottom section of page](https://github.com/SDL-mirror/SDL/blob/master/docs/README-macosx.md)

I see that `src/platform/julius.c` is the entry-point of the app. I imagine we'd run `main()` only after presenting a dialogue to the user allowing them to specify the location of the Caesar III files (or only after loading a preference that they've saved from previous use of such a dialogue).

# Do it all by shell commands

This would be ideal due to its simplicity, but we'd still need to provide some way to allow the user to choose the location of their Caesar III files.

## Example from the libSDL forums

Here are suggestions from [a conversation](https://forums.libsdl.org/viewtopic.php?p=51999) on the LibSDL forums (all quoted from here on):

---

To elaborate, if you're not using Xcode, you can write the whole process up into your Makefile. Basically, create the app bundle yourself. I've copied a few excerpts from my makefile below: 

```sh
OUTPUTPATH=/Build/Release/ 
BUNDLE=$(OUTPUTPATH)/$(PROGRAM).app 
mkdir -p $(BUNDLE)/Contents 
mkdir -p $(BUNDLE)/Contents/MacOS 
mkdir -p $(BUNDLE)/Contents/Resources 
mkdir -p $(BUNDLE)/Contents/Frameworks
```

Copy frameworks like SDL: 

```sh
cp -RH /Library/Frameworks/SDL2.framework $(BUNDLE)/Contents/Frameworks/ 
rm -fr $(BUNDLE)/Contents/Frameworks/SDL2.framework/Versions/A/Headers/ 
rm -fr $(BUNDLE)/Contents/Frameworks/SDL2.framework/Headers 
```

[...]

PR #111, opened 2019-03-02
updated CMakeLists.txt for macOS bundle creation

superzazu:
Hello,

This PR adds the possibility of building a .app bundle on macOS. It uses the CMake [MACOSX_BUNDLE](https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_BUNDLE.html) feature. If you want to test it: `mkdir build && cd build && cmake .. && make`.

Somes notes:

- most of the work is in the CMakeLists.txt, where I set some variables that will populate the bundle's Info.plist (like the bundle name/version/identifier/icon path)
- I added a bash script that auto-creates the macOS icon file (icns format) from the file "julius_256.png". This script is automatically called by CMake on build, and the ICNS file is copied into the bundle.
- while calling "make install", CMake will automatically copy the needed libraries into the "Frameworks" folder in the bundle, and link the binary against them. This makes the bundle distributable (see https://cmake.org/cmake/help/latest/module/BundleUtilities.html)
- the custom `Info.plist.in` is the default `MacOSXBundleInfo.plist.in` file from CMake 3.13.4, with one additional key "NSHighResolutionCapable" that fixes the DPI problem on Retina screens
- I arbitrarily set the minimum OSX deployment version to 10.10, but I suppose we can deploy for older systems (I had a working binary while setting this value at 10.6). I'm not sure which value is best.
- I have modified the argument handling in the function `main()` because the Finder automatically passes an argument "-psn_0_XXXXXX" as first argument when double-clicking on the bundle (see https://hg.libsdl.org/SDL/file/c005c49beaa9/test/testdropfile.c#l47).
- in the `pre_init()` function, I also added a `chdir()` to the [SDL_GetBasePath](https://wiki.libsdl.org/SDL_GetBasePath): this function returns the directory where the application is run from. On macOS/iOS, this function returns the path of the bundle's "Resources" directory (if the binary is inside a bundle). Which means that the user can just drag and drop Caesar III data inside the Resources folder (alongside the "julius.icns" file), and the bundle will work in an autonomous way. Note that it is still possible to pass a custom path: `./julius.app/Contents/MacOS/julius custom/path/to/caesarIII`.

This PR is related to #100 . Feel free to ask if you have any questions or code modification ideas.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -404,6 +404,11 @@ set(WINDOW_FILES
     ${PROJECT_SOURCE_DIR}/src/window/victory_video.c
 )
 
+set(MACOSX_FILES "")
+if(APPLE)
+    set(MACOSX_FILES ${PROJECT_SOURCE_DIR}/res/julius.icns)
+endif()
+
 add_executable(julius
     WIN32
     ${PLATFORM_FILES}
@@ -423,6 +428,7 @@ add_executable(julius
     ${WINDOW_FILES}
     ${SMK_FILES}
     ${PROJECT_SOURCE_DIR}/res/julius.rc
+    ${MACOSX_FILES}
 )
 
 find_package(SDL2 REQUIRED)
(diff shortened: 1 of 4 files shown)


bvschaik:
First of all thanks for figuring this out!

On my machine, `make install` fails because the optional dependencies for SDL_Mixer (such as vorbis, flac, ...) that are not required to play .wav files, are missing. Do you know if there's any way we can just exclude them?

And the SDL2.0.0 build now fails, maybe just put an `if (SDL_VERSION_ATLEAST(2,0,1)) { ... }` around the `SDL_GetBasePath` code. We're going to ship the macOS build with the newest SDL libraries anyway so it'll never be a problem on macOS.

superzazu:
> And the SDL2.0.0 build now fails, maybe just put an if (SDL_VERSION_ATLEAST(2,0,1)) { ... } around the SDL_GetBasePath code. We're going to ship the macOS build with the newest SDL libraries anyway so it'll never be a problem on macOS.

Thank you for the response. I added a check using SDL_VERSION_ATLEAST like you suggested, the builds now pass successfully.

> On my machine, make install fails because the optional dependencies for SDL_Mixer (such as vorbis, flac, ...) that are not required to play .wav files, are missing.

I reproduced your problem successfully on my machine by using official SDL frameworks instead of dylibs from Homebrew.

The problem was that during "install", CMake could not access (for some reason) the dependencies of SDL2_mixer, which are in `SDL2_mixer.framework/Versions/A/Frameworks/`.

I pushed a new commit with a fix, and I'm now able to build julius on macOS using both SDL frameworks and Homebrew dylibs.

> Do you know if there's any way we can just exclude them?

During the "fixup_bundle" process, CMake runs `otool` on the julius binary (and its dependencies) to check which librairies are linked against it. After this, CMake copies the librairies inside the bundle and updates the julius binary using `install_name_tool` to link against the librairies inside the bundle.

Official SDL_mixer framework is shipped with support for FLAC/Ogg/modplug/mp3/opus, and as a consequence, the library itself is linked against those dependencies. So I'm afraid we can't easily remove the dependencies of SDL_mixer.

The only obvious answer would be to re-compile SDL_mixer without FLAC/Ogg/modplug/mp3/opus support, and use it instead of the official framework file (note: this solution actually works, I've tested it).

CMakeLists.txt
```diff
@@ -481,7 +534,9 @@ if (VITA_BUILD)
     )
 else()
     target_link_libraries (${SHORT_NAME} ${SDL2_LIBRARY} ${SDL2_MIXER_LIBRARY})
-    install(TARGETS ${SHORT_NAME} RUNTIME DESTINATION bin)
+    if(NOT APPLE)
+        install(TARGETS ${SHORT_NAME} RUNTIME DESTINATION bin)
+    endif()
```
superzazu: On macOS, install() is already called once before, so we do not need to call it again. (calling it twice actually causes an error on a call to "install_name_tool")

bvschaik:
I didn't realize the FLAC/etc libraries were bundled in SDL_Mixer. It works properly on my machine using the framework bundles now, thanks!

PR #113, opened 2019-03-06
Multiple control enhancements on Playstation Vita

rsn8887:
- Bluetooth keyboard and mouse support
- On-screen keyboard by pressing start, useful to enter player name etc.
- Analog joystick mouse control via left stick
- Touchpad-style mouse control with gestures (short tap for left click, etc.)
- Map right analog stick to cursor keys
- Map L/R to mousebuttons, and square/triangle to page up/down
- Select toggles indirect (touchpad) / direct (pointer jumps to finger) touch

This PR is a result of several discussions with @devnoname120 and code-reviews by @devnoname120 that had already happened before the Vita port was upstreamed.


devnoname120:
Note: Instead of using the raw inputs directly from a system library, it gets the input from the SDL and generates corresponding SDL key / mouse events.

crudelios:
This could benefit from my touch PR if direct touch mode is used. For example, dragging a finger through the screen would scroll the city.

However, as-is, both PRs are incompatible with each other as both poll SDL for touch events.

rsn8887:
@crudelios: Your touch support will be better than mine and also work on Vita. (Vita version of SDL generates touch events just like any other touch device, nothing special is needed).

I will close this and make a new PR minus my touch stuff.

PR #119, opened 2019-03-07
Fix sound effects not working in Windows.

crudelios:
This happens because Windows does not have a working [`SDL_RWFromFP`](https://wiki.libsdl.org/SDL_RWFromFP) function (see https://wiki.libsdl.org/SDL_RWFromFP#Remarks).

### DIFF
--- a/src/platform/sound_device.c
+++ b/src/platform/sound_device.c
@@ -65,10 +65,14 @@ void sound_device_init_channels(int num_channels, char filenames[][CHANNEL_FILEN
         Mix_AllocateChannels(num_channels);
         for (int i = 0; i < num_channels; i++) {
             if (filenames[i][0]) {
+#ifdef __vita__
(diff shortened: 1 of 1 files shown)


src/platform/sound_device.c
```diff
@@ -65,10 +65,14 @@ void sound_device_init_channels(int num_channels, char filenames[][CHANNEL_FILEN
         Mix_AllocateChannels(num_channels);
         for (int i = 0; i < num_channels; i++) {
             if (filenames[i][0]) {
+#ifdef __vita__
                 FILE *fp = file_open(filenames[i], "rb");
                 SDL_RWops *sdl_fp = SDL_RWFromFP(fp, SDL_TRUE);
```
devnoname120:
```suggestion
                 // Does not work on Windows
                 SDL_RWops *sdl_fp = SDL_RWFromFP(fp, SDL_TRUE);
```

You get the idea.

PR #121, opened 2019-03-11
README: Improve CI status readibility

devnoname120:
I'm not sure we actually want the second commit since it uses [an external service](https://github.com/bjfish/travis-matrix-badges).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: c
 
 matrix:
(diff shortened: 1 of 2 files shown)


crudelios:
I like the new table. We can always revert back to the old style if the external dependecy goes offline.

Also, this makes the `Downloading latest Windows build` section of the readme unnecessary (but it could be kept to allow less tech-friendly users to easily understand what they need to download).

PR #122, opened 2019-03-12
added a simple message box when data files are not found at launch

superzazu:
On some systems (such as macOS/Windows), users don't have access to the console while running Julius. This means the user won't see logs such as `INFO: Julius requires the original files from Caesar 3 to run [...]`. This PR adds a simple message box using [SDL_ShowSimpleMessageBox](https://wiki.libsdl.org/SDL_ShowSimpleMessageBox) when the C3 files are not found at launch.

### DIFF
--- a/src/platform/julius.c
+++ b/src/platform/julius.c
@@ -365,6 +365,12 @@ static int pre_init(const char *custom_data_dir)
     SDL_Log("Move the Julius executable to the directory containing an existing Caesar 3 installation, or run:");
     SDL_Log("julius path-to-c3-directory");
 
+    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR,
+        "Julius requires the original files from Caesar 3 to run",
+        "Move the Julius executable to the directory containing an existing "
+        "Caesar 3 installation, or run: julius path-to-c3-directory",
+        NULL);
+
     return 0;
 }
 


devnoname120:
Same issue on the Vita, I'm not sure that `SDL_ShowSimpleMessageBox` actually works on this platform so I'm going to take a look.

devnoname120:
I can confirm that on Vita it doesn't display, which is fine as least it doesn't make it worse.

crudelios:
@superzazu, @devnoname120: Now that #111 has been merged I will improve my [`folder_dialogs`](https://github.com/bvschaik/julius/compare/master...crudelios:folder-dialogs) branch. This uses both a "Select Folder" Window and a MessageBox and uses an external library that makes it compatible with Windows, Linux and Mac OS, however it's not compatible with Android/Vita/Switch.

I know nothing about coding for Vita, but having a common interface to display the dialogs would be great.

rsn8887:
Is there comprehensive documentation of tinyfiledialog library anywhere? I see in the header that it claims to support SDL. If it draws the dialogs using SDL it should work on Vita and Switch...

rsn8887:
Ah never mind, I see it opens either native dialogs or console (text mode) dialogs, neither of which Switch or Vita have.

PR #123, opened 2019-03-13
Fixes the random game crashes when closing a message with video playing.

crudelios:
The crash was cause by a race condition between `smk_close` and `Mix_HookMusic`.

This was tested and fixed by calling `sleep(1)` during memory freeing in `smk_close`. Using `sleep` to simulate a thread interruption, the old code always crashed when the message was closed while the video was playing. The new code doesn't crash.

Hopefully this improves things for Vita/Switch as well.


bvschaik:
Nice one!

PR #130, opened 2019-03-20
Implement text display limits to text input fields.

crudelios:
If the text written to an input is larger than its input field width, only the text that fits to the width of the input will be shown.

Which portion of the text will be shown depends on the cursor position.

If there is text input to the left or right that is not shown, a small arrow will appear to indicate there is further text.

So far this is only used in the file load/save/delete dialog.

This fixes the visual glitches when using long file names for cities.

Video (gif):

capture

### DIFF
--- a/src/graphics/graphics.c
+++ b/src/graphics/graphics.c
@@ -234,24 +234,6 @@ void graphics_draw_horizontal_line(int x1, int x2, int y, color_t color)
     }
 }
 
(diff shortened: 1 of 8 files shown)


bvschaik:
The later games also do this, but without the arrows. Can we just skip them as well? They look a bit out of place.

crudelios:
Ok, got rid of the arrows!

PR #131, opened 2019-03-21
Nintendo Switch: Add Travis-CI

devnoname120:
This will make sure that Julius stays compilable.


rsn8887:
Wow thanks! I had no time, thanks for finishing this.

PR #136, opened 2019-04-01
Improve window positioning on resize and set minimum window size to 640x480.

### LINKED ISSUE #97: In-game options to resize screen place window in top-left corner
When using F7/F8/F9, or the Options -> Display menu to resize the window in windowed mode, the window jumps to the top-left corner of the screen.

**To Reproduce**
1. Start the game in windowed mode, with the window not in the top-left corner of the screen
2. Resize using F7/F8/F9, or using Options -> Display
3. Note that the window jumps to (0,0)

**Expected behavior**
Window stays at the position it's currently at.

Seen on macOS and Raspberry Pi.

crudelios:
Fixes #97.

Also fixes an assortment of other bugs that I found while working with the DPI branch.

### DIFF
--- a/src/platform/screen.c
+++ b/src/platform/screen.c
@@ -1,21 +1,27 @@
 #include "platform/screen.h"
 
-#include "SDL.h"
-
 #include "game/settings.h"
 #include "graphics/graphics.h"
 #include "graphics/screen.h"
+
+#include "SDL.h"
 
 static struct {
     SDL_Window *window;
     SDL_Renderer *renderer;
     SDL_Texture *texture;
-} SDL = {0, 0, 0};
+} SDL;
 
 static struct {
     int x;
     int y;
-} window_pos;
+    int centered;
+} window_pos = { 0, 0, 1 };
+
+static struct {
+    const int WIDTH;
+    const int HEIGHT;
+} MINIMUM = { 640, 480 };
 
 int platform_screen_create(const char *title)
 {
@@ -38,8 +44,9 @@ int platform_screen_create(const char *title)
         flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
     }
     SDL.window = SDL_CreateWindow(title,
-        SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
+        SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED,
         width, height, flags);
+
     if (!SDL.window) {
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Unable to create window: %s", SDL_GetError());
         return 0;
@@ -59,6 +66,7 @@ int platform_screen_create(const char *title)
         SDL_SetWindowGrab(SDL.window, SDL_TRUE);
     }
 #endif
+    SDL_SetWindowMinimumSize(SDL.window, MINIMUM.WIDTH, MINIMUM.HEIGHT);
     return platform_screen_resize(width, height);
 }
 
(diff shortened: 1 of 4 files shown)


src/platform/screen.c
```diff
@@ -1,22 +1,78 @@
...
+    int title_height = get_title_height();
+
+    if ((!window_pos.x && !window_pos.y) || !window_pos.width || !window_pos.height) {
+        platform_screen_center_window();
+    } else if (width != window_pos.width || height != window_pos.height) {
+        window_pos.x = window_pos.x + (window_pos.width / 2) - (width / 2);
+        window_pos.y = window_pos.y + (window_pos.height / 2) - (height / 2);
+    }
+
+    SDL_Rect monitor_coordinates;
```
bvschaik: Why do we need this complex calculation?

crudelios: I'm guessing you're talking about the calculation with the `monitor_coordinates`. This calculation is to make sure the new window position is within the bounds of all existing monitors.

bvschaik: Do we really want to do that ourselves? Why not delegate it to the window manager of the operating system? On macOS you have to consider the dock and global menu bar, on Windows the task bar can be on every edge, and Linux is even more flexible...

crudelios: The window manager unfortunately does nothing to prevent a window from being positioned outside the monitor range.

Regarding the task bar or the global menu bar, SDL has a function called [`SDL_GetDisplayUsableBounds`](https://wiki.libsdl.org/SDL_GetDisplayUsableBounds) that gets the valid bounds taking those into account, but it's only compatible with SDL 2.0.5.

bvschaik: I've played around with this on my mac and sometimes the window moves in unexpected ways when hitting F7/8/9 repeatedly. I haven't been able to find a pattern.
I've now fixed the positioning issue with just a simple `SDL_GetWindowPos` before resizing the window. This keeps the top left corner of the window in the same place. While I think that moving the window so the contents stay in the same place is nice, I don't think the complexity of the code (and weird bugs it may cause on different platforms) justifies it.

src/platform/screen.c
```diff
@@ -1,22 +1,78 @@
...
 #include "graphics/graphics.h"
 #include "graphics/screen.h"
 
+#include "SDL.h"
+
 static struct {
     SDL_Window *window;
     SDL_Renderer *renderer;
     SDL_Texture *texture;
-} SDL = {0, 0, 0};
```
bvschaik: Why did you remove the initialization? If the compiler does not clear the memory (some don't...), this will cause crashes when initializing the texture for the first time.

crudelios:
Aren't static members always initialized to zero? I get your point, I think it was a good ideabto keep the window centered.

How about a compromise: if the window was altready centered, recenter it by using `SDL_POSITION_CENTER`. The would involve no calculations, only a bool to let us know whether the window was centered before or not.

bvschaik:
> Aren't static members always initialized to zero?

I looked it up and you're right. I had a bad experience once with an old Visual Studio compiler (spent half a day tracking a weird bug caused by uninitialised memory, one day before the project was due..) and ever since I'm maybe too cautious about not initialising things :wink: 

> How about a compromise: if the window was altready centered and is resized via hotkey/options window, recenter it by using `SDL_POSITION_CENTER`. The would involve no calculations, only a bool to let us know whether the window was centered before or not.

Sound good! :+1:

crudelios:
So I remembered the reason for the complex code.

As the code was extracted from the DPI PR, it served another purpose besides positioning the window on resize. It also allowed proper window positioning when moving the window across monitors with different DPI settings.

This means the complex code will always be needed for the DPI branch. However for normal user resizes via hotkey or the window settings, a simpler code that keeps the current `x` and `y` position can be used.

bvschaik:
> This means the complex code will always be needed for the DPI branch. However for normal user resizes via hotkey or the window settings, a simpler code that keeps the current `x` and `y` position can be used.

Can you remove it from this one though so I can merge it without the added complexity?

Issue #137, opened 2019-04-02
Encoding issues with user-entered strings


bvschaik:
Using text in the save game/load game dialogs with characters that are not in ASCII results in incorrect filenames. Due to different operating systems having different encodings for filenames, saved games may either have invalid characters in the filename (Linux), wrong characters if the "code page" of the game differs from the system (Windows), or the game simply refuses to create invalid filenames (MacOS).

You also cannot type any characters other than plain ASCII symbols in the dialogs.

**To Reproduce**
Easiest to reproduce:
1. Load the Russian version in any of the three operating systems
2. Go to New career
3. Save the game with the default filename

**Expected behavior**
Game is saved with the right filename

crudelios:
Based on the info on this bug and on #104, perhaps we should indeed internally convert strings to UTF-8? That would make things easier with filenames.

bvschaik:
Actually, I was thinking of doing it the other way around, since we cannot display all of UTF-8 with the internal font graphics. The only places where using `uint8_t` for strings is a problem is when loading scenarios, and loading and saving `.sav` files. Changing all other places is going to be a major pain. Plus, filenames are encoded differently per platform, so that's also the place where you want to do conversion.

nwtour:
>Encoding issues with user-entered strings

Now in Jullius disabled non-english text input with comment: 'only accept non-multibyte chars' in platform/keyboard_input.c

It makes sense to solve this issue only if you enable input otherwise it is easier to simply replace 'Default savefile template' by any English text (full disable non-english filenames)

bvschaik:
> It makes sense to solve this issue only if you enable input otherwise it is easier to simply replace 'Default savefile template' by any English text (full disable non-english filenames)

That's indeed my intention. I started working on this yesterday evening, and I'm learning a lot about encodings and filesystems...

bvschaik:
@nwtour Entering Russian text should work properly now. One question for you: when I enable a Russian keyboard on my operating system, the standard hotkeys (P = pause, W = water overlay, etc) do not work because those keys now produce Cyrillic letters. How do you normally do that when playing games? Use a Latin keyboard?
And if I want to support Russian keyboards, what would be good Cyrillic alternatives for the following hotkeys?

- P: pause/unpause game
- F: show fire overlay
- D: show damage overlay
- C: show crime overlay
- T: show problem overlay
- W: show water overlay
- L: cycle through legions

nwtour:
[...]

PR #139, opened 2019-04-06
Support for Azerty keyboards in game

ysard:
First of all I want to thank you for this huge job. I am in admiration for this huge project to revive a game of our childhood.

The game on Linux is much cheaper in CPU resources than using Wine, and the free license (not just open source: p) allows us to contribute to this work!

--

I'm using an Azerty keyboard and I'm having some issues with keyboard shortcuts in play.

Shortcuts involving the following keys do not work:
[,], 1,2,3,4,5,7,8,9,0

Key 6 opens the ADVISOR_FINANCIAL tab instead of ADVISOR_POPULATION.

This is normal since the numbers are entered with the shift key as a modifier.

In this pull request I assume that codes for letters are supposed to be layout dependent, so keycodes are still used.
However the keys involving the numbers are dependent on modifiers but they are still in the same place on a keyboard, so I propose to use their scancodes rather than their keycodes; they are independent of the layout used and refer to the physical location of a key on the keyboard.
This also allows a big improvement for Azerty keyboard users over the original game as they no longer need to press shift to open the city advisor boards.

I also take the opportunity to restore default shortcuts to adjust the game speed with 'Altgr' and '[' or ']'.

Also, `keysym.sym` only returns the ASCII code of the lowercase letter, so you it is not needed to test the capital letter as well, so I use the variables provided by the SDL directly.

I am largely inspired by these 2 pages:
https://wiki.libsdl.org/SDL_Keycode
https://wiki.libsdl.org/MigrationGuide

If the pull request does not match you, maybe it will give you some tips to solve these international keyboards problems.

Thanks again for this great job!

### DIFF
--- a/src/input/hotkey.c
+++ b/src/input/hotkey.c
@@ -163,10 +163,10 @@ void hotkey_character(int c)
                 break;
 
             // Azerty keyboards need alt gr for these keys
-            case '[':
+            case '[': case '5':
                 change_game_speed(1);
                 break;
-            case ']':
+            case ']': case '-':
                 change_game_speed(0);
                 break;
         }
--- a/src/input/hotkey.c
+++ b/src/input/hotkey.c
@@ -150,94 +150,85 @@ void hotkey_character(int c)
 {
     if (data.alt_down) {
         switch (c) {
-            case SDLK_x: //'X'
+            case 'X': case 'x':
                 hotkey_esc();
                 break;
-            case SDLK_k: //'K'
+            case 'K': case 'k':
                 cheat_init_or_invasion();
-            case SDLK_c: //'C'
+            case 'C': case 'c':
                 cheat_money();
                 break;
-            case SDLK_v: //'V'
+            case 'V': case 'v':
                 cheat_victory();
-                break;
-
-            // Azerty keyboards need alt gr for these keys
-            case SDL_SCANCODE_5: //'['
-                change_game_speed(1);
-                break;
-            case SDL_SCANCODE_MINUS: //']'
-                change_game_speed(0);
                 break;
         }
         return;
     }
-
     switch (c) {
-        case SDLK_LEFTBRACKET: //'['
+        case '[':
             change_game_speed(1);
             break;
-        case SDLK_RIGHTBRACKET: //']'
+        case ']':
             change_game_speed(0);
             break;
-        case SDL_SCANCODE_SPACE: //' '
+        case ' ':
             toggle_overlay();
             break;
-        case SDLK_p: //'P'
+        case 'P': case 'p':
             toggle_pause();
             break;
-        case SDLK_f: //'F'
+        case 'F': case 'f':
             show_overlay(OVERLAY_FIRE);
             break;
-        case SDLK_d: //'D'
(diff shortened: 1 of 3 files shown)


src/input/hotkey.c
```diff
@@ -150,85 +150,94 @@ void hotkey_character(int c)
 {
     if (data.alt_down) {
         switch (c) {
-            case 'X': case 'x':
+            case SDLK_x: //'X'
```
bvschaik: Please, *no* SDL-related code/constants in anything other than the `platform` directory.

src/input/hotkey.h
```diff
@@ -1,6 +1,9 @@
 #ifndef INPUT_HOTKEY_H
 #define INPUT_HOTKEY_H
 
+#include "SDL.h"
```
bvschaik: Please, *no* SDL-related code/constants in anything other than the `platform` directory: if we decide to switch to something other than SDL, I want to keep the main codebase unchanged.

ysard:
I understood the reason of this structure, consequently I modified the code.
I tested on layout fr and us.
It's a bit hacky but using scancodes for the shortcuts of the panel city advisor seems inevitable.
However, the rest of the shortcuts will not work on keyboards having for example a Russian layout...

bvschaik:
Thanks for updating. I'll have a look at it later this week; I have little time to review things during the week. I also encountered the problem with Russian layout during my testing for #137. I wonder how other games handle keyboard hotkeys in different keyboard layouts.

src/platform/keyboard_input.c
```diff
@@ -89,12 +89,61 @@ void platform_handle_key_down(SDL_KeyboardEvent *event)
...
+                if (event->keysym.sym >= 97 && event->keysym.sym <= 122)
+                    hotkey_character(event->keysym.sym);
             }
             break;
     }
+
+    // Send scancodes for non-layout dependent keys
+    switch (event->keysym.scancode) {
+        case SDL_SCANCODE_1:
+            hotkey_character('1');
```
nwtour: SDL function SDL_GetScancodeName() return one symbol from english layot (less hardcode):

case SDL_SCANCODE_1:
case SDL_SCANCODE_2:
case SDL_SCANCODE_3:
case SDL_SCANCODE_EQUALS:
    // hotkey_character(1 or 2 or 3 or '=') 
   hotkey_character(SDL_GetScancodeName(event->keysym.scancode));
   break;

ysard:
@nwtour
Well seen, it is not stable across platforms and discouraged by the doc (https://wiki.libsdl.org/SDL_GetScancodeName) but for these keys in particular it should be ok and it avoids spaghetti code.

I just pushed the modification. It's up to bvschaik to say if it's ok, I can do a rebase if needed.

@bvschaik
Take your time, it's time consuming to manage this type of project...

[Here on gamedev.stackexchange](https://gamedev.stackexchange.com/questions/53064/how-can-i-handle-different-keyboard-layouts) there may be tips for supporting keyboards.
Several approaches. Scancodes are more robust but keycodes are inevitable when it comes to managing user inputs.

PR #142, opened 2019-04-17
add two Vita-specific touch modes, select button toggles modes

rsn8887:
press select to toggle between touch modes
1) touchpad mode (tap to click, new default)
2) jump-to-finger mode (no tap to click)
3) original julius touch control mode

### DIFF
--- a/src/platform/vita/vita_touch.c
+++ b/src/platform/vita/vita_touch.c
@@ -1,0 +1,397 @@
+#include <psp2/kernel/processmgr.h>
+#include "vita.h"
+#include "vita_touch.h"
+#include <math.h>
+
+#define NO_TOUCH -1 // finger id setting if finger is not touching the screen
+
+static void init_touch(void);
+static void preprocess_events(SDL_Event *event);
+static void preprocess_finger_down(SDL_Event *event);
+static void preprocess_finger_up(SDL_Event *event);
+static void preprocess_finger_motion(SDL_Event *event);
+static void set_mouse_button_event(SDL_Event *event, uint32_t type, uint8_t button, int32_t x, int32_t y);
+static void set_mouse_motion_event(SDL_Event *event, int32_t x, int32_t y, int32_t xrel, int32_t yrel);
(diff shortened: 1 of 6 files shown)


src/platform/vita/vita_touch.c
```diff
@@ -0,0 +1,400 @@
...
+static void preprocess_events(SDL_Event *event);
+static void preprocess_finger_down(SDL_Event *event);
+static void preprocess_finger_up(SDL_Event *event);
+static void preprocess_finger_motion(SDL_Event *event);
+static void set_mouse_button_event(SDL_Event *event, uint32_t type, uint8_t button, int32_t x, int32_t y);
+static void set_mouse_motion_event(SDL_Event *event, int32_t x, int32_t y, int32_t xrel, int32_t yrel);
+static Uint32 empty_event_type = 0;
+
+static int vita_rear_touch = 0; // always disable rear_touch for now
+extern int last_mouse_x;
```
bvschaik: Instead of having this implicit coupling: put these `extern` definitions in `vita_input.h` so it's clear that the variables are provided by `vita_input.c`.
Preferably I'd rather not see any `extern` variables at all, or have them grouped together in a struct, so there's only one name exposed.

rsn8887: What if I use vita.h to declare these variables? Since vita.h is included by vita_input.c and by vita_touch.c, this would mean I could get rid of the extern keyword.

bvschaik: Also OK with me.

rsn8887: fixed.

rsn8887: I couldn't get rid of the extern keyword though. I realized it had to be a declaration (not a definition) in the header file.

rsn8887: See here: https://github.com/bvschaik/julius/blob/9287910c0a185326e39147338bae27c54632cd65/src/platform/vita/vita.h#L17-L19

PR #146, opened 2019-04-19
Reduce the audio lag

rsn8887:
Without this change, the audio lag on Switch platform is almost
one full second, much larger than on other platforms.

This commit reduces the audio buffer which reduces the audio lag
without affecting sound quality. It reflects the audio buffer
setting used by almost all other homebrew software on the Switch.


src/platform/sound_device.c
```diff
@@ -12,7 +12,12 @@
 #define AUDIO_RATE 22050
 #define AUDIO_FORMAT AUDIO_S16
 #define AUDIO_CHANNELS 2
+
+#if defined(__SWITCH__)
+#define AUDIO_BUFFERS 512
```
devnoname120: This leads to the question: should we apply this change to all the platforms?

rsn8887: it shouldn’t hurt unless it leads to crackling or stuttering.

rsn8887: On Vita: The audio lag is much less than on the Switch to begin with. If I apply this change, and click on the icons very fast, I get audio drop-outs. 

So, I think it is best to make this Switch-only for now.

devnoname120: I'd suggest adding a comment to explain why it's set to 512 for the Switch.

bvschaik: For Mac, 4096 is also too large, 512 gives a way better experience, so I would like to lower it for all platforms. How does 1024 sound on Vita? Maybe that's a suitable default.

rsn8887: 1024 seems to work great even on Vita! I updated this PR.

PR #153, opened 2019-05-26
Add Vita/Switch instructions to readme.md

rsn8887:
With this PR, I am trying to provide control information and installation instructions in the readme.md for the Vita/Switch versions. These instructions were previously provided by devnoname120's and my release notes on our own forks. Since Vita/Switch users will now use this official repo to download their binaries, I thought it would be nice to provide instructions.

A review and any suggestions would be welcome.

Mainly, is this something we want to put in the readme.md? Or maybe somewhere else?

If you want to see the markdown formatted diff, just click on "view rich diff" in the top right corner of the readme.md diff view.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ After cloning the repo, run the following commands:
 
 This results in a `julius` executable.
 
+To build the Vita or Switch version, use `cmake .. -DVITA_BUILD=ON` or `cmake .. -DSWITCH_BUILD=ON` instead of `cmake ..`.
+
 ## Running the game
 
 After downloading the Windows build or compiling the game yourself, the game data should be in the same directory as in which the binary is run or it can be run as `julius /path/to/game-data`.
@@ -45,4 +47,4 @@ The easiest way to run the game is to copy either the compiled executable (in ca
 
 Note that you must have permission to write in the game data directory as the saves will be stored there.
 
-See [Running](RUNNING.md) for further instructions.
+See [Running](RUNNING.md) for further instructions, including instructions for Linux, Vita, and Switch.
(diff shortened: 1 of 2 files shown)


bvschaik:
Thanks! I was about to ask for instructions for Vita/Switch.

> Mainly, is this something we want to put in the readme.md? Or maybe somewhere else?

I think the controls section (and maybe also the installing section) should go in a separate file/page. Perhaps we should introduce a `doc` directory for stuff like this.

rsn8887:
I moved the Vita/Switch installation and control instructions from README.MD to RUNNING.MD. I think a doc directory might not be necessary, since there's only two files (README and RUNNING).

README.md
```diff
@@ -37,6 +37,8 @@ After cloning the repo, run the following commands:
 
 This results in a `julius` executable.
 
+To build the Vita or Switch version, use `cmake .. -DVITA_BUILD=On` or `cmake .. -DSWITCH_BUILD=On` instead of `cmake ..`.
```
devnoname120:
```suggestion
To build the Vita or Switch version, use `cmake .. -DVITA_BUILD=ON` or `cmake .. -DSWITCH_BUILD=ON` instead of `cmake ..`.
```

RUNNING.md
```diff
@@ -42,3 +42,78 @@ Another option is to get the game files by installing Caesar 3 using [WINE](http
 1. Run Julius with the path where the game is installed:
 
         $ julius path-to-c3-directory
+
+## Vita
+
+1. Install the .vpk like any other homebrew.
```
devnoname120: I don't think that we can assume that everyone knows about vita jailbreaking and installing third-party application. Maybe we should give some pointers?

rsn8887: Thanks for the feedback, I added a link to instructions for Vita and Switch.

README.md
rsn8887: fixed

rsn8887:
This is ready to merge

rsn8887:
Fixed another mistake (mentioned ScummVM instead of Julius due to a copy/paste error). Now it should be ready.

PR #154, opened 2019-06-03
TinyDialogs & C3 Folder selection fix for MinGW-w64 compiler

jdmillard:
The changes leave standard MinGW build process intact.

My IDE automatically cleared a bunch of trailing whitespace. This clutters the diff, so please view with `hide whitespace changes` enabled. EDIT: if these changes are contributing to the concern about diverging from upstream, I can edit the history on my fork to leave in the trailing spaces (upon request)

### DIFF
--- a/ext/tinyfiledialogs/tinyfiledialogs.c
+++ b/ext/tinyfiledialogs/tinyfiledialogs.c
@@ -104,7 +104,7 @@ misrepresented as being the original software.
 #include "tinyfiledialogs.h"
 
 #ifdef USE_TINYFILEDIALOGS
-/* #define TINYFD_NOLIB */ 
+/* #define TINYFD_NOLIB */
 
 #ifdef _WIN32
  #ifdef __BORLANDC__
@@ -347,7 +347,7 @@ static void replaceSubStr( char const * const aSource ,
         char const * p ;
         char const * lNewSubStr = "" ;
         size_t lOldSubLen = strlen( aOldSubStr ) ;
-        
+
         if ( ! aSource )
         {
                 * aoDestination = '\0' ;
@@ -484,7 +484,7 @@ static int dirExists(char const * const aDirPath)
         struct stat lInfo;
 
         if (!aDirPath || !strlen(aDirPath))
-                return 0;       
+                return 0;
         if (stat(aDirPath, &lInfo) != 0)
                 return 0;
         else if ( tinyfd_winUtf8 )
@@ -625,7 +625,7 @@ static void RGB2HexW(
                 {
                         /* wprintf(L"aoResultHexRGB %s\n", aoResultHexRGB); */
                         swprintf(aoResultHexRGB,
-#if !defined(__BORLANDC__) && !defined(__TINYC__) && !defined(__MINGW32__) && ( !defined(__GNUC__) || (__GNUC__) >= 5 )
+#if !defined(__BORLANDC__) && !defined(__TINYC__) && ( !defined(__MINGW32__) || defined(__MINGW64__) ) && ( !defined(__GNUC__) || (__GNUC__) >= 5 )
                                 8,
 #endif
                         L"#%02hhx%02hhx%02hhx", aRGB[0], aRGB[1], aRGB[2]);
@@ -672,7 +672,7 @@ static int sizeMbcs(wchar_t const * const aMbcsString)
 static wchar_t * utf8to16(char const * const aUtf8string)
 {
         wchar_t * lUtf16string ;
-        int lSize = sizeUtf16(aUtf8string);     
+        int lSize = sizeUtf16(aUtf8string);
         lUtf16string = (wchar_t *) malloc( lSize * sizeof(wchar_t) );
         lSize = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS,
                                         aUtf8string, -1, lUtf16string, lSize);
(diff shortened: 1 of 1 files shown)


jdmillard:
* If running MinGW (32bit standard) this will evaluate to false (REQUIREMENT for CI)
* If running MinGW-w64, this will evaluate to true (REQUIREMENT, for me at least)
* If running any other compiler not listed in the logic, this will evaluate to true (REQUIREMENT for general support)


For general record keeping: quoting my comments from Gitter conversation (with contextual edits):

> I have figured this out. The problem seems to be due to the fact that `swprintf` is handled differently by different compilers. We see in this logic:
> ```
>                         swprintf(aoResultHexRGB,
> #if !defined(__BORLANDC__) && !defined(__TINYC__) &&  !defined(__MINGW32__) && ( !defined(__GNUC__) || (__GNUC__) >= 5 )
>                                 8,
> #endif
>                         L"#%02hhx%02hhx%02hhx", aRGB[0], aRGB[1], aRGB[2]);
> ```
> That all compilers are expecting this extra integer, except for BorlandC, TinyC, and MinGW.
> 
> The problem is that I use `mingw-w64` for 64bit builds (linking against 64bit SDL2). Anyways, MinGW-w64 appears to use the standard implementation of `swprintf` and wants that integer as the second argument just like most other compilers. This logic fails because MinGW-w64 comiler defines both `__MINGW32__` and `__MINGW64__` so the `#if` evaluates to false and it doesn't get that integer it needs.
> 
> I have run some compile tests with some dummy logic to confirm all of the above. I also got this to compile and run perfectly by slightly adjusting the `#if` logic to account for they maybe-case of MinGW-w64. I suspect that, if I made this change correctly, this will not break the standard 32-bit MinGW builds.
> 
> If interested, the change looks like this:
> ```
>                         swprintf(aoResultHexRGB,
> #if !defined(__BORLANDC__) && !defined(__TINYC__) && ( !defined(__MINGW32__) || defined(__MINGW64__) ) && ( !defined(__GNUC__) || (__GNUC__) >= 5 )
>                                 8,
> #endif
>                         L"#%02hhx%02hhx%02hhx", aRGB[0], aRGB[1], aRGB[2]);
> ```

devnoname120:
I think that instead of diverting more and more from upstream, such changes should be contributed directly to the main repo: https://sourceforge.net/projects/tinyfiledialogs/

jdmillard:
> I think that instead of diverting more and more from upstream, such changes should be contributed directly to the main repo: https://sourceforge.net/projects/tinyfiledialogs/

Fair enough. However, by _not_ making [tinyfiledialogs](https://sourceforge.net/p/tinyfiledialogs/code/ci/master/tree/) a submodule (or an official dependency in the build process), it tends to send the message that we don't mind diverting a little. I could see an argument either way depending on where the priority is placed.

devnoname120:
- Submodules are [awkward to use](http://web.archive.org/web/20171101202911/http://somethingsinistral.net:80/blog/git-submodules-are-probably-not-the-answer/) and they can break a project.
- Subtrees break rebasing.
- [Subrepo](https://github.com/ingydotnet/git-subrepo) seems to be the best solution but it's not in the official Git client and it doesn't seem to be maintained anymore.

All these solutions are suboptimal, hence why the dependencies are put a well-compartmentalized directory. This is a simple solution that fits our needs.

bvschaik:
My thoughts:
- try to get this merged upstream, since it's a compiler fix
- submodules are not the way to go (I have worked with them before)
- the project itself also recommends copy-paste of the files

I'm off for a business trip - don't expect any activity from me for the next few days.

jdmillard:
The only time submodules have ever caused a problem for me was when they were being used for something they are not intended for. But I totally get it - y'all have your preferences and we can debate that another day :)

On a philosophical level, I agree with the principle that the fixes should go upstream. But I'm seeing mixed messages for a few reasons:
* We haven't configured this repo to track anything upstream, so it would be some sort of a cherry pick or manual copy+paste to pull in new features.
* We've committed the source directly to this repo (albeit by suggestion) - which means we plan to maintain and "own" that block of code.
* We've made changes and diverged from upstream to make that code work for our needs already.

I love this project and you all have done some excellent work here. But I'm not planning on creating a SourceForge account any time soon (even if there was an easy way for julius to track the new upstream changes in tinydialogs). Shall we cut bait and not support the MinGW-w64 compiler? It's really not the end of the world and we can remain united in our coding quests by our undying love for Ceasar3/Julius!

devnoname120:
Well I think that we should try to upstream changes as much as possible because keeping the internal 'fork' up-to-date will be annoying since we'll have to deal with merge conflicts.

Any reasons you're not planning to create an account? It takes 1 minute and you can use a temporary email service.

bvschaik:
I created a SF account, and created a [PR](https://sourceforge.net/p/tinyfiledialogs/code/merge-requests/5/) with some of the changes we made already, mainly to improve the code quality. Considering the track record for merge requests of tinyfiledialogs (4 created in the past 2 years, 0 merged, 0 rejected), I don't think these will be merged anytime soon -- if ever.
As an aside: SourceForge is definitely not a nice site to work with - forking the tinyfiledialogs repo took 25 minutes?!

While comparing the code, I also discovered that we apparently added the `!defined(__MINGW32__)` check ourselves - @crudelios is that correct? It's not in the original repo. In that case, I think we can safely merge this change.

PR #160, opened 2019-07-23
Fix switch compilation with latest toolchain

rsn8887:
Newest toolchain version of SDL_mixer requires linking with opusfile and opus.


rsn8887:
This is needed in preparation before I can do a Docker file update. The Switch toolchain recently got a new version update with lots of fixes etc.

rsn8887:
Ok I updated the docker file. So it will build with the latest toolchain version now on Switch.

PR #170, opened 2019-10-13
Bintray: list files in downloads

devnoname120:
https://github.com/travis-ci/docs-travis-ci-com/commit/d1e519856298a86e2476bc573c64bc0c0fb01390#diff-e68563f233b86867c9515e9a55ee9bfeR60


devnoname120:
Note that this doesn't work yet. I suspect that this change has not been added to `travis-ci.org` yet. It should work when it's merged though.

PR #171, opened 2019-10-19
AppImage: Add support and integrate in Travis-CI

devnoname120:
The generated image was tested on Archlinux, it seems to work properly. You can download it from here in order to test it locally: https://bintray.com/devnoname120/julius/linux-unstable/1.1.0-20191019-0b6cf74#files

Notes:
- [AppImage supports incremental updates](https://docs.appimage.org/packaging-guide/updates.html). Might be interesting to add this in the future.
- Since my initial attempt, it seems that a page was written about [how to build AppImage using Travis-CI](https://docs.appimage.org/packaging-guide/travis.html#). This pull request is not based on this. It also seems a bit complex and I'm not sure that we can trust these scripts not to break. `linuxdeployqt` seems promising but I'm not sure if it's mature enough yet.
- Although the changes are relatively small, it was **a lot of work** to actually get all the steps just right so that it works. A huge amount of trial and error was necessary. AppImage doesn't have enough human resources to polish the process and it shows.


src/platform/julius.c
```diff
@@ -405,6 +406,21 @@ static int pre_init(const char *custom_data_dir)
         return game_pre_init();
     }
 
+
+    char *prev = NULL;
+    prev = getcwd(prev, 0);
+
+    // AppImage support
+    if (chdir("../share/julius") == 0) {
```
devnoname120: This should be relative to the executable's location, not the current working directory.

bvschaik: I doubt we will ever have the game data shipped with the AppImage because it's copyrighted. And if we provide alternative graphics, I think whoever holds the rights to C3 nowadays will come after us because we then basically provide "their" game for free. Right now that's not the case because Julius requires the original game files.

Besides that: the AppImage is run from... where? `/usr/bin`? `/usr/local/bin`? Those are not locations where your average user has write access, and we need that for the savegames, user-created maps and settings files.

.ci_scripts/run_build.sh
```diff
@@ -11,8 +11,17 @@ case "$BUILD_TARGET" in
 	cd build && make && make test && make install
 	echo "Creating disk image"
 	hdiutil create -volname Julius -srcfolder julius.app -ov -format UDZO julius.dmg
+	cd ..
```
bvschaik: Why do we want to cd back up here?

.travis.yml
```diff
@@ -1,4 +1,7 @@
 language: c
+compiler: gcc
+sudo: require
+dist: xenial # Oldest still-supported LTS right now against which the AppImage is built. Update this to the next LTS when support drops.
```
bvschaik: Don't these three settings belong to the AppImage build only?

.travis.yml
```diff
@@ -31,13 +40,22 @@ cache:
 
 install:
   - ./.ci_scripts/start_docker.sh
-  - ./.ci_scripts/install_sdl.sh
+  - if [ "$BUILD_TARGET" != "appimage" ];
+    then
+      ./.ci_scripts/install_sdl.sh;
+    else
+      sudo apt-get -y install libgl1-mesa-dev libsdl2-dev libsdl2-mixer-dev;
```
bvschaik: Can we move this logic to the `install_sdl.sh` script?

.travis.yml
```diff
@@ -31,13 +40,22 @@ cache:
...
+    fi;
 
 before_script:
   - ./.ci_scripts/run_cmake.sh
 
 script:
   - ./.ci_scripts/run_build.sh
+  - if [ "$BUILD_TARGET" == "appimage" ];
+    then
+      ./.ci_scripts/package_appimage.sh;
```
bvschaik: Move this to `run_build.sh`: it's comparable to creating the `.app` file for macOS, which also lives in that script.

.ci_scripts/run_build.sh
```diff
@@ -11,8 +11,17 @@ case "$BUILD_TARGET" in
...
 	;;
 *)
 	cd build && make && make test
+	if [ "$BUILD_TARGET" == "appimage" ];
```
bvschaik: Create an `appimage` case instead of putting this `if` inside the "catch-all" `*)`. Yes, that's duplication of the `cd build && make && make test` line, but the code is clearer.

.ci_scripts/run_cmake.sh
```diff
@@ -10,6 +10,9 @@ case "$BUILD_TARGET" in
 "mac")
 	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release ..
 	;;
+"appimage")
+	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr ..
```
bvschaik: Do we need the `CMAKE_INSTALL_PREFIX` when we tell `make install` to use a different `DESTDIR`?

src/platform/julius.c
devnoname120: Even though we won't ship the game data, users can add the game data to the AppImage *after the fact*. Especially useful in the future in combination with the in-place updates that AppImage supports.

They run from a SquashFS mounted by FUSE. Even if FUSE is not available on your OS (it's installed by default on most distributions), you can anyway run the AppImage using the `--appimage-extract-and-execute`. The mount point is usually something like `/dev/loopX`. See [the doc](https://docs.appimage.org/user-guide/run-appimages.html) for more information.

The data is stored in `$HOME` by default. This can be changed and AppImage's [can be run portably](https://docs.appimage.org/user-guide/run-appimages.html) if necessary.

.travis.yml
devnoname120: Can `sudo: require` be actually moved in the matrix?
Feel free to do these changes directly on my branch.

.ci_scripts/run_cmake.sh
devnoname120: `CMAKE_INSTALL_PREFIX` is the base directory for installing the game, which is prepended to `DESTDIR` which contains the path where we store the base directory and the children. I suppose it could be all moved to `DESTDIR` but it would be less idiomatic. See https://cmake.org/cmake/help/v3.16/envvar/DESTDIR.html

See https://cmake.org/cmake/help/v3.13/envvar/DESTDIR.html

.ci_scripts/run_build.sh
devnoname120: Otherwise `package_appimage.sh` would start in another directory than this script, which would be confusing.
See https://github.com/bvschaik/julius/blob/6f4c95ed9bcaf419788ac5d675f9f1fb1a8eaa43/.travis.yml#L54-L58

devnoname120:
Note that since this AppImage is built on Ubuntu Xenial, the resulting image is expected to run on distributions from mid-2016 and newer, as well as older distributions that still get updated and thus have the required packages.

.ci_scripts/run_build.sh
bvschaik: No. The directory is changed for the .sh file only. After that's done, you end up in the same directory as where you started, regardless of what the .sh script has done to the working directory.

.ci_scripts/run_build.sh
```diff
@@ -11,8 +11,17 @@ case "$BUILD_TARGET" in
...
+	;;
+"appimage")
+	cd build && make && make test
+	make DESTDIR=AppDir install
+	cd ..
+	./.ci_scripts/package_appimage.sh
 	;;
 *)
 	cd build && make && make test
+	make install
```
bvschaik: Why run `make install` here? That usually requires root / admin privileges.

.travis.yml
bvschaik: Logic has been moved, remove the `if` statement please :)

src/platform/julius.c
bvschaik: > Even though we won't ship the game data, users can add the game data to the AppImage _after the fact._
 
Can you explain how? I toyed around with the appimage: it's basically running in /usr/bin from a sandboxed Linux system, which is read-only. It doesn't make sense to try `../share/julius` as data directory because we need write access.

devnoname120: > it's basically running in /usr/bin 

This sounds impossible to me. Are you sure it was not running from a mounted location?

> Can you explain how?

Yes, just extract the content with `--appimage-extract`, add the game data, then repack it using `./linuxdeploy-x86_64.AppImage --appdir AppDir --output appimage`.

There might be a way to directly add data to an existing AppImage but I don't have enough experience with this tool so I don't know how one would do this.

.ci_scripts/run_build.sh
devnoname120: I stand corrected then, it can be removed.

.ci_scripts/run_build.sh
devnoname120: Added by mistake while refactoring, then refactoring back. Note that I think it's a good thing in order to make sure that this command actually works. It's off-topic for this pull request however.

.travis.yml
devnoname120: Nice catch!

src/platform/julius.c
bvschaik: I did some tests:

- All data in the virtual file system of the app image is *read-only*. If you pack the whole C3 directory with data files etc and use that, then you won't be able to save the game, save settings, save created scenarios, etc. Basically useless.
- The working directory of an appimage executable works exactly the same as for a normal executable: when started from the command-line, it's the directory where you run the command from. When started from a file manager: the directory of the appimage executable. "../share/julius" is _never_ going to resolve to a path inside the appimage file system.

Therefore: please remove this check for "../share/julius".

devnoname120:
Note that I've made sure that the resulting appimage was conform using https://github.com/TheAssassin/appimagelint.

devnoname120:
@bvschaik I think I've addressed all your comments. Is there anything else you want me to change?

src/platform/julius.c
```diff
@@ -18,6 +18,7 @@
 #include "tinyfiledialogs/tinyfiledialogs.h"
 
 #include <signal.h>
+#include <stddef.h>
```
bvschaik: Include is no longer necessary.

devnoname120: I swear I had removed it, thanks for noticing.

Issue #173, opened 2019-11-12
Windows 10 tiny cursor


dcnieho:
The in game cursor is tiny. This might well be because I'm on a dell laptop with 4K screen, display scaling in Windows is set to 250%. Is that honored taken into account by the game? I'd be happy to try and fix this one if you can't point me in the direction of where in the code the cursor is drawn.

This is on up to date Windows 10 Enterprise, running yesterday's nightly build.

bvschaik:
That might indeed be related to the 4K screen and display scaling. Do the interface elements and city also look too small or are they OK?
Since I don't have a 4K screen (or Windows) this will be hard to debug... I'll check whether Julius supports display scaling on Windows; on my Mac it does, but that's a completely different implementation under the hood.

dcnieho:
Whoops, i replied to these questions in the other item (#172) as well ;)

dcnieho:
we may as well close this as discussion is taking place in #178

PR #178, opened 2019-11-16
take input argument for cursor scale

dcnieho:
solves #173 for me. Lots of code taken from @crudelios's DPI aware branch. Logic: if we can't know dpi without making our exe dip-aware, the only option left is to give the user control over cursor size.


dcnieho:
@crudelios: this PR doesn't build for Vita and Switch. Since i do not have one, i cannot test and prefer to thus not touch those files. Do you know what changed might be needed? It seems a Switch (the one i looked at) takes and SDL texture instead of an SDL cursor. So I'll try using the code for desktop systems to create cursor surface and then convert to texture. sdl_createtexturefromsurface

crudelios:
I created the dpi-aware branch before Vita and Switch support were added, but I can have a look.

dcnieho:
Thanks!

crudelios:
No problem!

Also, an alternative to the user setting the scale is to [run the code suggested on this post](https://stackoverflow.com/a/36864741/10664000) when the game boots. I don't like the fact that it is Windows-specific, but at least it would be just during boot.

dcnieho:
I'd be happy to play with that. That it is Windows-only is perhaps not an
issue if Windows is the only platform with the problem? Or does this play
for HiDPI on other platforms too?
Where would one hide such platform-specific code in the Julius code-base?
All the code in the platform subfolder seems to be generic as well ;)
What I'd personally like here is to use code like this to figure out a good
default for the user, but allow the command line parameter to
override (since we have it already, and smart guesses aren't always right
guesses).

Cheers!
Dee

crudelios:
I'll test on a Linux VM first, even though the result might be different due to the mouse virtualization.

crudelios:
As I thought, running things in virtualbox doesn't allow me to understand if the cursor scaling is working.

However, what I found out is that on Linux, in higi DPI mode, the game doesn't scale at all! Here's an image:

c3_linux_hidpi

crudelios:
As for the compiler errors you're getting, try the following:

In both `platform/vita/cursor.c` and `platform/vita/cursor.c`. change `system_init_cursors` to:

```
void system_init_cursors(cursor_scale cur_scale)
{
    for (int i = 0; i < CURSOR_MAX; i++) {
        cursors[i] = init_cursor(input_cursor_data(i), cur_scale);
    }
    system_set_cursor(CURSOR_ARROW);
}
```

That should fix the problem.

dcnieho:
@crudelios: thanks! It shouldn't be that simple though, right? as the cursor struct passed in now has different member variables and the cursor needs to be created in a different way? 
@devnoname120 and @rsn8887, are you able to have a look at this?

crudelios:
I think if you send the default scale to `system_init_cursors` things will work properly, but I'm not entirely sure. Try changing the code to see if it at least compiles.

dcnieho:
It wouldn't work, init_cursor in both those functions expect a 32x32 bit of
data in the cursor struct, but with the changed code, that is no longer the
case.

crudelios:
You're right, it won't work, but I don't think it would be too hard to fix. Right now `init_cursor` in both `vita` and `switch` have hardcoded `32x32` pixel sizes, I think replacing those with `c->width` and `c->height`, while making sure only the smallest cursor is used, may do the trick, unless vita and switch texture sizes need to be a multiple of 2, in which case some sort of padding would need to be added.

A simpler fix would be to change the smallest cursor size to 32x32, in which case things should work out of the box.

bvschaik:
I'll have a look at the cursor this evening.

@crudelios For Linux, does this mean we also need a --display-scale parameter? I toyed around with [SDL_RenderSetLogicalSize](https://wiki.libsdl.org/SDL_RenderSetLogicalSize) yesterday and it's not too difficult to implement.

crudelios:
@bvschaik I think you're right, we should need a `--display-scale` paramater. Now that left me thinking, if the DPI-aware branch is changed to only be used at game boot (only scales the game when it loads, and not when the user changes the DPI or monitor during gameplay), we may achieve a similar result to the command line options without the need to use them. It would also not require some hackish code I added to support the DPI/monitor changes during gameplay.

This is because essencialy what my DPI-aware branch did was scale the game window (using, precisely, [`SDL_RenderSetLogicalSize`](https://wiki.libsdl.org/SDL_RenderSetLogicalSize)).

dcnieho:
@crudelios: Just to chime in, as I said previously, such "smartness"
is certainly desirable, but I'd like to keep a way to override it when
it fails on me. Having users save and restart the game when they
switch between monitors with different scaling seems like a small
price to pay, so good thinking!

dcnieho:
And just had to show, Julius running at 5K (never mind the city, i am relearning, clearly :P):
[image](https://postimg.cc/WDvZ4mkN).
Scaling on this monitor is 150%. But things work fine for me as well (except cursor size) on a 4K laptop screen with 250% scaling. Is that because if the app is not DPI aware, Windows does some scaling itself automatically? (dunno). So I guess we may have to look into the logical size thing for windows as well then, once it is declared DPI-aware. I'd be very happy to test!

crudelios:
Yes, Windows auto-scales a DPI-unaware app so it doesn't appear too small. That's why some apps have blurry text on Windows with high-DPI monitors, as the GUI is basically converted to an image and streched.

You can try compiling my old DPI-aware branch to test, just beware that some new features of julius will not work.

bvschaik:
Since this has been implemented in the hidpi branch (PR #183), I'm closing this.

PR #180, opened 2019-11-22
MP3 support.

crudelios:
This adds simple MP3 support for julius, allowing the replacement of the default music with the higher-quality version (for those few who have them).

Right now it searches for the following files:

```
"mp3/ROME1.mp3",
"mp3/ROME2.mp3",
"mp3/ROME3.mp3",
"mp3/ROME4.mp3",
"mp3/ROME5.mp3",
"mp3/Combat_Long.mp3",
"mp3/Combat_Short.mp3",
"mp3/setup.mp3"
```

You can either use the higher quality files or you can improve file quality by using Audacity, for example (I removed most of the hiss from `setup.wav`, albeit at the cost of some distortion in the music).

In order to use MP3 files, the library `libmpg123-0` must be added alongside `SDL_mixer`. This is possibly installed by default on Linux, but on Windows it is not. Therefore it is a new dependency. However, if `libmpg123-0` is not detected, Julius falls back to using the wav files.


devnoname120:
Why are the paths hardcoded? Why not always try to load with `.mp3` extension and if it doesn't exist fall back on `.wav`?

crudelios:
It just seemed like overkill to me to be manipulating strings just for 8 files...

src/platform/sound_device.c
```diff
@@ -127,7 +127,10 @@ void sound_device_play_music(const char *filename)
         #ifdef __vita__
         free(resolved_filename);
         #endif
+
+        return (music) ? 1 : 0;
```
bvschaik: Parentheses are not necessary.

src/sound/music.c
```diff
@@ -15,7 +15,8 @@ enum {
     TRACK_CITY_5 = 5,
     TRACK_COMBAT_SHORT = 6,
     TRACK_COMBAT_LONG = 7,
-    TRACK_INTRO = 8
+    TRACK_INTRO = 8,
+    TRACK_TOTAL = 9
```
bvschaik: For consistency with other enums, call this one `TRACK_MAX`.

crudelios:
> And for Windows: shouldn't we package the `libmpg123-0.dll` from the SDL_Mixer binaries for this to work?

Yes we should. Since I didn't know if you wanted to have an extra dependency, I just let Julius load the wavs if `libmpg123-0.dll` wasn't found.

bvschaik:
> > And for Windows: shouldn't we package the `libmpg123-0.dll` from the SDL_Mixer binaries for this to work?
> 
> Yes we should. Since I didn't know if you wanted to have an extra dependency, I just let Julius load the wavs if `libmpg123-0.dll` wasn't found.

Fair point. For the few percent of users that may use this feature, perhaps a "how to use mp3 files" documentation page should be written which also links to the mp3 dll for windows users.

crudelios:
I added it now, but let me know if you want me to remove it again!

When this is merged I will update the instructions in #184.

bvschaik:
I think it's best to have it as optional download, so please remove it. Sorry for the extra work!

crudelios:
No problem. I'm out for the weeked, so I'll do it on Monday!

bvschaik:
Just tested: if you don't have the DLL but do have MP3s, the game plays the WAV file instead. I was afraid it would crash without the DLL but it's all good.

I'll cherry-pick your other commits to master. I'm trying to catch up on all activity here before the weekend 😉

bvschaik:
Relevant changes have been cherry-picked to master in commit 0b78d7757ae06b5a6d788c893a2af93e376e5a08

crudelios:
>Just tested: if you don't have the DLL but do have MP3s, the game plays the WAV file instead. I was afraid it would crash without the DLL but it's all good.

Yeah I had that in place just in case ;)

> I'll cherry-pick your other commits to master. I'm trying to catch up on all activity here before the weekend 😉

Nice!

PR #183, opened 2019-11-24
Feature/hidpi

bvschaik:
Branch for hidpi fixes. Most of the code in PR #178 has been integrated.

Command line parameters `--display-scale` and `--cursor-scale` have been implemented.

Suggestions for code changes and test results are welcome.


src/platform/arguments.c
```diff
@@ -0,0 +1,95 @@
...
+}
+
+int platform_parse_arguments(int argc, char **argv, julius_args *output_args)
+{
+    int ok = 1;
+
+    // Set sensible defaults
+    output_args->data_directory = 0;
+    output_args->display_scale_percentage = 100;
+    output_args->display_scale_percentage = 100;
```
dcnieho: this should probably be output_args->cursor_scale_percentage?

src/platform/switch/cursor.c
```diff
@@ -15,8 +16,7 @@ extern struct {
 
 static SDL_Texture *init_cursor(const cursor *c)
 {
-
-    SDL_Texture *tex = SDL_CreateTexture(SDL.renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STREAMING, 32, 32);
+    SDL_Texture *tex = SDL_CreateTexture(SDL.renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STATIC, 32, 32);
```
dcnieho: the worry i have with this and the vita code is that the below is hardcoded to expect a 32*32 array of cursor data. Instead, e.g., the arrow cursor at scale 1 is 13*21

src/platform/arguments.c
crudelios: Yeah, if this is not properly initialized the cursor is always loaded at 2x scale for me without setting anything on the command line. It's looking great on my surface, but probably wrong for most people!

src/platform/julius.c
```diff
@@ -509,20 +513,12 @@ static void teardown(void)
...
-        // https://hg.libsdl.org/SDL/file/c005c49beaa9/test/testdropfile.c#l47
-        if (SDL_strncmp(argv[i], "-psn", 4) == 0) {
-            continue;
-        }
-
-        custom_data_dir = argv[i];
-        break;
+    julius_args args;
+    if (!platform_parse_arguments(argc, argv, &args)) {
+        return 1;
```
crudelios: Should julius really exit on invalid cmd arguments? How about logging a warning and proceeding to use default values?

crudelios:
For Vita and Switch, please use the following code otherwise either the cursor will be garbled or the app will crash on boot.

For `platform/switch/cursor.c`:

```
static SDL_Texture *init_cursor(const cursor *c)
{

    SDL_Texture *tex = SDL_CreateTexture(SDL.renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STATIC, 32, 32);

    uint32_t pixels[32 * 32] = { 0 };

    for (int i = 0; i < c->height; i++) {
        for (int j = 0; j < c->width; j++) {
            switch (c->data[i * c->width + j]) {
                case 'X':
                    pixels[i * 32 + j] =  0xFF000000; // Black
                    break;
                case '\'':
                    pixels[i * 32 + j] = 0xFFFFFFFF; // White
                    break;
                case ' ':
                    break; // Transparent is 0x00000000 which is the default value of the buffer
            }
        }
    }

    SDL_UpdateTexture(tex, NULL, pixels, 32 * 4);
    SDL_SetTextureBlendMode(tex, SDL_BLENDMODE_BLEND);

    return tex;
}
```

And for `platform/vita/cursor.c:`

```
static vita2d_texture *init_cursor(const cursor *c)
{
    vita2d_texture *tex = vita2d_create_empty_texture_format(32, 32, SCE_GXM_TEXTURE_FORMAT_U8U8U8U8_ARGB);
    color_t *cursor_buf = vita2d_texture_get_datap(tex);

    for (int i = 0; i < c->height; i++) {
        for (int j = 0; j < c->width; j++) {
            switch (c->data[i * c->width + j]) {
                case 'X':
                    cursor_buf[i * 32 + j] = COLOR_BLACK | 0xFFu << 24;
                    break;
                case '\'':
                    cursor_buf[i * 32 + j] = COLOR_WHITE | 0xFFu << 24;
                    break;
                case ' ':
                    break; // Transparent is 0x00000000 which is the default value of the buffer
            }
        }
    }

    return tex;
}
```

crudelios:
Scales properly on Linux with `--display-scale 2` and `--cursor-scale 1`!

c3_scaled

dcnieho:
I'll give it a test after the code issues have been addressed.

src/platform/arguments.c
bvschaik: Whoops, copy/paste error..

bvschaik:
Hmm, I did update the vita and switch code yesterday, looks like I forgot to push the commit. I'll fix those comments when I get home again this evening.

I fixed the copy/paste cursor_scale_percentage initialization error: Windows/Linux/Mac can be tested.

src/platform/julius.c
bvschaik: I was doubting this one. Command-line programs usually terminate when you call them with invalid arguments. On the other hand: Julius is not a command-line program and expected to launch from UI as well.

crudelios:
@bvschaik You probably didn't do anything wrong, I pasted that code as a comment to #178 so it was not actually in the files.

bvschaik:
@crudelios No, I definitely forgot to push a commit or two yesterday 😉 I was a bit sick yesterday evening but wanted to open the PR so you guys could take a look, and that's when I forgot to check that I had pushed everything to Github.

crudelios:
Haha alright, anyway it seems to be working perfectly now! I finally got a proper mouse cursor for my surface.

Also, you probably don't notice as you don't use high DPI, but the mouse cursors are in high definition and look really sharp, so that's a plus.

dcnieho:
I have not so great experience:
1. printf() outputs nothing for me when starting julius from command line ([googling](https://stackoverflow.com/questions/11068565/printf-is-not-printing-anything-to-output-c-sdl) seems to suggest it doesn't unless the application is built as a console type, but then you would have an extra window open. i don't have the stdout.txt file referred to in that thread).
2. i get an access violation on line 182 of graphics.c when i run with display-scale 200. at that scale, not the whole interface fits on my screen. when i then open the advisor panel all is fine until i hover over the advisor buttons on the bottom, that causes the crash.
I also have some suggestions about the argument parsing code, will add them in as code review

src/platform/arguments.c
```diff
@@ -0,0 +1,109 @@
+#include "arguments.h"
+
+#include "SDL.h"
+
+#include <stdio.h>
+
+#define CURSOR_SCALE_ERROR_MESSAGE "Option --cursor-scale must be followed by a scale value of 1, 1.5 or 2\n\n"
+#define DISPLAY_SCALE_ERROR_MESSAGE "Option --display-scale must be followed by a scale value between 0.5 and 5\n\n"
+
```
dcnieho: add:
#define UNKNOWN_OPTION_ERROR_MESSAGE "The option %s is not understood\n\n"

src/platform/arguments.c
```diff
@@ -0,0 +1,109 @@
...
+        // https://hg.libsdl.org/SDL/file/c005c49beaa9/test/testdropfile.c#l47
+        if (SDL_strncmp(argv[i], "-psn", 4) == 0) {
+            continue;
+        }
+        if (SDL_strcmp(argv[i], "--display-scale") == 0) {
+            if (i + 1 < argc) {
+                int percentage = parse_decimal_as_percentage(argv[i + 1]);
+                i++;
+                if (percentage < 50 || percentage > 500) {
+                    printf(DISPLAY_SCALE_ERROR_MESSAGE);
```
dcnieho: printf doesn't work for me. SDL_Log actually prints to console here (perhaps because not associated with a file yet?).
Platform: Windows 10 up to date.

src/platform/arguments.c
```diff
@@ -0,0 +1,109 @@
...
+                } else {
+                    printf(CURSOR_SCALE_ERROR_MESSAGE);
+                    ok = 0;
+                }
+            } else {
+                printf(CURSOR_SCALE_ERROR_MESSAGE);
+                ok = 0;
+            }
+        } else if (SDL_strcmp(argv[i], "--help") == 0) {
+            ok = 0;
```
dcnieho: add:
```
}
        else if (SDL_strncmp(argv[i], "--",2) == 0) {
            SDL_Log(UNKNOWN_OPTION_ERROR_MESSAGE, argv[i]);
            ok = 0;
```

src/platform/arguments.c
```diff
@@ -0,0 +1,109 @@
...
+        } else {
+            output_args->data_directory = argv[i];
+        }
+    }
```
dcnieho: add:

```
if (!ok) {
            break;
        }
```

crudelios:
> i get an access violation on line 182 of graphics.c when i run with display-scale 200. at that scale, not the whole interface fits on my screen. when i then open the advisor panel all is fine until i hover over the advisor buttons on the bottom, that causes the crash.

This crash only happens on fullscreen, but in my opinion on fullscreen the game should always fit in the monitor. Perhaps setting to fullscreen shouldn't be allowed when, if using scaled mode, the in-game resolution would be lower than 640x480?

src/platform/switch/cursor.c
bvschaik: Should be fixed now.

bvschaik:
I updated the call to `SDL_SetWindowMinimumSize` to also take the scale factor into account. But SDL just gives us the maximum window it's able to get for the display if the minimum requested size is larger than the display's resolution, which I think is fair behaviour. This happens both in fullscreen and in windowed mode, by the way.

However, that does mean that Julius should be able to handle resolutions smaller than 640x480, even though the user interface does not fit on that (that's the user's problem I guess). I have spent quite some effort to make it at least not crash on smaller window sizes, but that work has basically been removed when we implemented performance improvements like minimap caching.

src/platform/arguments.c
bvschaik: Alright, I'll revert it back to SDL_Log :)

src/platform/arguments.c
bvschaik: Why stop parsing arguments when one is wrong? Might as well catch as many errors as we can in one go.

src/platform/julius.c
bvschaik: I decided not to exit on invalid args and continue with the defaults.

src/platform/vita/cursor.c
```diff
@@ -11,17 +11,17 @@ vita2d_texture *current_cursor;
 
 static vita2d_texture *init_cursor(const cursor *c)
 {
-    vita2d_texture *tex = vita2d_create_empty_texture_format(32, 32, SCE_GXM_TEXTURE_FORMAT_U8U8U8U8_ARGB);
+    vita2d_texture *tex = vita2d_create_empty_texture_format(c->width, c->height, SCE_GXM_TEXTURE_FORMAT_U8U8U8U8_ARGB);
     color_t *cursor_buf = vita2d_texture_get_datap(tex);
```
crudelios: I'm not sure but sometimes textures need to be either divisable by 2 or a power of 2 in order to work, that's why I left them at 32x32 in size

crudelios:
> However, that does mean that Julius should be able to handle resolutions smaller than 640x480, even though the user interface does not fit on that (that's the user's problem I guess). I have spent quite some effort to make it at least not crash on smaller window sizes, but that work has basically been removed when we implemented performance improvements like minimap caching.

If I remember correctly, those crashes can be fixed by checking for out of bounds in `graphcis_save_to_buffer` and `graphics_copy_from_buffer`.

bvschaik:
Yup, that's exactly where I'm fixing stuff now. They go out of bounds on the minimap and the tooltip, exactly the places we optimized.

crudelios:
I was hoping having to check for an out of bounds wouldn't be needed (it's kinda hassly), but here we are 6 months later having to do it. Since it was me who started that optimization mess, let me know if you need anything!

bvschaik:
I'm re-using the clip code for the out of bounds checking. Crashes appear to be gone now, but the code could use another pair of eyes (and some testing; for ease of testing: comment out the `SDL_SetWindowMinimumSize` call).

And no worries about the optimization mess. Good thing I had prepared for this and we only need to change 2 functions.
A story in the same category: I spent a lot of time making the read/write saved game logic endian-aware. A programmer-colleague of mine responded to that with: "who the hell is still using a big endian computer??". Promptly I got a bug report about endianness by someone running on an old Mac, who discovered the one place that still missed a conversion :wink:

crudelios:
> and some testing; for ease of testing: comment out the SDL_SetWindowMinimumSize call

I tested and tested and the game failed to crash on me now. Great job!

> And no worries about the optimization mess. Good thing I had prepared for this and we only need to change 2 functions.

As a certified spaghetti coder, I have to say I am amazed with the clarity of your code. I've learned a lot from the way you code and I can definitely say it improved the way I code now as well.

src/platform/arguments.c
```diff
@@ -0,0 +1,111 @@
...
+                int percentage = parse_decimal_as_percentage(argv[i + 1]);
+                i++;
+                if (percentage < 50 || percentage > 500) {
+                    SDL_Log(DISPLAY_SCALE_ERROR_MESSAGE);
+                    ok = 0;
+                } else {
+                    output_args->display_scale_percentage = percentage;
+                }
+            } else {
+                    SDL_Log(DISPLAY_SCALE_ERROR_MESSAGE);
```
crudelios: Identation :P

src/platform/arguments.c
dcnieho: Ok!

dcnieho:
I don't know if you want to be warning-clean, but i get:
C:\src\platform\arguments.c(29): warning C4244: 'initializing': conversion from '__int64' to 'int', possible loss of data

dcnieho:
Runs real well now, i can't get it to crash either

dcnieho:
And yes, this is really nice code, and super well organized for as far as i have seen :)

src/platform/arguments.c
bvschaik: Oops 🤡

bvschaik:
> I don't know if you want to be warning-clean, but i get:
> C:\src\platform\arguments.c(29): warning C4244: 'initializing': conversion from '__int64' to 'int', possible loss of data

Yes, I'd like to be warning-clean. Clang on MacOS didn't complain about this, thanks!

crudelios:
I'll be cleaning all warnings I can find once both this and #120 are merged. Apart from two warnings which are related to the Switch SDK header files, I think all of them can be removed.

devnoname120:
@bvschaik 
image

bvschaik:
@devnoname120 Oops, I see that the switch build was expecting the cursor to be 32x32, that's why it looks mangled. I have fixed it, I hope. Could you test again?

rsn8887:
The hotspot is wrong on Switch and Vita it is always 0 0 even with shovel. It was always wrong.

devnoname120:
@rsn8887 This is true, but should be in another pull request I think.

bvschaik:
Agreed. I almost have it implemented, I'll create a branch for it once this one is merged.

rsn8887:
I thought sdl doesn’t allow negative coordinates

bvschaik:
> I thought sdl doesn’t allow negative coordinates (sdl_rect uses uint not int).

[SDL_Rect](https://wiki.libsdl.org/SDL_Rect) uses `int`, so its possible to specify negative x/y coords. See also this branch: https://github.com/bvschaik/julius/tree/feature/cursor-hotspot. For Vita it seems to work fine.

rsn8887:
The hidpi branch doesn’t work on Switch: no mouse cursor visible, but title screen and menu display correctly.

rsn8887:
With this PR, on Switch, the mouse pointer is invisible everything else works correctly.

rsn8887:
I fixed the missing mouse pointers on Switch in this PR:
https://github.com/bvschaik/julius/pull/185

PR #188, opened 2019-12-02
Fix all compiler warnings

crudelios:
Except for two on the Switch build which are related to DevKitPro header files.

Tested: Windows MSVC 2019, all AppImage and Travis builds. Please test for warnings in specific compilers.

### DIFF
--- a/src/core/dir.c
+++ b/src/core/dir.c
@@ -46,8 +46,6 @@ static const char *filename_to_utf8(const wchar_t *str)
 #define CURRENT_DIR "."
 #endif
 
-#define MAX_FILE_NAME_BUFFER (FILE_NAME_MAX * 2)
-
 static dir_listing listing;
 static int listing_initialized = 0;
 
@@ -55,7 +53,7 @@ static void clear_dir_listing(void)
 {
     if (!listing_initialized) {
         for (int i = 0; i < DIR_MAX_FILES; i++) {
-            listing.files[i] = malloc(MAX_FILE_NAME_BUFFER * sizeof(char));
+            listing.files[i] = malloc(FILE_NAME_MAX * sizeof(char));
         }
         listing_initialized = 1;
     }
@@ -82,7 +80,7 @@ const dir_listing *dir_find_files_with_extension(const char *extension)
     while ((entry = fs_dir_read(d)) && listing.num_files < DIR_MAX_FILES) {
         const char *name = dir_entry_name(entry);
         if (file_has_extension(name, extension)) {
-            strncpy(listing.files[listing.num_files], name, MAX_FILE_NAME_BUFFER);
+            strncpy(listing.files[listing.num_files], name, FILE_NAME_MAX);
             listing.files[listing.num_files][FILE_NAME_MAX - 1] = 0;
             ++listing.num_files;
         }
--- a/src/core/dir.c
+++ b/src/core/dir.c
@@ -46,6 +46,8 @@ static const char *filename_to_utf8(const wchar_t *str)
 #define CURRENT_DIR "."
 #endif
 
+#define MAX_FILE_NAME_BUFFER (FILE_NAME_MAX * 2)
+
 static dir_listing listing;
 static int listing_initialized = 0;
 
@@ -53,7 +55,7 @@ static void clear_dir_listing(void)
 {
     if (!listing_initialized) {
         for (int i = 0; i < DIR_MAX_FILES; i++) {
-            listing.files[i] = malloc(FILE_NAME_MAX * sizeof(char));
+            listing.files[i] = malloc(MAX_FILE_NAME_BUFFER * sizeof(char));
         }
         listing_initialized = 1;
     }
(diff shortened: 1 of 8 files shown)


src/core/dir.c
```diff
@@ -46,14 +46,16 @@ static const char *filename_to_utf8(const wchar_t *str)
...
+
 static dir_listing listing;
 static int listing_initialized = 0;
 
 static void clear_dir_listing(void)
 {
     if (!listing_initialized) {
         for (int i = 0; i < DIR_MAX_FILES; i++) {
-            listing.files[i] = malloc(FILE_NAME_MAX * sizeof(char));
+            listing.files[i] = malloc(MAX_FILE_NAME_BUFFER * sizeof(char));
```
bvschaik: What was the warning here? Allocating twice the amount of memory seems overkill.

test/sav/run.c
```diff
@@ -53,7 +53,6 @@ static int run_autopilot(const char *input_saved_game, const char *output_saved_
     printf("Running autopilot: %s --> %s in %d ticks\n", input_saved_game, output_saved_game, ticks_to_run);
     signal(SIGSEGV, handler);
 
-    chdir("data/sav");
```
bvschaik: Changing dir is no longer necessary?

test/sav/sav_compare.c
```diff
@@ -288,12 +289,18 @@ static int unpack(const char *filename, unsigned char *buffer)
     }
     int offset = 0;
     for (int i = 0; save_game_parts[i].length_in_bytes; i++) {
+        int result = 0;
         if (save_game_parts[i].compressed) {
-            read_compressed_chunk(fp, &buffer[offset], save_game_parts[i].length_in_bytes);
+            result = read_compressed_chunk(fp, &buffer[offset], save_game_parts[i].length_in_bytes);
         } else {
-            fread(&buffer[offset], 1, save_game_parts[i].length_in_bytes, fp);
+            result = (fread(&buffer[offset], 1, save_game_parts[i].length_in_bytes, fp) == save_game_parts[i].length_in_bytes);
```
bvschaik: Unnecessary parentheses.

src/game/file_io.c
```diff
@@ -597,10 +602,14 @@ static void savegame_read_from_file(FILE *fp)
 {
     for (int i = 0; i < savegame_data.num_pieces; i++) {
         file_piece *piece = &savegame_data.pieces[i];
+        int result = 0;
         if (piece->compressed) {
-            read_compressed_chunk(fp, piece->buf.data, piece->buf.size);
+            result = read_compressed_chunk(fp, piece->buf.data, piece->buf.size);
         } else {
-            fread(piece->buf.data, 1, piece->buf.size, fp);
+            result = (fread(piece->buf.data, 1, piece->buf.size, fp) == piece->buf.size);
```
bvschaik: Unnecessary parentheses.

test/sav/run.c
crudelios: No, when I added a check to fix the warning all the tests reported as failed as that dir does no longer exist. I'm guessing that's leftover code from a time when tests were structured differently.

src/core/dir.c
crudelios: This one was problematic.

The Switch and Vita builds were reporting this warning:

```
/build/git/src/core/dir.c:83:13: warning: 'strncpy' output may be truncated copying 199 bytes from a string of length 255 [-Wstringop-truncation]
             strncpy(listing.files[listing.num_files], name, FILE_NAME_MAX - 1);
```

The problem is that `dirent.h` does not define a standard size for its member `d_name` in the `dirent` structure. In some platforms its size is set to `200`, and in others to `255` or even `260`.

Since `FILE_NAME_MAX` is always `200`, hence the problem with that warning. Also, there is no platform-specific macro for the `d_name` string length. Therefore I decided to create a new macro, making it twice as large as `FILE_NAME_MAX` for convenience.

However it could probably be set to `300` and still work fine.

bvschaik: Would increasing `FILE_NAME_MAX` to 300 help? I was looking at the UTF-8 encoding/decoding functions today, and having a filename full of "weird" characters would go over the limit.

crudelios: > Would increasing `FILE_NAME_MAX` to 300 help? 

Yes it would!

> I was looking at the UTF-8 encoding/decoding functions today, and having a filename full of "weird" characters would go over the limit.

Yeah, UTF-8 can take up to 4 bytes per char, especially on special glyphs like chinese characters... I wonder what the char limit for chinese windows is!

bvschaik: > > Would increasing FILE_NAME_MAX to 300 help?

> Yes it would!

Then please do :smile:

> I wonder what the char limit for chinese windows is!

Still 255, since Windows uses 16-bit unicode for filenames. We only get them as UTF-8 because of the compatibility `dirent.h` file, which is there because Windows does not support opendir natively.

crudelios: Done!

I didn't think `FILE_NAME_MAX` was set by Julius, but then again I didn't really check 😆

PR #190, opened 2019-12-04
Add ability to use shift with hotkeys

scresante:
Small addition to allow future hotkeys to be modified with shift key

### DIFF
--- a/src/input/hotkey.c
+++ b/src/input/hotkey.c
@@ -159,7 +159,7 @@ static void input_number(int number)
     }
 }
 
-void hotkey_character(int c, int with_ctrl, int with_alt)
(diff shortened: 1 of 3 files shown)


bvschaik:
Could you explain why I should merge this while the `is_shift_down` is never used anywhere? (see also: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it))

scresante:
I agree with it you, it's not necessary to merge it.

I'm pretty new to contributing to projects on github (and C). I'm working, slowly, on an idea I had to build a keyboard menu access system, which would be a major UX improvement. 

Has any decision been made for a way to track new enhancements that aren't simply bugs? Until then, I'll develop on my own fork, and wait for making pull requests until I've made real progress.

PR #195, opened 2019-12-11
Android support

crudelios:
Julius on Android

This PR adds support for Android.

**What works:**
* Travis automated builds (with bintray deployment as soon as the code is merged)
* Storage access permissions
* Requesting C3 directory
* Touch support
* The game!!!

**What needs to be done**
* [x] Virtual keyboard support
* [x] Code refactoring
* [x] General improvements
* [x] Removing access to `Display options` (should always display on fullscreen using detected resolution and scaling)

**Building the apk**

First, copy the `SDL` and `SDL_mixer` sources to `julius/ext/SDL2`.

Then either:

* Download Android Studio and install the SDK as well as its NDK and CMake dependencies.

* Select `Open existing project` and select folder `android.

Or:

* Install Android Studio for command line and run, from the julius directory, `cd android && ./gradlew assembleDebug`

    A docker image is available at https://hub.docker.com/r/npetrovsky/docker-android-sdk-ndk, preventing the need to download Android SDK. Just beware as the image itself is almost 2GB in size when compressed.

In either case, the `apk` will be built to the short and easy to remember path `julius/android/distribution/android/julius/outputs/apk/debug`

Since the code is more or less ready, I'd like some reviews or suggestions if possible.


crudelios:
Builds seem to be failing due to a problem with the cache. (I messed with the cache too much) @bvschaik can you delete the travis cache for this PR, please?

bvschaik:
> @bvschaik can you delete the travis cache for this PR, please?

Done. The cache was 2GB 😮

crudelios:
There are some problems related to video playback, but I'll fix those once #208 is merged. Also, I'm thinking about the best way to cache the SDL builds for android in order to make Travis about a minute faster.

crudelios:
Due to a major issue I just found regarding save files, I will close this PR for now and rework the implementation.

For reference, it's impossible (at least as of Android 10) to save to an external SD card using `fopen`, the only way being using Android's SAF implementation. So I need to create a wrapper around each file write operation in order to use SAF on the Java side.

crudelios:
As I've implemented the necessary changes, this PR can be reopened.

CMakeLists.txt
```diff
@@ -5,9 +5,14 @@ if(VITA_BUILD AND SWITCH_BUILD)
...
 endif()
 
+if((VITA_BUILD OR SWITCH_BUILD) AND ANDROID_BUILD)
+    message(FATAL_ERROR "Cannot build with -DVITA_BUILD=ON or -DSWITCH_BUILD=ON and -DANDROID_BUILD=ON at the same time")
+endif()
+
 option(DRAW_FPS "Draw FPS on the top left corner of the window." OFF)
 cmake_dependent_option(VITA_BUILD "Build for the PlayStation Vita handheld game console." OFF "NOT MSVC" OFF)
 cmake_dependent_option(SWITCH_BUILD "Build for the Nintendo Switch handheld game console." OFF "NOT MSVC; NOT VITA_BUILD" OFF)
+cmake_dependent_option(ANDROID_BUILD "Build for the Android environment." OFF "NOT VITA_BUILD; NOT SWITCH_BUILD" OFF)
```
devnoname120: Is `NOT MSVC` missing here?

crudelios: Yes it is. Good catch.

crudelios:
I can't think of anything else to change.

I'll stop adding commits, awaiting review.

m-mohr:
Great! I'd love to get this running on my phone so that I can play it on the next train ride or so. I'm happy to test, but can't really help with reviewing, I guess.

crudelios:
@m-mohr thanks for your interest in testing the Android version.

Here's a link to the Android apk: https://drive.google.com/open?id=1RaIQh5Y7H8bviobp7f84cieve0R7RAZ6

You'll need to allow installation from untrusted sources. Your system should warn you about it when you try to install and then provide a button to allow the installation.

Android version 5.0 or later is required. You will obviously also need to copy all C3 files to your device.

What needs testing: if it opens, initial folder selection, display scaling (it should scale but still at least allow access to the entire UI), time it takes to show the first screen and loading times in general, saving and deleting cities, any random crashes and slowdowns.

Touch support is less than perfect and is not currently working properly for the map editor, but any feedback about that would also be appreciated. For instructions on using Julius with touch, please check #107.

Also note: although both should work, this should be more playable in a tablet than in a cell phone.

411752230:
I‘d like to help too. Folder selection works fine. And so does the UI.

It seems that the game will crash when clicking the button right to the "Overlays" ,or repeatedly clicking a building button like 'building housing'.

crudelios:
> It seems that the game will crash when clicking the button right to the "Overlays" ,or repeatedly clicking a building button like 'building housing'.

Yet another crash that I can't seem to replicate... Does it exit the game or does it just hang?

411752230:
> > It seems that the game will crash when clicking the button right to the "Overlays" ,or repeatedly clicking a building button like 'building housing'.
> 
> Yet another crash that I can't seem to replicate... Does it exit the game or does it just hang?

I got a screen recording for ya.
https://drive.google.com/open?id=1TEIg70NWF6WtI2x1y2F5-hTOk9gAHRHt

Here are some info about my android phone.
S00122-19020510

crudelios:
@411752230 Thank you for the video. That's definitely not happening on any device I test. I have a suspicion of what might be causing it. I'll message you in gitter later with a custom build to see if the crash persists. But I feel this will be done by trial and error ;\

m-mohr:
@crudelios Thanks, I got it running. I'm using a Xiaomi Mi Mix S2 with MI UI, which is a modded Android 9. 

Some remarks:
1. Choosing the Caesar folder was hard as the folder chooser opened just listed "Downloads" and "Recent" (see screenshot).
2. Indeed the touch support is somewhat counter-intuitive.
  A. It would be great to be able to scroll with to fingers for example instead of touching near the edge.
  B .How can I "reset" the tool used (e.g. planting, house, ...)? Once I clicked something in the menu it's always building houses and can't get that away and scrolling builds a lot of houses then. 
  C. Is there an equivalent for a right click? It's hard to get out of menus or delete messages without right clicking.
3. I don't have the issue @411752230 has.

crudelios:
> Choosing the Caesar folder was hard as the folder chooser opened just listed "Downloads" and "Recent"

If you click on the vertical `...` to the left, an option entitled `Show internal folder` or something should appear. If you click it, you'll have access to the internal storage of your phone where it shows `Downloads`.

> It would be great to be able to scroll with two fingers for example instead of touching near the edge.

If you don't have any construction selected you can scroll with just one finger. Just click and drag on the city.

> How can I "reset" the tool used (e.g. planting, house, ...)? Once I clicked something in the menu it's always building houses and can't get that away and scrolling builds a lot of houses then.
Related: Is there an equivalent for a right click? It's hard to get out of menus or delete messages without right clicking.

Both are achieved by clicking and holding with a finger on the screen, and then tapping with a second finger.

For checking the details of a house or walker, the first finger (the one that is held in place) selects the tile and the second finger can be tapped anywhere.

m-mohr:
Thanks, @crudelios !

> If you click on the vertical `...` to the right, an option entitled `Show internal folder` or something should appear. If you click it, you'll have access to the internal storage of your phone where it shows `Downloads`.

I tried that, but it didn't show any folders. The screen was basically blank.

> Note that this dialog is Android-native, not from Julius, and its use is required (otherwise julius doesn't have permissions to access the files).

I'm aware of that. When that dialog opens in other apps, there are more folders and tings shown, but not here. Not sure why? Maybe some options that need to be set in addition?

> If you don't have any construction selected you can scroll with just one finger. Just click and drag on the city.

> Both are achieved by clicking and holding with a finger on the screen, and then tapping with a second finger.

Is there a better gesture for that? Maybe touching twice? That's a bit annoying, I think. It works for deleting messages, but for "tool reset" I need to do that twice (click one finger and then tap with another finger twice).

> For checking the details of a house or walker, make sure no construction is active, then with the first finger (the one that is held in place) selects the tile and the second finger can be tapped anywhere.

Why is it not just a single or double touch?

Sometimes when placing buildings, I'm regularly coming close to the edge and the screen starts scrolling. Maybe it would be good to disable scrolling in this case?

Also, the screen sometimes says "Quit with right click" (or so, roughly translated from German). Shouldn't that be removed?

crudelios:
@m-mohr The whole problem is that internally Julius is based on a mouse and not really adapted for touch. I had to figure out the... best?... way to emulate the mouse behavior using touch.

So my idea was: single click with one finger = left click; hold with one finger, tap with the other = right click. So whenever you see `right click`, you can use that touch gesture.

I then added the ability to scroll the city and empire map clicking and dragging with a finger (because it simply made sense, and I even though about implementing it by clicking and dragging the mouse as well).

Cancelling a build only requires two "right-clicks" for houses, roads, aqueducts and walls. The first is to prevent the "build-on-drag" (so you can relocate your starting point), the second to exit that construction. This feature is not even available using the mouse.

> Why is it not just a single or double touch?

Double touch is not a bad idea! Thanks. (single touch is reserved for scrolling the city)

> Sometimes when placing buildings, I'm regularly coming close to the edge and the screen starts scrolling. Maybe it would be good to disable scrolling in this case?

Perhaps not disable it entirely, but reduce the area where it starts scrolling. Right now I find that annoying too.

> Not sure why? Maybe some options that need to be set in addition?

Possibly. I'll check.

m-mohr:
Hmm, so let's think about best gestures. I'm also thinking about how other apps do that.

So what do we have on a computer with a mouse:
* movement (scrolling): Taken from some maps: Use two fingers and move/scroll. Alternative: One tap, keep pressing and move.
* left mouse click: Single tap
* left mouse pressed for longer time (e.g. building houses): Double tap, keep pressing after the second tap and move.
* right mouse click (exit menus, delete messages, reset tool, ...): Double tap? Or some gesture? In menus: Use hardware back button from Android?
* middle mouse scroll (not sure, is that used?)
* more?

These gestures are what I would expect intuitievely.
What do you think @crudelios ?

crudelios:
New version: https://drive.google.com/open?id=1RaIQh5Y7H8bviobp7f84cieve0R7RAZ6

@411752230  This version should no longer crash on your phone with Android 7
@m-mohr This version should also improve the folder select dialog

m-mohr:
@crudelios Yes, thanks, the folder select dialog works like a charm now!

What do you think about https://github.com/bvschaik/julius/pull/195#issuecomment-577422448?

crudelios:
After trying Julius on Samsung DeX, I think I should implement proper support for it. For those who have Samsung DeX, they could get both a mobile and a desktop experience at the same time.

411752230:
> New version: https://drive.google.com/open?id=1RaIQh5Y7H8bviobp7f84cieve0R7RAZ6
> @411752230 This version should no longer crash on your phone with Android 7
> @m-mohr This version should also improve the folder select dialog

Nice fix! I just tested it with my phone Android 7.0. The game stops crashing. But doing the same like the video I uploaded yesterday, the touch support of the game will get stuck.

crudelios:
> But doing the same like the video I uploaded yesterday, the touch support of the game will get stuck.

Yeah I had it get stuck sometimes, but it's random and doesn't happen very often (usually requires some crazy inputs).

Also, how are your load times? On my Galaxy Tab A it takes upwards of 20 seconds to first load the game :/

m-mohr:
The scrolling sometimes is a bit ugly: https://drive.google.com/file/d/16LnktPTWSDB4AILEh2v4bernXP1601nI/view?usp=drivesdk

crudelios:
The scrolling is like that because the camera position is tile-based. It's how the game is coded.

Implenenting pixel based (smooth) scrolling is something I want to do, but it's a large project that I won't start it before this one is merged.

m-mohr:
That would be great, I guess also for Windows etc

411752230:
> > But doing the same like the video I uploaded yesterday, the touch support of the game will get stuck.
> 
> Yeah I had it get stuck sometimes, but it's random and doesn't happen very often (usually requires some crazy inputs).
> 
> Also, how are your load times? On my Galaxy Tab A it takes upwards of 20 seconds to first load the game :/

I see. touching support getting stuck is easy for me as I described. Just start a game and play until its end, you will get stuck sooner or later. It shows a button always being tapped and pressed while others cannot be tapped again.

My loading time is quick which wont spend many seconds.

The scrolling system is a bit of strange i think especially for devices with lower resolutions. You will find the the camera scrolling too sensitive when dragging your finger near the edge of the screen to build roads.

crudelios:
> I see. touching support getting stuck is easy for me as I described.

I had it happen only once, but I'll try to pinpoint the cause.

Slow loading times are, once again, related to the audio sample rate. The game takes about 5 seconds to load if the sample rate is `22050`, but more than 20 if the sample rate is `44100`. This seems to only happen on android with armv7 processors.

The scrolling will be changed, but independently of the android build, since touch is not 100% android-related.

Right now I just want this PR to be merged. AFAIK there are no pending issues exclusively related to Android.

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
+package bvschaik.julius;
+
+import android.content.Intent;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.widget.Toast;
+import org.libsdl.app.SDLActivity;
+
+public class JuliusMainActivity extends SDLActivity
```
bvschaik: Please keep to the Java code style conventions of putting the curly brace on the same line

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+{
+    private static final int GET_FOLDER_RESULT = 500;
+    private static final int rwFlagsPermission = Intent.FLAG_GRANT_READ_URI_PERMISSION
```
bvschaik: Constants should be ALL_CAPS

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+                                               | Intent.FLAG_GRANT_WRITE_URI_PERMISSION;
+    public static boolean paused = true;
```
bvschaik: Does this need to be public and static?

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Override
+    public void onCreate(Bundle savedInstanceState)
```
bvschaik: Useless override, method can be removed.

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+    {
+        super.onCreate(savedInstanceState);
+    }
+
+    @Override
+    public void onStart()
+    {
+        super.onStart();
+        paused = false;
```
bvschaik: Either use onStart/onStop for setting paused, or use onResume/onPause (I think this one is preferred here). Don't mix both.

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+    protected void onResume()
+    {
+        super.onResume();
+        paused = false;
+    }
+
+    public void showDirectorySelection()
+    {
+        // Wait before showing window
+        Sleep(1000);
```
bvschaik: Sleeping on the main thread? Bad idea. Android might kill the app because of an [ANR](https://developer.android.com/topic/performance/vitals/anr) (app not responding).

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+        handler.post(new Runnable()
+        {
+            public void run()
+            {
+                Toast.makeText(getApplicationContext(), message, Toast.LENGTH_SHORT).show();
+            }
+        });
+    }
+
+    static public void waitOnPause()
```
bvschaik: Modifier order is `public static`

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
```diff
@@ -0,0 +1,120 @@
...
+            Sleep(1000);
+        }
+    }
+
+    public float getScreenScale()
+    {
+        return getResources().getDisplayMetrics().density;
+    }
+
+    static private void Sleep(int ms)
```
bvschaik: Method naming: snakeCase. Modifier order is `private static`.

src/window/input_box.c
```diff
@@ -0,0 +1,8 @@
+#include "input_box.h"
```
bvschaik: This is not a window, it's a widget.

src/window/input_box.h
```diff
@@ -0,0 +1,16 @@
+#ifndef WINDOW_INPUT_BOX_H
```
bvschaik: This is not a window, it's a widget.

android/julius/src/main/java/bvschaik/julius/JuliusMainActivity.java
crudelios: I just wanted to show a toast message before the "select folder" appears, as that makes the toast message disappear before the user has the time to read. Do you know of an alternative? I really don't want to show a message box...

crudelios:
@bvschaik I removed all the platform-related defines from `core/dir.c` and `core/file.c`.

I'm not entirely happy with the end result, but it was the best I could come up with.

m-mohr:
@crudelios Did you intend to make the game stop when you press the screen, e.g. for scrolling?

crudelios:
@m-mohr Yes. Touching initiantes a scroll action and scrolling stops the screen.

m-mohr:
What is the reason for it? It felt weird yesterday. It doesn't stop on Windows when scrolling, right?

crudelios:
It does, the game always stops while scrolling.

crudelios:
@411752230:
> I see. touching support getting stuck is easy for me as I described. Just start a game and play until its end, you will get stuck sooner or later. It shows a button always being tapped and pressed while others cannot be tapped again.

This should be fixed in the latest build. Please test:

https://drive.google.com/open?id=1RaIQh5Y7H8bviobp7f84cieve0R7RAZ6

> The scrolling system is a bit of strange i think especially for devices with lower resolutions. You will find the the camera scrolling too sensitive when dragging your finger near the edge of the screen to build roads.

This is being fixed on #287.

m-mohr:
Gameplay with smooth scrolling is much better!
Overall it feels good to play, but the 'right click' is still very annoying. Placing buildings is also a bit complicated. Other than that no issues in the last hour.

411752230:
> @411752230:
> 
> > I see. touching support getting stuck is easy for me as I described. Just start a game and play until its end, you will get stuck sooner or later. It shows a button always being tapped and pressed while others cannot be tapped again.
> 
> This should be fixed in the latest build. Please test (note: due to a one-time setting change, please uninstall the old Julius for android before you install this, or you'll end up with two Julius in your device):
> 
> https://drive.google.com/open?id=1RaIQh5Y7H8bviobp7f84cieve0R7RAZ6

Like a charm. The inertia of the new smooth scrolling is unbelievable! With this latest one, I have not seen any stuck until now.

crudelios:
> New Discovery: Tick the play intro video and restart the game, the sierra logo is skipped with Impression logo played directly.

Confirmed. If you fully close the app (i.e. list all open Android apps and select "Close all" or only close that one), then when you reopen the game it will show the first video again. Definitely something for me to check.

crudelios:
> New Discovery: Tick the play intro video and restart the game, the sierra logo is skipped with Impression logo played directly.

Fixed by #325. I will post a new Android build here once the fix is merged.

crudelios:
As promised, here's the latest build with the intro video fix:

https://drive.google.com/open?id=1RaIQh5Y7H8bviobp7f84cieve0R7RAZ6

crudelios:
Per @bvschaik's suggestion, I'm closing this and dividing it into smaller, more manageable PR's that will be merged independently before merging this one.

PR #199, opened 2019-12-20
Improve CMake to allow SDL and SDL_mixer to be detected from folder ext/SDL2

crudelios:
Allows travis to build julius without requiring make install, makng builds faster.

Also allows not having to set CMAKE_PREFIX_PATH in MSVC/MinGW, as long as the SDL libraries are saved to ext/SDL2.

In order for travis to work, its cache must first be deleted.

### DIFF
--- a/.ci_scripts/install_sdl.sh
+++ b/.ci_scripts/install_sdl.sh
@@ -37,22 +37,25 @@ function get_sdl_lib_url {
 
 function install_sdl_lib {
   local LIB=$1
-  if [ -d $LIB/build ]
+  if [ ! "$(ls -A $LIB)" ]
   then
-    cp -r $LIB $LIB-final
-    cd $LIB-final/build
-  else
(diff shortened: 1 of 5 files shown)


crudelios:
This obviously only works after the cache is rebuilt.

On my fork, build times on Linux gcc and Linux clang (comfigurations #1 and #3) went from ~2m30s to ~1m10s. The others are unnaffected by this change.

crudelios:
I realized I still need to edit the cmake files to make Windows MinGW not require `CMAKE_PREFIX_PATH`, so please don't merge until I change the code.

crudelios:
Done. Should be ready for merge.

This will probably make #195 fail to merge, but I'll rebase that PR with the latest master.

bvschaik:
Do we need to update the build instructions for this?

crudelios:
Yes, for Windows, though it still works as-is.

PR #208, opened 2020-01-05
Video improvements, incl libsmacker replacement

bvschaik:


crudelios:
I'll be checking/testing this one out tomorrow!

bvschaik:
@devnoname120 @rsn8887 could you perhaps test the videos on Vita / Switch as well? The new smacker implementation is 35% faster than the old libsmacker, and the game now only loads videos once instead of twice. Perhaps we don't need this vita performance improvement anymore?

https://github.com/bvschaik/julius/blob/3f3b84922e074dfb9c481fba0692fbd723d9644c/src/graphics/video.c#L49-L55

crudelios:
I have tested about 15 different videos (including the one in German which gave audio problems before) and so far, everything works perfectly.

I'll report when I tested them all.

crudelios:
I have finished testing all the videos, and while the videos played fine, there is an issue with the audio.

If you pay close attention, you'll notice that most, if not all, videos have many constant click noises, usually many per second. It's hard to catch, but it's definitely there. This is more noticeable in some videos than others, but the first one where I noticed it was in `Lose_game.smk`.

bvschaik:
> If you pay close attention, you'll notice that most, if not all, videos have many constant click noises, usually many per second. It's hard to catch, but it's definitely there. This is more noticeable in some videos than others, but the first one where I noticed it was in `Lose_game.smk`.

Hmm. I can't hear that on my Mac. There are two changes that may be responsible: I changed the `AUDIO_RATE` in sound_device.c to 44100 (was: 22050), and the videos are now not loaded twice, but that may in some cases result in an audio buffer underrun.
Could you perhaps try putting the audio rate back to 22050 and see if that fixes it?

crudelios:
Setting `AUDIO_RATE` back to `22050` seems to fix it.

However, I think there are issues with the audio when streaming the game using `22050`. I'm not sure it's a problem with julius or SDL however.

bvschaik:
> However, I think there are issues with the audio when streaming the game using `22050`.

What do you mean? It's been on 22050 for ages, I only upped it to 44100 in this branch because some videos use 44.1kHz.

crudelios:
I tested on `master` (which uses `libsmacker`) and the clicks can also be heard. So it's not a problem with your implementation.

> What do you mean?

When GamerZakh posted his video about Julius, he complained that he had problems with the audio when listening to it through his created videos of the game. I always figured it had something to do with the `AUDIO_RATE`, and have wanted to set it to `44100` since.

bvschaik:
You tested on master with the 22050 or 44100 setting?
Soooo, what to do now?

crudelios:
Tested on `master` with `44100`.

I can try building with the latest unstable versions of SDL and SDL_mixer to see if this was fixed in the meantime.

bvschaik:
Do the rest of the sounds/music work OK with 44100 or does that also produce clicks?

crudelios:
Only the videos trigger the clicks, the rest of the sounds work fine.

crudelios:
How about keeping audio rate at `22050` by default but adding an `--audio-rate` command line option?

bvschaik:
I propose the following timeline:
1. Set audio rate to 22050
2. Merge this branch
3. Release 1.2.0
4. After release: research performance impact of setting audio rate to 44100, investigate the click in videos, and then decide on a command-line argument or perhaps compile-time switch.

Thoughts?

crudelios:
> Thoughts?

Agreed. Go ahead with the merge, as the videos are working.

Also, I already had a look at the cause of the clicks, and the only possible culprit I can find is `SDL_ConvertAudio`.

bvschaik:
> Also, I already had a look at the cause of the clicks, and the only possible culprit I can find is `SDL_ConvertAudio`.

Or `SDL_MixAudioFormat` to change the video's volume? That one is easier to disable for testing.
I'm going to modify libsmacker's `driver` program so it writes a .wav instead of raw audio. Perhaps we can see something by analyzing the resulting .wav.

crudelios:
> Perhaps we can see something by analyzing the resulting .wav.

I already tried capturing the recorded sound using Audacity and, while the clicks can be heard, I couldn't find anything strange with the waveform itself. Perhaps you'll have better luck than I did.

src/core/smack.c
```diff
@@ -0,0 +1,953 @@
...
+    }
+    uint8_t value = bs->data[bs->index] >> bs->bit_index;
+    bs->index++;
+    value |= (bs->data[bs->index] & BIT_MASKS[bs->bit_index]) << (8 - bs->bit_index);
+    return value;
+}
+
+// 8-bit huffman tree functions
+
+static huffnode8 *build_tree8_nodes(bitstream *bs, hufftree8 *tree, int value, int bits)
```
crudelios: Unless I'm missing something, `int value` is passed on recursively, but never actually used.

src/core/smack.c
```diff
@@ -0,0 +1,953 @@
...
+            free(tree->escape_nodes[i]);
+        }
+    }
+    free_node16(tree->root);
+    free_tree8(tree->low);
+    free_tree8(tree->high);
+    free(tree);
+}
+
+static huffnode16 *build_tree16_nodes(bitstream *bs, hufftree16 *tree, int value, int bits)
```
crudelios: Same here. `int value` and `int trees` are passed on recursively but never actually used.

src/core/smack.c
```diff
@@ -0,0 +1,953 @@
...
+    } else {
+        node->is_leaf = 1;
+        uint8_t lo_val = lookup_tree8(bs, tree->low);
+        uint8_t hi_val = lookup_tree8(bs, tree->high);
+        uint16_t leaf_value = lo_val | (hi_val << 8);
+        node->value = leaf_value;
+
+        for (int i = 0; i < 3; i++) {
+            if (leaf_value == tree->escape_codes[i]) {
+                tree->escape_nodes[i] = node;
```
crudelios: This may not make a difference for our videos, but shouldn't `node->value` be set to `0` here? Or do you do it somewhere else?

As per the [documentation](https://wiki.multimedia.cx/index.php/Smacker#Optimized_Compression): `5a: When you read Leaf and it's value matches one of the markers obtained on step 4, remember the leaf for this marker and set leaf's value to zero.`

src/core/smack.c
```diff
@@ -0,0 +1,953 @@
...
+                log_error("SMK: invalid palette data", 0, 0);
+                return 0;
+            }
+            memcpy(&new_palette[color_index], &s->frame_data.palette[offset], sizeof(int32_t) * num_entries);
+            color_index += num_entries;
+            index += 2;
+        } else {
+            // Literal color
+            new_palette[color_index] =
+                    (PALETTE_MAP[data[index] & 0x3f] << 16) |
```
crudelios: Although more verbose, the `& 0x3F` is not strictly needed. If we ever get to this execution branch, the byte is already `00xxxxxx`.

crudelios:
Sorry for the rather late analysis, only now did I have the time to fully compare the code to the documentation.

src/core/smack.c
bvschaik: Exactly :)

src/core/smack.c
bvschaik: It's actually a leftover from when I had verbose logging on practically everything. Good catch.

src/core/smack.c
bvschaik: True, but you have to do it for the next two bytes anyway, so I'd rather keep the operation the same for all three bytes.

crudelios: Fair enough. Looks better anyway.

devnoname120:
> 
> 
> @devnoname120 @rsn8887 could you perhaps test the videos on Vita / Switch as well? The new smacker implementation is 35% faster than the old libsmacker, and the game now only loads videos once instead of twice. Perhaps we don't need this vita performance improvement anymore?
> 
> https://github.com/bvschaik/julius/blob/3f3b84922e074dfb9c481fba0692fbd723d9644c/src/graphics/video.c#L49-L55

@bvschaik I've tested on both my PS Vita and @rinnegatamante's using `SMACKER_MODE_DISK` and I can confirm that with your libsmacker replacement videos are smooth. I'll open a pull request to remove this Vita-specific wart.

PR #214, opened 2020-01-08
Audio fixes, and set rate to 44100

crudelios:
* Fix audio clicks during video playback at 44100 audio frequency for SDL >= 2.0.7.
* Fix volume of videos.
* Set global audio frequency rate to 44100.

### DIFF
--- a/src/platform/sound_device.c
+++ b/src/platform/sound_device.c
@@ -9,29 +9,37 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define AUDIO_RATE 22050
+#define AUDIO_RATE 44100
 #define AUDIO_FORMAT AUDIO_S16
 #define AUDIO_CHANNELS 2
 #define AUDIO_BUFFERS 1024
 
 #define MAX_CHANNELS 150
 
+#if SDL_VERSION_ATLEAST(2, 0, 7) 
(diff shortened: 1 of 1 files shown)


bvschaik:
I'll have a look at the SDL_mixer source this weekend to see how they do stuff without AudioStreams.

crudelios:
They use AudioStreams :P That's how I learned about them.

devnoname120:
I've just tested on the PS Vita with the https://github.com/bvschaik/julius/pull/218 changes and it seems to work fine as well. I've noticed there can be a bit of stuttering at the end of the video but it doesn't happen every time. (And it might just be bad luck, not certain it's caused by this change).

crudelios:
@devnoname120 are you referring to this PR regarding the stutters? Do you mean it doesn't happen just with #208?

Also, with #208 do you hear any clicks in the audio? It's more noticeable in some videos than others...

bvschaik:
I'm wondering about this as well. PR #208 can cause audio stutters if the decoder cannot keep up with the video playback and lags more than one second behind. The stuttering should happen more often on longer videos, such as `lose_game.smk`, could you confirm that?
Maybe I should implement frame skipping if the decoding process cannot keep up.

PR #231, opened 2020-01-15
Intro video improvements

crudelios:
* Play the logo, the intro video and the credits (in that order) when clicked on the main menu. Click each one to move to the next
* Play the intro music (not ROME1) after intro videos are played

### DIFF
--- a/src/window/intro_video.c
+++ b/src/window/intro_video.c
@@ -4,20 +4,35 @@
 #include "graphics/screen.h"
 #include "graphics/video.h"
 #include "graphics/window.h"
+#include "sound/music.h"
+
+#define NUM_INTRO_VIDEOS 2
 
 static struct {
     int width;
     int height;
+    int current_video;
 } data;
+
+static const char *intro_videos[NUM_INTRO_VIDEOS] = { "smk/logo.smk", "smk/intro.smk" };
+
+static int start_next_video(void)
+{
+    while(data.current_video < NUM_INTRO_VIDEOS) {
+        if (video_start(intro_videos[data.current_video])) {
+            video_size(&data.width, &data.height);
+            video_init();
+            ++data.current_video;
+            return 1;
+        }
+    }
+    return 0;
+}
 
 static int init(void)
 {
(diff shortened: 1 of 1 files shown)


bvschaik:
Hah, you're just ahead of me, I had a similar implementation in mind for tonight :wink:

src/window/intro_video.c
```diff
@@ -4,22 +4,37 @@
...
+#define NUM_INTRO_VIDEOS 2
 
 static struct {
     int width;
     int height;
+    int current_video;
 } data;
 
-static int init(void)
+static const char *intro_videos[NUM_INTRO_VIDEOS] = { "smk/logo.smk", "smk/intro.smk" };
```
bvschaik: Also add `smk/credits.smk` (and increase NUM_VIDEOS to 3)

src/window/intro_video.c
```diff
@@ -4,22 +4,37 @@
...
+
+static int start_next_video(void)
 {
-    if (video_start("smk/intro.smk")) {
-        video_size(&data.width, &data.height);
-        video_init();
-        return 1;
+    while(data.current_video < NUM_INTRO_VIDEOS) {
```
bvschaik: Nitpick: space between `while` and `(`

src/window/intro_video.c
```diff
@@ -33,8 +48,12 @@ static void draw_foreground(void)
 static void handle_mouse(const mouse *m)
 {
     if (m->left.went_up || m->right.went_up || video_is_finished()) {
+        sound_music_stop();
```
bvschaik: There should be no need to stop music here, since the video player manages its own video.

crudelios: It is, otherwise you can hear ROME1 start playing between the videos :/

src/window/intro_video.c
crudelios: Right now the videos are playing in succession. If I do that, credits will play right after the intro videos. Are you sure about that?

bvschaik: Yes. That's what C3 also does.

PR #235, opened 2020-01-15
Prevent gatehouses from adding deleted roads when the map is rotated.

### LINKED ISSUE #233: That annoying gatehouse road!
All of you know that...

step1
Build some gatehouse
_______________________

step2
Then delete that useless part of road
_______________________

step3
Seems to be ok... 
_______________________

step4
Now we need to rotate camera
_______________________

step5
WUT? That piece of sh..road came back! Only way we can handle it is...
_______________________

step6
... to keep that tile busy by some buildings
_______________________

Maybe, it makes sense to fix that oldie one? Thanks!

crudelios:
This should work, even though I can't remember exactly why since I created this branch over an year ago lol

Fix #233

### DIFF
--- a/src/map/orientation.c
+++ b/src/map/orientation.c
@@ -258,7 +258,7 @@ void map_orientation_update_buildings(void)
                     }
                 }
                 map_building_tiles_add(i, b->x, b->y, b->size, image_id, TERRAIN_GATEHOUSE | TERRAIN_BUILDING);
-                map_terrain_add_gatehouse_roads(b->x, b->y, b->subtype.orientation);
+                map_terrain_add_gatehouse_roads(b->x, b->y, 0);
                 break;
             case BUILDING_TRIUMPHAL_ARCH:
                 if (b->subtype.orientation == 1) {


src/map/orientation.c
```diff
@@ -258,7 +258,7 @@ void map_orientation_update_buildings(void)
                     }
                 }
                 map_building_tiles_add(i, b->x, b->y, b->size, image_id, TERRAIN_GATEHOUSE | TERRAIN_BUILDING);
-                map_terrain_add_gatehouse_roads(b->x, b->y, b->subtype.orientation);
+                map_terrain_add_gatehouse_roads(b->x, b->y, 0);
```
bvschaik: Do we need the function call at all?

crudelios: Err.... In my defense, I had about three days experience with Julius when I wrote that change.

bvschaik: No problem, I'm just thinking out loud here :)

crudelios: Turns out the function is actually needed to add the roads *under* the gatehouse, otherwise nothing can cross it, not even destination walkers.

bvschaik: Ah yes, I see. And `0` is not a valid orientation so the extra roads will not be added.

PR #236, opened 2020-01-16
Fix loading empty files

### LINKED ISSUE #234: Folders ending with ".sav" get listed as files and can be opened
If you save a folder as `<any name>.sav`, then that folder will be listed as a valid city in the Load/Save/Delete dialog.

If you then try to open the file, this will happen:

empty_city

As you can see, the game loads an empty city with no terrain and everything set to `0`. The `emperor wrath` message appears because your empire rating is also `0`.

In C3, while the folder appears as a city in the Load/Save/Delete dialog, if you try to open it it will refuse saying that the file does not exist.

Suggestions to fix the bug:
* Do not show folders ending with `.sav` as cities in the Load/Save/Delete dialog. This can easily be done by adding a check for  `entry->d_type == DT_REG` in the function `dir_find_files_with_extension` of file `src/core/dir.c`, which should only check for files anyway.
* This, however, does not prevent the user from manually typing the folder name, thus loading an empty file. To fix that, we should check the return value of the `game_file_load_saved_game` (and friends) in function `button_ok_cancel` of file `src/window/file_dialog.c`, triggering a `File does not exist` message if the function returns `0`.

crudelios:
Fix #234

### DIFF
--- a/src/core/dir.c
+++ b/src/core/dir.c
@@ -91,7 +91,7 @@ const dir_listing *dir_find_files_with_extension(const char *extension)
         const char *name = dir_entry_name(entry);
         if (stat(name, &file_info) != -1) {
             int m = file_info.st_mode;
-            if (S_ISDIR(m) || S_ISCHR(m) || S_ISBLK(m) || S_ISFIFO(m) || S_ISLNK(m) || S_ISSOCK(m)) {
+            if (S_ISDIR(m) || S_ISCHR(m) || S_ISBLK(m) || S_ISFIFO(m) || S_ISSOCK(m)) {
                 continue;
             }
         }
--- a/src/core/dir.c
+++ b/src/core/dir.c
@@ -8,6 +8,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/stat.h>
 
 #ifdef _WIN32
 #include <windows.h>
@@ -36,6 +37,14 @@ static const char *filename_to_utf8(const wchar_t *str)
 #define fs_dir_read readdir
 #define dir_entry_name(d) ((d)->d_name)
 #define dir_entry_close_name(n)
+#endif
+
+#ifndef S_ISLNK
+#define S_ISLNK(m) 0
+#endif
+
+#ifndef S_ISSOCK
+#define S_ISSOCK(m) 0
 #endif
 
 #ifdef __vita__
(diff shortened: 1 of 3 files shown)


src/core/dir.c
```diff
@@ -77,8 +86,15 @@ const dir_listing *dir_find_files_with_extension(const char *extension)
         return &listing;
     }
     fs_dir_entry *entry;
+    struct stat file_info;
     while ((entry = fs_dir_read(d)) && listing.num_files < DIR_MAX_FILES) {
         const char *name = dir_entry_name(entry);
+        if (stat(name, &file_info) != -1) {
```
bvschaik: We do not need stat for this, the entry contains a d_type variable. Or is that not good enough?

src/core/dir.c
```diff
@@ -77,8 +86,15 @@ const dir_listing *dir_find_files_with_extension(const char *extension)
...
+            int m = file_info.st_mode;
+            if (S_ISDIR(m) || S_ISCHR(m) || S_ISBLK(m) || S_ISFIFO(m) || S_ISLNK(m) || S_ISSOCK(m)) {
```
bvschaik: Allow symlinks please.

src/game/file_io.c
```diff
@@ -633,14 +635,18 @@ int game_file_io_read_saved_game(const char *filename, int offset)
     log_info("Loading saved game", filename, 0);
     FILE *fp = file_open(dir_get_case_corrected_file(filename), "rb");
     if (!fp) {
+        log_error("Unable to load game", 0, 0);
```
bvschaik: Add filename as 2nd param so it is included in the log message.

src/game/file_io.c
```diff
@@ -633,14 +635,18 @@ int game_file_io_read_saved_game(const char *filename, int offset)
...
     }
     if (offset) {
         fseek(fp, offset, SEEK_SET);
     }
-    savegame_read_from_file(fp);
+    int result = savegame_read_from_file(fp);
     file_close(fp);
-
+    if (!result) {
+        log_error("Unable to load game", 0, 0);
```
bvschaik: Same as above

bvschaik:
Did you test all campaign saved games from mission1.pak? Those were made using a pre-release version of C3 so they may miss some fields.

src/game/file_io.c
crudelios: I didn't do that because the log will already be:

```
Loading save game: <save game>
ERROR: Unable to load game
```

Writing the file name twice feels redundant...

src/core/dir.c
crudelios: MinGW for windows does not have `d_type`, so I had to resort to `stat`

crudelios:
>Did you test all campaign saved games from mission1.pak?

No I did not! I also don't have them available. Do you have them? If so please send them so I can test.

bvschaik:
They are contained in `mission1.pak` which ismpart of the C3 installation. You can test them by cheating your way through the campaign.

crudelios:
I'm no cheat! I'll do them all fair and square! Give me a month, I'll report back 😆

bvschaik:
Haha, a month is a bit optimistic I guess 😉

crudelios:
Because I'm such a good player, I already legitimately played through both the peaceful and military missions (as well as all the base CCK cities) and can happily report that everything works.

bvschaik:
To be honest... I played through the economic campaign on very easy when I was just learning, I later played through the military one on normal, I believe. I never went back to do them all on hard. I should perhaps do that one day using Julius 😂

crudelios:
Challenge: try getting more than 200 people on very hard :P

Issue #229, opened 2020-01-14
GameSpeed Info


ArtySh0ck:
Please, add some info-string (label on top panel for example...) about **current gamespeed**. Sometimes its difficult to remember which  current game speed is while Pause when it controlled by hotkeys ( "[" and "]" ), as example, and when its unpaused sometimes its unexpectedly too fast. Another reason: it would be just comfortable thing, for real.

411752230:
Agree. Message bar like this would be perfect.
![QQ截图20200115142734](https://user-images.githubusercontent.com/14128007/72410370-3f55b580-37a3-11ea-85da-c4c8c44f3dc2.jpg)

crudelios:
I think this is actually a good idea, and could be placed at the top menu with up/down arrows at large enough resolutions.

crudelios:
I have added the game speed with to buttons to the top menu, here's how it looks:

![](https://user-images.githubusercontent.com/6518876/72721646-709ffe00-3b74-11ea-9304-3d52e6dc129d.JPG)

@jfkonecn has mentioned that the arrow buttons don't look so good in the top menu.

The buttons can be ditched, and an alternative would be for the game to display `Game speed: X%` (perhaps with the darker background like for the pop and Dn) and, upon left clicking on the text, the `Speed settings` window would open.

Any suggestions? [Here's a link to the windows build of my current implementation with the arrow buttons](https://ci.appveyor.com/api/buildjobs/wov4x98gvaag4ksv/artifacts/build%2Fjulius.exe).

bvschaik:
I agree that the arrow buttons look out of place. Just "Game speed X%" would be good, and I think we should keep the regular background: the dark background comes only in a fixed size, and even the English text will overflow.

bvschaik:
Further discussion should continue in #281 .

PR #244, opened 2020-01-18
Game speed is now displayed (#229)

jfkonecn:
Added game speed "warning" message after an unpause event and after the hotkeys '[' ']' are pressed.

Screenshot from 2020-01-18 14-35-03

### DIFF
--- a/src/city/warning.c
+++ b/src/city/warning.c
@@ -5,6 +5,7 @@
 #include "core/string.h"
 #include "core/time.h"
 #include "game/settings.h"
+#include <stdio.h>
 
 #define MAX_WARNINGS 5
 #define MAX_TEXT 100
@@ -80,4 +81,17 @@ void city_warning_clear_outdated(void)
             warnings[i].in_use = 0;
         }
     }
+}
+
+void city_warning_game_speed(void)
+{
+    const uint8_t *langText = lang_get_string(45, 2);
+    struct warning *w = new_warning();
+    if (!w) {
+        return;
+    }
+    w->in_use = 1;
+    // 15 seconds is a long time to display the game speed
+    w->time = time_get_millis() - TIMEOUT_MS * 0.9;
+    sprintf((char*)w->text, "%s %d%%", langText, setting_game_speed());
 }
--- a/src/city/warning.h
+++ b/src/city/warning.h
@@ -67,5 +67,6 @@ const uint8_t *city_warning_get(int id);
 
 void city_warning_clear_all(void);
 void city_warning_clear_outdated(void);
+void city_warning_game_speed(void);
 
 #endif // CITY_WARNING_H
--- a/src/input/hotkey.c
+++ b/src/input/hotkey.c
@@ -37,6 +37,8 @@ static void change_game_speed(int is_down)
         } else {
             setting_increase_game_speed();
         }
+        city_warning_clear_all();
+        city_warning_game_speed();
     }
 }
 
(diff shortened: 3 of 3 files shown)


411752230:
Fantastic! This is exactly what i have been waiting for.

crudelios:
I like this one, but I've been thinking about adding the game speed to the top menu with +/- buttons if the resolution is high enough. Perhaps this warning can be used for low resolutions and the permanent display for higher resolutions?

jfkonecn:
> I like this one, but I've been thinking about adding the game speed to the top menu with +/- buttons if the resolution is high enough. Perhaps this warning can be used for low resolutions and the permanent display for higher resolutions?

I think we should keep this solution, at least at low resolution, since it is the least risky for UI bugs. 

I agree that top menu button are only viable for high resolution screens. 

I think the biggest difficulty with implementing that will be making the interface look like it belongs on the top menu since there are no existing buttons on the top menu. 

I think it might be easier to pull off it on the side menu since there is plenty of space at high resolution. You would also have the added benefit of matching the button based UI which is already apart of the game so it probably wouldn't look out of place.  
Screenshot from 2020-01-19 11-07-27

crudelios:
I was thinking on the top menu above the `Overlays` button, with a darker background like for the `Pop` and `Dn`.

jfkonecn:
> I was thinking on the top menu above the `Overlays` button, with a darker background like for the `Pop` and `Dn`.

I think that would work nicely.

crudelios:
Here's what I came up with (sorry for the huge image size):

top_menu_game_speed

This will always appear on that spot when the screen width `>= 1100`. Note that I haven't submitted a PR yet.

If this is agreed, then the alert could be used for resolution `< 1100`, otherwise the game speed is always visible.

jfkonecn:
> Here's what I came up with (sorry for the huge image size):
> 
> top_menu_game_speed
> 
> This will always appear on that spot (above the `Overlays` panel, regardless of resolution) when the screen width is `>= 1100`. Note that I haven't submitted a PR yet.
> 
> If this is agreed, then the alert could be used for resolution `< 1100`, otherwise the game speed is always visible.

The arrow buttons feel out of place to me. 
I know the arrows are the only button assets available which work for this scenario so I'm not sure how much you can change. 
That said, UI design is not my strength.
Maybe we push this conversation to the issue thread to see what the reporters think?

411752230:
I think message bar is better. Why should the game speed info appears all the time in UI? That's unnecessary. Message bar is not only temporary when you click adjust the game speed by [ and ], but also consistent to the next series of sierra city-building sim games.

crudelios:
Actually many city builders always show the game speed (think SimCity or Cities: Skylines).

I changed the code to remove the buttons (check #258) and besides, with all the screen real estate now available in high resolutions, I don't think it clutters the UI too much.

411752230:
> Actually many city builders always show the game speed (think SimCity or Cities: Skylines).
> I changed the code to remove the buttons (check #258) and besides, with all the screen real estate now available in high resolutions, I don't think it clutters the UI too much.

I understood but I dont agree with that. SimCity and Cities: Skylines are both city-building sim games nowadays with moderen preferences. If julius is a remake i totally agree. But truthfully it is a re-implementation. Adding to the top bar is a bit of aggressive.

C3, a classic game with sequels already existed which have developed many new features for reference, could inherits ready-made advantage from its descendants.

Anyway, it depends on @bvschaik. No matter what the finally applied solution is, I shall love it completely.

crudelios:
You are right, and that's why there is a thin line between what should be added and what shouldn't.

bvschaik:
Closing this because game speed indication will be added to the sidebar in #281

PR #254, opened 2020-01-20
Add roads to granary

### LINKED ISSUE #247: Granary's off road
ambar-road

Actualy, idea belongs by GamerZakh =) but i found it realy cool, all of us knows how ugly looks next roads to granary.
More detailed about an issue linked with timecode here https://youtu.be/OXONBt5U-U0?t=2355

But, i think, by picture on top it would be even more perfectly looks than just a road-direction fix, road must ENTER closely to granary's enter/exit points.

crudelios:
Fix #247

### DIFF
--- a/src/map/image_context.c
+++ b/src/map/image_context.c
@@ -413,7 +413,7 @@ static void set_tiles_road(int grid_offset, int tiles[MAX_TILES])
                 tiles[i] = 1;
             }
         } else if (map_terrain_is(offset, TERRAIN_BUILDING)) {
-            building* b = building_get(map_building_at(offset));
+            building *b = building_get(map_building_at(offset));
             if (b->type == BUILDING_GRANARY) {
                 tiles[i]  = (offset == b->grid_offset + map_grid_delta(1, 0)) ? 1 : 0;
                 tiles[i] |= (offset == b->grid_offset + map_grid_delta(0, 1)) ? 1 : 0;
--- a/src/map/image_context.c
+++ b/src/map/image_context.c
@@ -4,6 +4,7 @@
 #include "city/view.h"
 #include "map/building.h"
 #include "map/elevation.h"
+#include "map/grid.h"
 #include "map/property.h"
 #include "map/terrain.h"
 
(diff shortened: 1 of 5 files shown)


test/sav/sav_compare.c
```diff
@@ -356,6 +356,10 @@ static int is_exception_image_grid(int global_offset)
     if ((v1 == 734 && is_between(v2, 743, 770)) || (v2 == 734 && is_between(v1, 743, 770))) {
         return 1;
     }
+    // road to granary: in julius the roads lead into the granary
+    if (both_between(v1, v2, 591, 657)) {
```
bvschaik: Please split it up into two checks: 591-639 for the paved roads, and 640-657 for the dirt roads. I don't quite like the fact that we're now basically not checking any road graphics at all, so at least make a difference between dirt/paved roads 😉

src/map/image_context.c
```diff
@@ -411,6 +412,14 @@ static void set_tiles_road(int grid_offset, int tiles[MAX_TILES])
                 b->subtype.orientation == 1 + ((i / 2) & 1)) { // 1,2,1,2
                 tiles[i] = 1;
             }
+        } else if (map_terrain_is(offset, TERRAIN_BUILDING)) {
+            building* b = building_get(map_building_at(offset));
```
bvschaik: `building *b` please

test/sav/sav_compare.c
crudelios: Is there a way to checo from a savefile if a granary is adjacent to the tile?

bvschaik: Yes, you would check the same way as in the game but then in the raw data. It's quite cumbersome: first you have to look up the building id for a grid offset in part `"building_grid"`, then you have to look into part `"buildings"` for that building (entries of 128 bytes) whether the type as short at index 10 (0-based) is equal to 71 (granary).

test/sav/sav_compare.c
```diff
@@ -356,8 +356,12 @@ static int is_exception_image_grid(int global_offset)
     if ((v1 == 734 && is_between(v2, 743, 770)) || (v2 == 734 && is_between(v1, 743, 770))) {
         return 1;
     }
-    // road to granary: in julius the roads lead into the granary
-    if (both_between(v1, v2, 591, 657)) {
+    // road to granary: in julius the dirt roads lead into the granary
```
bvschaik: The other way around 😁  640-657 are dirt roads

test/sav/sav_compare.c
crudelios: Thanks for the info, I'll try to implement it.

test/sav/sav_compare.c
crudelios: How do you know that??? I'll try to implement this exception only for adjacent granaries

bvschaik: Use SGReader. If you hover over an image in the left tree, the tooltip will give you "ID 123" and some other info. If you open up C3.sg2, then expand "Land2a.bmp", scroll down to image 44, you'll see that that's the first road tile. It's a paved road and the tooltip says "ID: 591". Scroll down to the first dirt road at image 93, it's ID 640.

test/sav/sav_compare.c
crudelios: Before I commit and push the final changes, the code is now only excempting from tests only the road tiles that are adjacent to granaries, so the normal roads are tested again. Do you still want a separation between the paved roads and dirt roads?

test/sav/sav_compare.c
crudelios: Never thought about using sgreader... Thanks!

test/sav/sav_compare.c
bvschaik: If we only check those roads next to the granary, it's OK to skip the paved/dirt road distinction.

PR #258, opened 2020-01-20
Add a "game speed" information to the top menu

crudelios:
It sits above the `Overlays` panel and only appears when the resolution width is `>= 1150`.

top_menu_game_speed

Right-clicking on it will open the `Speed settings` menu.

### DIFF
--- a/src/widget/top_menu.c
+++ b/src/widget/top_menu.c
@@ -33,7 +33,8 @@ enum {
     INFO_NONE = 0,
     INFO_FUNDS = 1,
     INFO_POPULATION = 2,
-    INFO_DATE = 3
+    INFO_DATE = 3,
+    INFO_GAME_SPEED = 4
 };
 
 static void menu_file_new_game(int param);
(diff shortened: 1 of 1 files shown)


411752230:
Looks good but strange to me. Only one reason: a temporary notifying message without need to be permanent and occupy some space in the top bar.

crudelios:
> Only one reason: a temporary notifying message without need to be permanent and occupy some space in the top bar.

As I said, I disagree and I think a the permanent information is useful.

Also as I said on #244, I think both can be used. Show only notifications on lower resolutions, show the permanent display on higher resolutions.

411752230:
> > Only one reason: a temporary notifying message without need to be permanent and occupy some space in the top bar.
> 
> As I said, I disagree and I think a the permanent information is useful.
> 
> Also as I said on #244, I think both can be used. Show only notifications on lower resolutions, show the permanent display on higher resolutions.

Well. As @bvschaik said, an external program might be added to implement some extra features. Anyway, i think this could be added into the program as one of the optional preferences, while the message bar could be resident for its consistency to the Sierra city-building sequels.

crudelios:
Due to #281, I'm closing this.

PR #260, opened 2020-01-21
Add image displaying whether the trade route to open is land or sea based

crudelios:
On the empire map, when selecting a city to open a new trade route, there will be an icon telling whether this will be a land or sea-based trade route:

trade_route_type

The game will also display a tooltip when the mouse is over the image displaying the trade route type.

For me, this was overdue and saves time saving/reloading.

Also, I tested in the portuguese and german languages and there is no text overlap.


bvschaik:
I was refactoring the generic buttons which caused a conflict. This PR/commit has actually been merged now (7f35d716b9f6231bedabfdf5f6f40a579053d2f4), but Github thinks it hasn't.

PR #273, opened 2020-01-24
Fixed very hard immigration bug (#262)

### LINKED ISSUE #262: Very Hard Immigration Bug
I know the goal of this project to keep the game logic the way it was so that people have the option to play Caesar 3 with only UI updates, but there's a bug which I really hurts the user experience.

On very hard when the population is between 200 and 300. Your citizens will randomly become angry with you and start leaving your city. This results in you having to do silly things to quickly rush 100 people into your city just to get past the bug.

Steps to reproduce
1. Set game to very hard difficultly
2. Have 200 people in your city
See between 11:56 - 15:04
https://youtu.be/DIfvCtI3OXc?t=716

The fix is a one liner
https://github.com/jfkonecn/julius/commit/787080c186b33871c0ebd42c998290f915e238f8

Basically your base housing happiness is 40 when you play on very hard.
Before you hit a population of 300 only base housing happiness is accounted for.
So no matter what you do, your overall happiness is 40. If you happiness is below 50 then people start leaving your city.

This, however, only happens after a population of 200 because the game gives you a boost to prevent this immigration problem from happening to you in the first place. If I had to guess, I think the creators meant to have the boost until 300, but since only crazy people play on very hard no one caught the bug before release.

jfkonecn:
Fix #262

### DIFF
--- a/src/city/sentiment.c
+++ b/src/city/sentiment.c
@@ -194,6 +194,11 @@ void city_sentiment_update(void)
             b->sentiment.house_happiness = default_sentiment;
             if (city_data.population.population < 200) {
                 b->sentiment.house_happiness += 10;
+            } else if (default_sentiment < 50) {
+                // give a boost for very hard so that 
+                // immigration is not prevented simply because
+                // you are between pop 300 and 200
+                b->sentiment.house_happiness += 50 - default_sentiment;
             }
             continue;
         }


crudelios:
Please edit your comment to "Fix #262" or "Fixes #262" so that issue is automatically closed when this PR is merged.

PR #285, opened 2020-01-28
Add metainfo file; FreeDesktop spec compliance fixes

barthalion:


crudelios:
Just for consistency: in my Android build I kept the app name as `bvschaik.julius`. Here the app id is `com.github.bvschaik.julius`.

I know both are independant, but I think the id should still be the same. Which one is best to keep, I have no idea.

barthalion:
I used `com.github.bvschaik.julius` as it's "real" reverse DNS. While nothing enforces it, at Flathub we expect that domain to exist.

barthalion:
Seems AppImage exits with 0 even when it actually failed: https://travis-ci.org/bvschaik/julius/jobs/642802514?utm_medium=notification&utm_source=github_status

bvschaik:
Let's use package name `com.github.bvschaik.julius` for both Android and flatpak, to keep things consistent.

PR #290, opened 2020-01-30
Add version info to main menu

crudelios:
Now you can easily see what version you have:

When the resolution is low and the version appears over the main menu image, you'll see this:

version_inside_image

But normally you'll see this:

version_outside_image

### DIFF
--- a/src/window/main_menu.c
+++ b/src/window/main_menu.c
@@ -1,5 +1,6 @@
 #include "main_menu.h"
 
+#include "core/string.h"
 #include "editor/editor.h"
 #include "game/game.h"
 #include "game/system.h"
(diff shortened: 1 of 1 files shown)


src/window/main_menu.c
```diff
@@ -34,12 +38,33 @@ static generic_button buttons[] = {
 static image_button movie_button =
     {591, 442, 33, 22, IB_NORMAL, 89, 0, button_intro_movie, button_none, 0, 0, 1};
 
+static void draw_version_string(void)
+{
+    char version_string[100] = "v";
+    int version_length = string_length(JULIUS_VERSION);
+    int text_y = screen_height() - 30;
+
+    string_copy(JULIUS_VERSION, version_string + 1, 99);
```
bvschaik: Rebase and use the new `string_from_ascii()` to get rid of the compile warnings.

PR #296, opened 2020-02-04
Redraw backgrounds on window resize

### LINKED ISSUE #66: In-game Menus and Alt-Tab
The following applies to these menus:

* Load Game
* Save Game
* Delete Game
* Quit Game
* Replay Map (also "review your assignment")
* Display Settings
* Sound Settings
* Speed Settings
* Difficulty Settings

If you pull up any of these menus, then alt+tab, then return to game they whole background screen is black. This issue self-corrects after leaving the menu. I would just ignore this, but there are some menus and windows for which this does not happen such as Help and About. I figured there's a chance it's an easy fix.

Also, if you get an in-game message with a video, then alt-tab. The video box is blank upon returning to game. Normally there is a final still image that remains displayed.

All of these are **very** minor in my opinion. If it takes more than a few minutes of poking around, or it's a strange beast to fix, just go ahead and dismiss this ticket.

crudelios:
Redraws the backgrounds (for example, on file dialog window or on most sub-windows of the advisors) when the Julius window is resized, instead of just showing a black background.

Fix #66.

### DIFF
--- a/src/graphics/window.c
+++ b/src/graphics/window.c
@@ -62,6 +62,19 @@ window_id window_get_id(void)
     return current_window->id;
 }
 
+void window_draw_old_behind(void)
+{
+    decrease_queue_index();
+    window_type *window_behind = &window_queue[queue_index];
+    increase_queue_index();
+    if (window_behind->draw_background) {
+        window_behind->draw_background();
+    }
+    if (window_behind->draw_foreground) {
+        window_behind->draw_foreground();
+    }
+}
+
 void window_show(const window_type *window)
 {
     increase_queue_index();
--- a/src/graphics/window.c
+++ b/src/graphics/window.c
@@ -12,6 +12,7 @@ static int queue_index = 0;
 static window_type *current_window = 0;
 static int refresh_immediate;
 static int refresh_on_draw;
+static int underlying_windows_redrawing;
 
 static void noop(void)
 {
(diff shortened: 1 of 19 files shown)


src/graphics/window.c
```diff
@@ -62,6 +62,19 @@ window_id window_get_id(void)
     return current_window->id;
 }
 
+void window_draw_old_behind(void)
```
bvschaik: Naming: I think a better name would be `window_draw_underlying_window()`. Also, put this function at the very bottom so it's after `window_draw` which is related.

src/widget/top_menu.c
```diff
@@ -428,24 +428,28 @@ static void menu_file_exit_game(int param)
 static void menu_options_display(int param)
 {
     clear_state();
+    window_city_show();
```
bvschaik: This makes the "Pause" bar show behind the dialog if you have the game on pause. That does not quite look good...

src/graphics/window.c
crudelios: I didn't quite like the name too. Nice name idea ;)

crudelios:
>Game crashes with a stack overflow segfault when doing the following:

>Go to options -> any option dialog
>Hit Escape

Nice find! It should also happen if you open the load game dialog and hit escape. It should be easy to fix.

PR #297, opened 2020-02-04
Add a "go to problem" button on disaster and invasion messages with videos

crudelios:
Result:

go_to_problem

Also displays a tooltip with the text `Go to problem area` when hovering the button.

### DIFF
--- a/src/window/message_dialog.c
+++ b/src/window/message_dialog.c
@@ -136,6 +136,11 @@ static int resource_image(int resource)
     return image_id;
 }
 
+static int is_problem_message(const lang_message *msg)
+{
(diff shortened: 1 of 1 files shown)


src/window/message_dialog.c
```diff
@@ -450,6 +451,10 @@ static void draw_foreground_video(void)
     video_draw(data.x + 8, data.y + 8);
     image_buttons_draw(data.x + 16, data.y + 408, get_advisor_button(), 1);
     image_buttons_draw(data.x + 372, data.y + 410, &image_button_close, 1);
+    const lang_message *msg = lang_get_message(data.text_id);
+    if (msg->type == TYPE_MESSAGE && (msg->message_type == MESSAGE_TYPE_DISASTER || msg->message_type == MESSAGE_TYPE_INVASION)) {
```
bvschaik: Move the condition within the `if` to a new function `is_problem_message(msg)` and use that both here and in the mouse handler below.

PR #304, opened 2020-02-06
Make construction cost color slightly lighter on central and northern maps

crudelios:
Since I'm color blind, I had trouble reading the construction cost on the northern/central maps as for me it looked almost the same as the terrain color.

This improves things without changing the overall color of the construction cost. It's slightly lighter on northern and central maps, and the same color on desert maps.

However, short of adding a white box like in the tooltips, some color blind people with different types of color blindness will still have trouble reading the cost. Hopefully that's rare enough that it won't matter.

### DIFF
--- a/src/widget/city.c
+++ b/src/widget/city.c
@@ -74,7 +74,7 @@ void widget_city_draw_construction_cost(void)
     color_t color;
     if (cost <= city_finance_treasury()) {
         // Color blind friendly
-        color = (scenario_property_climate() == CLIMATE_DESERT) ? COLOR_ORANGE : COLOR_ORANGE_LIGHT;
+        color = scenario_property_climate() == CLIMATE_DESERT ? COLOR_ORANGE : COLOR_ORANGE_LIGHT;
     } else {
         color = COLOR_RED;
     }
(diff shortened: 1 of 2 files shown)


m-mohr:
Could you share a screenshot, please? :-)

bvschaik:
Top - original, bottom - new.
build-color

src/widget/city.c
```diff
@@ -72,7 +73,8 @@ void widget_city_draw_construction_cost(void)
     set_city_clip_rectangle();
     color_t color;
     if (cost <= city_finance_treasury()) {
-        color = COLOR_ORANGE;
+        // Color blind friendly
+        color = (scenario_property_climate() == CLIMATE_DESERT) ? COLOR_ORANGE : COLOR_ORANGE_LIGHT;
```
bvschaik: Parentheses are not necessary

m-mohr: By the way, do you all know that you can make commitable suggestions here?

For example:
```suggestion
        color = scenario_property_climate() == CLIMATE_DESERT ? COLOR_ORANGE : COLOR_ORANGE_LIGHT;
```

crudelios: Nice! Thanks! Saves some work.

PR #309, opened 2020-02-07
Save full-city screenshots as a bmp by pressing CTRL+F12

### LINKED ISSUE #300: Screenshot Maps
Perhaps add a button or hotkey that screenshots the entire map of the game, somenthing simillar to this:

0s9w1vX

So that people don't have to resort to shit like this:
https://imgur.com/a/AHGRS

crudelios:
Fix #300.

Note: For enourmous maps, the image size will be 135MB. Convert manually to a different image format to reduce file size (palette-based PNG is recommended).

### DIFF
--- a/src/graphics/screenshot.c
+++ b/src/graphics/screenshot.c
@@ -6,8 +6,8 @@
 #include "core/log.h"
 #include "graphics/screen.h"
 #include "graphics/graphics.h"
-#include "map/grid.h"
 #include "graphics/window.h"
+#include "map/grid.h"
 #include "widget/city_without_overlay.h"
 
 #include <stdio.h>
@@ -18,11 +18,16 @@
 #define HEADER_SIZE 26
 #define TILE_X_SIZE 60
 #define TILE_Y_SIZE 30
-#define IMAGE_HEIGHT_CHUNK (2 * TILE_Y_SIZE)
+#define IMAGE_HEIGHT_CHUNK TILE_Y_SIZE
 #define BMP_BITS_PER_PIXEL 24
 #define BMP_BYTES_PER_PIXEL (BMP_BITS_PER_PIXEL / 8)
 
-static const char filename_format[] = "city %Y-%m-%d %H.%M.%S.bmp";
+enum {
+    FULL_CITY_SCREENSHOT = 0,
+    DISPLAY_SCREENSHOT = 1
+};
+
+static const char filename_format[] = "full city %Y-%m-%d %H.%M.%S.bmp";
 
 static void full_city_screenshot(void);
 
@@ -64,13 +69,13 @@ static int create_bmp_chunk(int width, int height)
     return 1;
 }
 
-static const char *generate_filename(void)
+static const char *generate_filename(int city_screenshot)
 {
     static char filename[100];
     time_t curtime = time(NULL);
     struct tm *loctime = localtime(&curtime);
 
-    strftime(filename, 100, filename_format, loctime);
+    strftime(filename, 100, filename_format + city_screenshot * 5 * sizeof(char), loctime);
 
     return filename;
 }
@@ -118,7 +123,7 @@ void graphics_save_screenshot(int full_city)
         return;
     }
 
-    const char *filename = generate_filename();
+    const char *filename = generate_filename(DISPLAY_SCREENSHOT);
     FILE *fp = file_open(filename, "wb");
     if (!fp) {
         log_error("Unable to write screenshot to:", filename, 0);
@@ -159,7 +164,7 @@ static void full_city_screenshot(void)
         return;
     }
 
-    const char *filename = generate_filename();
+    const char *filename = generate_filename(FULL_CITY_SCREENSHOT);
     FILE *fp = file_open(filename, "wb");
     if (!fp) {
         log_error("Unable to write screenshot to:", filename, 0);
@@ -170,7 +175,7 @@ static void full_city_screenshot(void)
     uint8_t header[HEADER_SIZE];
     buffer buf;
     buffer_init(&buf, header, HEADER_SIZE);
-    write_bmp_header(&buf, city_height_pixels);
+    write_bmp_header(&buf, city_height_pixels + TILE_Y_SIZE);
     fwrite(header, 1, HEADER_SIZE, fp);
 
     int new_width = city_width_pixels + (city_view_is_sidebar_collapsed() ? 40 : 160);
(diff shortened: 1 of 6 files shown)


src/graphics/screenshot.c
```diff
@@ -1,20 +1,69 @@
 #include "screenshot.h"
 
+#include "city/view.h"
 #include "core/buffer.h"
 #include "core/file.h"
 #include "core/log.h"
 #include "graphics/screen.h"
 #include "graphics/graphics.h"
+#include "map/grid.h"
+#include "graphics/window.h"
```
bvschaik: Alphabetical order
```suggestion
#include "graphics/window.h"
#include "map/grid.h"
```

src/input/hotkey.c
```diff
@@ -400,6 +400,6 @@ void hotkey_func(int f_number, int with_modifier)
         case 7: system_resize(640, 480); break;
         case 8: system_resize(800, 600); break;
         case 9: system_resize(1024, 768); break;
-        case 12: take_screenshot(); break;
+        case 12: take_screenshot(with_modifier); break;
```
bvschaik: `with_modifier` includes every key (shift, alt, control). This does not work properly on Mac, since F12 is a system key, so I have to press at least one of those to be able to send F12. I can only save full city screenshots now.
Suggestion: add `with_ctrl` as extra parameter and use that.

crudelios:
Question: should the screenshot default to north orientation or keep the current one?

bvschaik:
I would say keep the current orientation.

devnoname120:
I would suggest to encode directly to a lossless compressed format such as PNG. You can use libpng and it's pretty portable so it could be included in the source code (like libsmacker used to be).

crudelios:
Besides having little time during weekends, this change is exposing some bugs regarding camera position and orientation, so I'll fix those as well in this PR.

devnoname120:
libpng is around 300 KB and zlib is around 400 KB (should be way less thanks to link-time optimization: all the decoding logic should be removed from the executable). In my opinion this makes it still worth it. For reference, here is how OpenRCT2 does it:
https://github.com/OpenRCT2/OpenRCT2/blob/83b1d1f72ea1260e45468d5d80225d93068bc8f8/src/openrct2/core/Imaging.cpp#L164-L244

bvschaik:
800KB extra is still quite a lot considering the julius exe is currently 2MB.

But most importantly: taking a full-city screenshot on my relatively powerful mac takes exactly a second for an enormous map. Converting that screenshot to PNG using ImageMagick takes a whopping 10 seconds. I don't want to freeze the game for 11 seconds, or way longer for less powerful devices. And no, multithreading is not an option here.

@crudelios I'm curious to find out which bugs you encountered.

crudelios:
@bvschaik I need to test more by trial and error, but the camera limits seem to be off by 1 tile both in the `x` and `y` axis on `DIR_2_RIGHT` and `DIR_6_LEFT` orientation. If you rotate the map to east and go to the bottom, you'll notice half of the bottomost tile won't show.

This affects the full screenshot, which offsets the whole image.

I suspect it's due to the set starting values in `calculate_lookup` in `city/view.c`, but as I said, I need to test to make sure.

bvschaik:
I have an SSD so writing shouldn't be a problem. I just disabled the `fwrite` call and it didn't make a difference:
```
2020-02-09 18:32:13.661 julius[93229:7208489] INFO: Saving full city screenshot:  city 2020-02-09 18.32.13.bmp
2020-02-09 18:32:14.636 julius[93229:7208489] INFO: Saved full city screenshot:  city 2020-02-09 18.32.13.bmp
```
I think the `pixel()` function being called 46 million times might just be the bottleneck.

Found it: I'm running debug builds by default. If I create a release build: 100ms for an enormous map.

The 10 seconds for encoding the PNG is pure CPU though, so my argument still stands.

crudelios:
> Found it: I'm running debug builds by default. If I create a release build: 100ms for an enormous map.

That makes more sense. So the game should speed up significantly for you. Off topic but that reminds me, android build may be building sdl as debug, that can also be causing the bottleneck at 44100hz, right?

devnoname120:
> 800KB extra is still quite a lot considering the julius exe is currently 2MB.

My bad it's actually around 400 KB and should be around 200 KB after LTO. I think that the slow conversion is due to I/O. If you directly encode the screenshot to PNG it should be way faster. For the PS Vita, considering the fact that I/O is really slow encoding directly to PNG might be faster.

src/input/hotkey.c
crudelios: On original C3, what's supposed to be the modifier hotkey for bookmarks?

I ask because I think a function with the params `with_modifier, with_ctrl` doesn't look good, might as well change it `with_shift<or alt for bookmarks?>, with_ctrl`.

crudelios:
@bvschaik, @devnoname120: Just for the sake of it, I tried adding libpng and zlib to generate the screenshot as PNG.

The results are the following:

* 22 extra files to compile (14 for libpng, 8 for zlib). This may be reduce as PNG read support can be disable, which should also remove the need to compile a few files.
* Executable size increased by about 10% (from 1850KB before to 2070KB after).
* For small maps, saving to PNG is fast. For enourmous maps it isn't (takes about 6 seconds on my Surface Pro 4).
* The PNG image size is about 80% less than BMP (enormous maps went from 136MB to 25MB).

bvschaik:
Ok. Can we merge this PR first and then create another one for PNG support? I'd like to keep changes small.

crudelios:
> Ok. Can we merge this PR first and then create another one for PNG support? I'd like to keep changes small.

Sure, go ahead and merge this. In fact I'm still trying to optimize the PNG, see if I can make it faster by reducing the compression, so it will still take a bit.

src/input/hotkey.c
```diff
@@ -381,25 +381,25 @@ static void handle_bookmark(int number, int with_modifier)
...
 
-static void take_screenshot(void)
+static void take_screenshot(int full_city)
 {
-    graphics_save_screenshot();
+    graphics_save_screenshot(full_city);
 }
 
-void hotkey_func(int f_number, int with_modifier)
+void hotkey_func(int f_number, int with_ctrl)
```
bvschaik: This loses the Shift+F1 to bookmark functionality. Please put the "with_modifier" back (and perhaps rename it to "with_any_modifier")

bvschaik:
> On my latest commit I made it only accept ctrl, I guess I should change this before you merge.

Indeed 😀

crudelios:
Done ;)

PR #317, opened 2020-02-11
Allow cycling temples on build

### LINKED ISSUE #267: Make building temples quicker
Just an idea: I'm always annoyed when building temples. Most of the time I want to build all 5, but then I need to click a lot, first the temple button, then small temple or big temple and then place the actual building. Could that be improved somehow? Maybe by opening the sub-menus by default or so? Not sure what's possible with the menus...

crudelios:
Besides adding an enum to the building groups, this PR adds an option to the new config that allows building one of each temple in succession without having to go the build menu.

Example:

Cycling temples on build

Fixes #267.


src/window/build_menu.c
```diff
@@ -202,38 +220,42 @@ static void button_menu_index(int param1, int param2)
     button_menu_item(button_index_to_submenu_item(param1 - 1));
 }
 
+static int set_submenu_for_type(building_type type)
+{
+    building_menu_group current_menu = data.selected_submenu;
+    switch (type) {
+    case BUILDING_MENU_FARMS:
```
bvschaik: Indent everything within the switch statement please.

src/window/build_menu.c
```diff
@@ -202,38 +220,42 @@ static void button_menu_index(int param1, int param2)
...
-            case BUILDING_MENU_LARGE_TEMPLES:
-                data.selected_submenu = 23;
-                break;
-            case BUILDING_FORT:
-                data.selected_submenu = 24;
-                break;
-            default:
-                break;
-        }
+    if(set_submenu_for_type(type)) {
```
bvschaik: Space after `if`:
```suggestion
    if (set_submenu_for_type(type)) {
```

src/window/build_menu.c
bvschaik: Indent everything within the switch statement please.

src/building/menu.h
```diff
@@ -3,6 +3,40 @@
 
 #include "building/type.h"
 
+typedef enum {
+    BUILDING_MENU_VACANT_HOUSES = 0,
```
bvschaik: Some building types are already prefixed with `BUILDING_MENU_*`, which is quite confusing. Suggestion: call this `BUILD_MENU_*` instead?

src/window/build_menu.c
crudelios: Done already while you were reviewing, sorry.

src/building/menu.h
crudelios: Indeed, I had to add an `ALL` for those cases where the name was the same.

Your idea is simple and effective 👍

src/building/construction.c
```diff
@@ -259,6 +260,7 @@ static int place_reservoir_and_aqueducts(int measure_only, int x_start, int y_st
 void building_construction_set_type(building_type type)
 {
     data.type = type;
+    data.sub_type = 0;
```
bvschaik: Use `BUILDING_NONE` since this variable is of type `building_type`

src/building/construction.c
```diff
@@ -310,12 +318,13 @@ void building_construction_set_type(building_type type)
 void building_construction_clear_type(void)
 {
     data.cost = 0;
+    data.sub_type = 0;
```
bvschaik: Use `BUILDING_NONE` since this variable is of type `building_type`

src/building/menu.c
```diff
@@ -1,13 +1,12 @@
...
 #include "city/buildings.h"
+#include "core/config.h"
 #include "empire/city.h"
 #include "game/tutorial.h"
 #include "scenario/building.h"
 
-#define MAX_BUILDING_ITEMS 30
-
-static const building_type MENU_BUILDING_TYPE[MAX_BUILDING_ITEMS][MAX_BUILDING_ITEMS] = {
+static const building_type MENU_BUILDING_TYPE[BUILD_MENU_MAX][BUILD_MENU_MAX] = {
```
bvschaik: The second dimension should not be BUILD_MENU_MAX, since it contains buildings rather than build menus. If we remove the unused menu items later on, this may cause an out-of-bounds error. Same goes for the enabled variable, and the loop variables.

src/building/menu.h
```diff
@@ -3,6 +3,40 @@
...
+    BUILD_MENU_SMALL_TEMPLES = 22,
+    BUILD_MENU_LARGE_TEMPLES = 23,
+    BUILD_MENU_FORTS = 24,
+    BUILD_MENU_HERDING_POINTS = 25,
+    BUILD_MENU_UNUSED_1 = 26,
+    BUILD_MENU_UNUSED_2 = 27,
+    BUILD_MENU_UNUSED_3 = 28,
+    BUILD_MENU_UNUSED_4 = 29,
+    BUILD_MENU_MAX = 30
+} building_menu_group;
```
bvschaik: Also rename the type to `build_menu_group`?

src/window/build_menu.c
```diff
@@ -51,28 +51,30 @@ static generic_button build_menu_buttons[] = {
...
     data.selected_submenu = submenu;
     data.num_items = building_menu_count_items(submenu);
     data.y_offset = Y_MENU_OFFSETS[data.num_items];
-    if (submenu == 0 || submenu == 1 || submenu == 2) {
-        button_menu_item(0);
-        return 0;
+    if (submenu == BUILD_MENU_VACANT_HOUSE ||
+        submenu == BUILD_MENU_CLEAR_LAND ||
+        submenu == BUILD_MENU_ROAD) {
+            button_menu_item(0);
```
bvschaik: Do not extra indent the content of the if here.

m-mohr:
Hey, thanks a lot for adding this! I'll test it now! 👍

PR #327, opened 2020-02-15
Save screenshots as PNG instead of BMP

crudelios:
Replaces former BMP screenshots.

For normal game screenshots, it should take about 100ms.

For full city shots, it depends on the map size. It goes from about half a second on smaller maps to up to three seconds on the biggest ones, on my Surface Pro 4.

The image size is however reduced by between 66% to 80% when compared to BMP.

There are 12 extra files to compile which don't bloat the executable size too much.

Also, this includes libpng and zlib as external dependencies, both of which have been *heavily* trimmed, with files, comments and unused code removed.


dalerank:
better use https://github.com/nothings/stb/blob/master/stb_image_write.h

crudelios:
I looked and looked and found nothing else. I thought there only was `stb_image_read`!

I'll have to check to see if this has all the options we need, as well as the performance, but if it checks out it's definitely a vetter alternative.

crudelios:
`stb_image_write` does not write a png by row (it requires the full image to be loaded into memory beforehand), so unfortunately it's not an option for our use case.

devnoname120:
In my opinion PNG should be the default. Full city screenshots are not something that are very frequently done so it's reasonable to take 5-10 seconds (except for time lapses, but then BMP will take an ungodly amount of space). I would suggest adding a dialog saying 'Full-city screenshot in progress (can take up to 15 seconds)' if this is possible.

bvschaik:
I tried out `stb_image_write` just for the regular screenshot function. Conclusion: it's 4 times slower than Crudelios' libpng-based solution and the file size is about 50% more than libpng. It does not support streaming so that's indeed a no-go for our use case.

As for PNG as default: I agree.
It's unfortunately not easily possible to add a message. Once you press the key, the game starts working on the screenshot and won't redraw the screen until the screenshot is complete, so any message will not be drawn until after the fact..

Linkers won't do a good job on Julius since there's nothing in the project to configure them to remove dead code. Besides, they won't be able to find all unused code in the PNG library because they're runtime switches. And from what I heard from Crudelios, he made PNG writing quite a bit faster by removing some compression filters.

I don't mind a "butchered" version of libpng, if we can strip it down in such a way that it does not need to be updated. Ideally I'd write my own implementation like for the smk videos, but that's a bit much to ask in this case I guess 😉

crudelios:
> Ideally I'd write my own implementation like for the smk videos, but that's a bit much to ask in this case I guess

Well.... If libpng's zlib compression (which uses deflate) could be changed to use Julius' own `pk_zip` compression (which seems to also use deflate), some extra code could be removed. But that seems like too much work for little gain.

bvschaik:
I think pkzip's deflate is different from zlib's deflate, seeing that pkzip was a proprietary format and all.

> Also: if png is to be set as default, then perhaps bmp isn't really needed? I mean, who would use it, even though it's faster?

I think BMP can be removed, yes.

PR #331, opened 2020-02-19
Doc: Fix MinGW SDL DLL path

CommonLoon102:
The wrong path was mentioned in the docs for the SDL DLLs in case of MinGW builds.


bvschaik:
Thanks!

PR #335, opened 2020-02-19
Make smooth scroll speed closer to original

crudelios:
Fix Keriew/julius#8

Also get rid of some magic numbers.


src/input/scroll.c
```diff
@@ -118,11 +120,22 @@ int scroll_in_progress(void)
...
 
+static int get_scroll_speed_factor(void)
+{
+    int factor = (100 - setting_scroll_speed()) / 10;
+    if (factor < 0) {
+        factor = 0;
+    } else if (factor > 10) {
+        factor = 10;
+    }
+    return factor;
```
bvschaik: You can use `return calc_bound((100 - setting_scroll_speed()) / 10, 0, 10);` for this (include `core/calc.h`)

PR #336, opened 2020-02-19
List 2000 saved games intead of just 200.

### LINKED ISSUE #333: Cannot list more than 207 saved games
When I click "Load saved game" only the first 207 saved games are visible in the list. This is a bug in the original version too.

CommonLoon102:
Fixes #333 

Not sure if you will like it.
I decided to raise the limit only for saved games, because in the original you can still load the 201th by typing its name (as before) but in case of scenarios there is no such option.

### DIFF
--- a/src/core/dir.c
+++ b/src/core/dir.c
@@ -61,13 +61,13 @@ static int listing_initialized = 0;
 static void clear_dir_listing(void)
 {
     if (!listing_initialized) {
-        for (int i = 0; i < DIR_MAX_FILES; i++) {
+        for (int i = 0; i < DIR_MAX_SAVE_FILES; i++) {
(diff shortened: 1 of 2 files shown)


bvschaik:
Thanks for your effort. However, I indeed do not like it for the following reasons:
- Architecture: `core/dir` should not have knowledge of the type of files it can serve up, and certainly not make decisions based on the extension
- Memory: this will consume a fixed 500kB extra memory, even though most users won't have that many savegames or scenarios
- Scenarios: you increase the limit for savegames only, giving as reason that people can type them anyway. Using that reasoning it actually makes sense to increase the limit for scenarios only.
- It's still a limit: this is a temporary fix until someone comes along with 2001 files

I'm going to implement supporting unlimited files in for both scenarios and savegames.

PR #339, opened 2020-02-19
Replay Map Confirmation Window

TheBeardedOne:
(Basically a re-post from the issues section)

Hello!

I have always been annoyed when I am clumsy, and hit "replay map" by accident, without saving!

So I decided it would be a great and minor addition to this project to introduce a replay mission confirmation window!

I have already made the changes, and would love to push them!

Here is a screenshot of it in action!

image
Julius_confirmation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -493,6 +493,7 @@ set(WINDOW_FILES
     ${PROJECT_SOURCE_DIR}/src/window/config.c
     ${PROJECT_SOURCE_DIR}/src/window/difficulty_options.c
     ${PROJECT_SOURCE_DIR}/src/window/display_options.c
+	${PROJECT_SOURCE_DIR}/src/window/replay_confirm.c
     ${PROJECT_SOURCE_DIR}/src/window/donate_to_city.c
     ${PROJECT_SOURCE_DIR}/src/window/empire.c
     ${PROJECT_SOURCE_DIR}/src/window/file_dialog.c
--- a/src/widget/top_menu.c
+++ b/src/widget/top_menu.c
@@ -19,6 +19,7 @@
 #include "widget/city.h"
 #include "window/advisors.h"
 #include "window/city.h"
+#include "window/replay_confirm.h"
 #include "window/difficulty_options.h"
 #include "window/display_options.h"
 #include "window/file_dialog.h"
@@ -385,13 +386,19 @@ static void menu_file_replay_map(int param)
 {
     clear_state();
     building_construction_clear_type();
-    if (scenario_is_custom()) {
+	window_city_show();	// Show the window ... now we have to draw
+	window_confirm_replay_show(window_city_show);
+
+
+	// I need this logic for actually restarting, so don't delte it
+	
+	/*if (scenario_is_custom()) {
         game_file_start_scenario_by_name(scenario_name());
         window_city_show();
     } else {
         scenario_save_campaign_player_name();
         window_mission_briefing_show();
-    }
+    }*/
 }
 
 static void menu_file_load_game(int param)
--- a/src/widget/top_menu.c
+++ b/src/widget/top_menu.c
@@ -386,19 +386,8 @@ static void menu_file_replay_map(int param)
 {
     clear_state();
     building_construction_clear_type();
-	window_city_show();	// Show the window ... now we have to draw
+	window_city_show();
 	window_confirm_replay_show(window_city_show);
-
-
-	// I need this logic for actually restarting, so don't delte it
-	
-	/*if (scenario_is_custom()) {
-        game_file_start_scenario_by_name(scenario_name());
-        window_city_show();
-    } else {
-        scenario_save_campaign_player_name();
-        window_mission_briefing_show();
-    }*/
 }
 
 static void menu_file_load_game(int param)
--- a/src/window/replay_confirm.c
+++ b/src/window/replay_confirm.c
@@ -1,0 +1,84 @@
+#include "replay_confirm.h"
+
+#include "game/settings.h"
+#include "game/system.h"
+#include "graphics/generic_button.h"
+#include "graphics/graphics.h"
+#include "graphics/lang_text.h"
+#include "graphics/panel.h"
+#include "graphics/window.h"
+
+static void button_confirm(int param1, int param2);
+static void button_cancel(int param1, int param2);
+
+static generic_button buttons[] = {
+	{128, 140, 224, 20, button_confirm, button_none, 1, 0},
+	{128, 208, 224, 20, button_cancel, button_none, 1, 0},
+};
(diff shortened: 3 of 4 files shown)


TheBeardedOne:
Was the issue with my implementation the visuals or the way in which I implemented? There was no conflict when I did a pull request but it looks like you just did your own implementation?

bvschaik:
> Was the issue with my implementation the visuals or the way in which I implemented? There was no conflict when I did a pull request but it looks like you just did your own implementation?

I saw your issue, then I thought "ooh, good idea, I have 10 minutes left before bedtime" so I went and implemented it. I'll do a review of your code just for your information, but since this is now in master already I will not merge it.

CMakeLists.txt
```diff
@@ -493,6 +493,7 @@ set(WINDOW_FILES
     ${PROJECT_SOURCE_DIR}/src/window/config.c
     ${PROJECT_SOURCE_DIR}/src/window/difficulty_options.c
     ${PROJECT_SOURCE_DIR}/src/window/display_options.c
+	${PROJECT_SOURCE_DIR}/src/window/replay_confirm.c
```
bvschaik: Keep files in alphabetical order, and keep formatting the same as the source (spaces vs tabs)

src/widget/top_menu.c
```diff
@@ -19,6 +19,7 @@
 #include "widget/city.h"
 #include "window/advisors.h"
 #include "window/city.h"
+#include "window/replay_confirm.h"
```
bvschaik: Alphabetical order

src/widget/top_menu.c
```diff
@@ -385,13 +386,8 @@ static void menu_file_replay_map(int param)
...
     clear_state();
     building_construction_clear_type();
-    if (scenario_is_custom()) {
-        game_file_start_scenario_by_name(scenario_name());
-        window_city_show();
-    } else {
-        scenario_save_campaign_player_name();
-        window_mission_briefing_show();
-    }
+	window_city_show();
```
bvschaik: Keep formatting the same as the source (spaces, not tabs)

src/window/replay_confirm.c
```diff
@@ -0,0 +1,84 @@
+#include "replay_confirm.h"
```
bvschaik: Use spaces, not tabs

src/window/replay_confirm.c
```diff
@@ -0,0 +1,84 @@
...
+static void button_cancel(int param1, int param2);
+
+static generic_button buttons[] = {
+	{128, 140, 224, 20, button_confirm, button_none, 1, 0},
+	{128, 208, 224, 20, button_cancel, button_none, 1, 0},
+};
+
+static struct {
+	int focus_button_id;
+	void(*close_callback)(void);
```
bvschaik: Add a space between `void` and `(`

src/window/replay_confirm.c
```diff
@@ -0,0 +1,84 @@
...
+
+static void draw_foreground(void)
+{
+	graphics_in_dialog();
+
+	outer_panel_draw(96, 80, 18, 12);
+
+	lang_text_draw_centered(0, 2, 128, 94, 224, FONT_LARGE_BLACK);
+
+	label_draw(128, 140, 14, data.focus_button_id == 3 ? 1 : 2);
```
bvschaik: Focus button 3? There are only two buttons.

TheBeardedOne:
I think creating a function in the top menu for a window is not an implementation that follows the design of the original developers. Each window should have its own file, and my implementation follows that pattern.

src/window/replay_confirm.c
```diff
@@ -0,0 +1,84 @@
...
+
+	graphics_reset_dialog();
+}
+
+static void handle_mouse(const mouse *m)
+{
+	if (m->right.went_up) {
+		data.close_callback();
+	}
+	else {
```
bvschaik: `}` and `else {` on the same line please.

src/window/replay_confirm.c
```diff
@@ -0,0 +1,84 @@
...
+{
+	window_type window = {
+		WINDOW_DISPLAY_OPTIONS,
+		window_draw_underlying_window,
+		draw_foreground,
+		handle_mouse
+	};
+	init(close_callback);
+	window_show(&window);
+}
```
bvschaik: Newline at end of file please

src/window/replay_confirm.c
```diff
@@ -0,0 +1,84 @@
...
+
+static void button_cancel(int param1, int param2)
+{
+	data.close_callback();
+}
+
+void window_confirm_replay_show(void(*close_callback)(void))
+{
+	window_type window = {
+		WINDOW_DISPLAY_OPTIONS,
```
bvschaik: Do not reuse window IDs.

src/window/replay_confirm.h
```diff
@@ -0,0 +1,6 @@
+#ifndef WINDOW_CONFIRM_REPLAY_H
```
bvschaik: I think adding a separate window just for displaying a confirmation dialog is a bit overkill, since we already have the popup_dialog for asking OK/cancel questions.

src/window/replay_confirm.h
```diff
@@ -0,0 +1,6 @@
...
+#define WINDOW_CONFIRM_REPLAY_H
+
+void window_confirm_replay_show(void(*close_callback)(void));
```
bvschaik: Naming: file is called `replay_confirm` while the constant and functions are called `confirm_replay`

src/window/replay_confirm.h
TheBeardedOne: If you ever want to change the popup_dialog for any reason, you will also be changing the replay_mission window. If you ever want to change the replay_mission window then you will then have to make a new file anyways. I'm only following the original design pattern.

bvschaik:
> I think creating a function in the top menu for a window is not an implementation that follows the design of the original developers. Each window should have its own file, and my implementation follows that pattern.

Um, I am the original developer of Julius?

I do not agree that every window should have its own file. If that were the case then we would have 10 more popup dialogs, about 20 numeric input dialogs and about 10 different select list windows.

This case is generic enough to warrant a generic solution. This confirm window can now be used in other places as well.

TheBeardedOne:
Ok, that makes sense. This entire time I was under the assumption that you only did not like the visuals since that is only what you mentioned.

Next time to be more friendly, I would advise not just implementing something that someone else has already taken the time to implement.

bvschaik:
> Next time to be more friendly, I would advise not just implementing something that someone else has already taken the time to implement. (And instead, tell them what is wrong)

Point taken. I'm sorry, I'm a bit stressed lately because Julius has been getting so much attention.

PR #340, opened 2020-02-19
Music restart fixes

CommonLoon102:
- Music is not restarted when loading a save from the main menu if the music to be played is the same as the main menu music
- Music sometimes restarted twice when loading a save because `initialize_saved_game()` calls `sound_music_reset()` after `sound_music_update()`
- Music is not started after selecting Replay Map in campaign mode


bvschaik:
Thanks! I thought I heard some skipping in the music last night but couldn't place it.

Issue #274, opened 2020-01-24
Hotkeys to select buildings to construct


pinouchon:
Have you considered adding hotkeys (customisable or not) to select buildings? At the very least for constructing roads, housing, deleting buildings and cancelling the last action. Something similar to the hotkey system you would find in warcraft/starcraft games.

I feel when I play the game that I constantly click those. I think it's very unfortunate that these are not available in the original game.

KillahNeo:
Excellent idea

bvschaik:
Do you have a suggestion for the key mapping? The following letters are already taken:

- `P` pause
- `F` fire overlay
- `D` damage overlay
- `C` crime overlay
- `T` problems overlay
- `W` water overlay
- `L` go to legion

crudelios:
How about using the numpad? It more or less matches the button configuraton... for submenus, use `1` for the first item, `2` for the second... etc

bvschaik:
Who still has a numpad on their laptops these days?

crudelios:
They do? I have zero knowledge about current hotkeys. It's a pity because I alwsys thought it would be a good combination...

bvschaik:
Hmm... just tested: apparently they don't (while they do in C3).
But my first point still stands: how many people play using a full keyboard these days?

crudelios:
> But my first point still stands: how many people play using a full keyboard these days?

I still have full keyboards, but yeah it's not so common.

Another alternative that maps the button positions to the sidebar layout:

`6`, `7`, `8` -> first row of buttons;
`Y`, `U`, `I` ->second row of buttons;
`H`, `J`, `K` ->third row of buttons;
`N`, `M`, `<whatever is to the right of M>` ->fourth row of buttons

Undo, city messages and goto problem would be missing, but those are not commonly used.

For submenus, we could select the position by the corresponding number: `1` for the upmost option, `2` for the second, etc...

bvschaik:
The whole top row is used for shortcuts to the advisors, I forgot to mention those. Also, I don't like a keypad-like mapping on the main keyboard because it's a hell to implement because of different keyboard layouts.

crudelios:
Granted, it's not the best idea.

The overlay keyboards make things hard.

Some ideas though:

* `H` for housing
* `R` for roads
* `Delete` for clear land
* `R` for religion
* `E` for education
*  `A` for administration
* `S` for security
* `I` for industrial
* `U` for undo
* `M` for messages
* `Space` for cycling through trouble spots

[...]

PR #344, opened 2020-02-20
Ctrl+S/O/N shortcuts for Save/Load/New game

queria:
Ctrl+S is standard shortcut for saving (anything) these days (unlike in
old times). So could be good choice for being default save-game hotkey.

This is also related to issue bvschaik/julius#274.
For that I've added also Load and New Game shortcuts.

### DIFF
--- a/src/input/hotkey.c
+++ b/src/input/hotkey.c
@@ -1,15 +1,18 @@
 #include "hotkey.h"
 
+#include "building/construction.h"
 #include "building/type.h"
 #include "city/finance.h"
 #include "city/victory.h"
 #include "city/view.h"
 #include "city/warning.h"
+#include "editor/editor.h"
 #include "figure/formation.h"
 #include "game/orientation.h"
 #include "game/settings.h"
 #include "game/state.h"
 #include "game/system.h"
+#include "game/undo.h"
 #include "graphics/screenshot.h"
 #include "graphics/video.h"
 #include "graphics/window.h"
@@ -19,11 +22,14 @@
 #include "scenario/invasion.h"
 #include "window/advisors.h"
 #include "window/building_info.h"
+#include "window/file_dialog.h"
+#include "window/main_menu.h"
 #include "window/numeric_input.h"
 #include "window/plain_message_dialog.h"
 #include "window/popup_dialog.h"
 #include "window/city.h"
 #include "window/editor/empire.h"
+#include "window/editor/map.h"
 
 static struct {
     int is_cheating;
@@ -160,10 +166,58 @@ static void input_number(int number)
     }
 }
 
+static void new_map(void)
+{
+    if (editor_is_active() && window_is(WINDOW_EDITOR_MAP)) {
+        // TODO: here we need the size selection for triggering window_select_list_show?
+    } else if (window_is(WINDOW_CITY)) {
+        building_construction_clear_type();
+        game_undo_disable();
+        game_state_reset_overlay();
+        window_main_menu_show(1);
+    }
+}
+
+static void load_file(void)
+{
+    if (editor_is_active() && window_is(WINDOW_EDITOR_MAP)) {
+        window_editor_map_show();
+        window_file_dialog_show(FILE_TYPE_SCENARIO, FILE_DIALOG_LOAD);
+    } else if (window_is(WINDOW_CITY)) {
+        building_construction_clear_type();
+        window_city_show();
+        window_file_dialog_show(FILE_TYPE_SAVED_GAME, FILE_DIALOG_LOAD);
+    }
+}
+
+static void save_file(void)
+{
+    if (editor_is_active() && window_is(WINDOW_EDITOR_MAP)) {
+        window_editor_map_show();
+        window_file_dialog_show(FILE_TYPE_SCENARIO, FILE_DIALOG_SAVE);
+    } else if (window_is(WINDOW_CITY)) {
+        window_city_show();
+        window_file_dialog_show(FILE_TYPE_SAVED_GAME, FILE_DIALOG_SAVE);
+    }
+}
+
 void hotkey_character(int c, int with_ctrl, int with_alt)
 {
-    if (with_ctrl && c == 'a') {
-        editor_toggle_battle_info();
+    if (with_ctrl) {
+        switch (c) {
+            case 'a':
+                editor_toggle_battle_info();
+                break;
+            case 'o':
+                load_file();
+                break;
+            case 's':
+                save_file();
+                break;
+            case 'n':
+                new_map();
+                break;
+        }
         return;
     }
     if (with_alt) {


queria:
Load/New game - still need finishing, not tested those even.
Missing load game shortcut from main menu too (or limiting scope of others)?

btw: copy-pasting of all the calls from top_menu* and such into
       hotkeys, does not seem ideal, more likely the desired
       functions covering the overall functionality could be
       exposed from inside top menu and calling them as whole?
       (make them non-static, properly named and add them to .h)

i've opened the pull request more to get comments instead of expecting merge any time soon,
i myself wanted just save game shortcut (maybe even quicksave one, without dialog could be good for me), but not sure if the approach is correct or if different keys should be used etc

CommonLoon102:
> i've opened the pull request more to get comments instead of expecting merge any time soon,
> i myself wanted just save game shortcut (maybe even quicksave one, without dialog could be good for me), but not sure if the approach is correct or if different keys should be used etc

You can open a PR as `draft only` which cannot be merged, for similar reasons you have.

src/input/hotkey.h
```diff
@@ -17,6 +17,9 @@ void hotkey_esc(void);
 void hotkey_page_up(void);
 void hotkey_page_down(void);
 void hotkey_enter(void);
+void hotkey_new_map(void);
```
bvschaik: These functions are only used within `hotkey.c`, so they do not need to be exposed to other code.

src/input/hotkey.c
```diff
@@ -362,6 +381,41 @@ void hotkey_enter(void)
...
+    } else {
+        building_construction_clear_type();
+        window_city_show();
+        window_file_dialog_show(FILE_TYPE_SAVED_GAME, FILE_DIALOG_LOAD);
+    }
+}
+
+void hotkey_save(void)
+{
+    if(editor_is_active()) {
```
bvschaik: Using just `editor_is_active()` will also show the dialog and return to the editor map when you for example price changes in the scenario settings open. I think it's best to check for the editor map window here.
Also, please add a space after the `if`.

src/input/hotkey.c
```diff
@@ -362,6 +381,41 @@ void hotkey_enter(void)
...
+        window_editor_map_show();
+        window_file_dialog_show(FILE_TYPE_SCENARIO, FILE_DIALOG_SAVE);
+    } else {
+        window_city_show();
```
bvschaik: This will also go to the city window when you are on the main menu, on the mission briefing or anywhere else.

src/input/hotkey.c
```diff
@@ -362,6 +381,41 @@ void hotkey_enter(void)
...
+        game_undo_disable();
+        game_state_reset_overlay();
+        window_main_menu_show(1);
+    }
+}
+
+void hotkey_load(void)
+{
+    if(editor_is_active()) {
+        window_editor_map_show();
```
bvschaik: See comment for `hotkey_save()`

src/input/hotkey.c
```diff
@@ -362,6 +381,41 @@ void hotkey_enter(void)
     }
 }
 
+void hotkey_new_map(void)
```
bvschaik: These functions can be static since they're only used from this file.

src/input/hotkey.h
queria: Makes sense, I've just followed the other ones there.
So I guess those should be changed too? (Although in different PR imho)

src/input/hotkey.c
queria: So should be solved with different check for "city window" being active (if possible, similar to editor map window check suggested above)?

queria:
> I agree that copying the code from `top_menu` is not ideal to say the least. We need some central place to set UI state.
> Also, on Mac the commands are not `Ctrl+S` but `Cmd+S`, etc. Should we make an exception for that?

About Mac, maybe the modifier should be more dynamic (i mean the 'with_ctrl' could be true if on Mac and Cmd is down)? Or will support for completely different mappings be needed there? (In style of Cmd+XY instead of Ctrl+S for ex.)

About the central place for UI state, not copying the code, sounds like You have small refactoring in mind, not just exposing top_menu* functions for reuse elsewhere. Any direction You would like me to proceed with this PR? Or to be paused and wait for overhaul of the UI state switching?

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+    } else if (window_is(WINDOW_CITY)) {
+        building_construction_clear_type();
+        window_city_show();
+        window_file_dialog_show(FILE_TYPE_SAVED_GAME, FILE_DIALOG_LOAD);
+    }
+}
+
+static void save_file(void)
+{
+    if (editor_is_active() && window_is(WINDOW_EDITOR_MAP)) {
```
bvschaik: No need to check for `editor_is_active()`, since you can  only be on the `WINDOW_EDITOR_MAP` window when the editor is active.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+        window_editor_map_show();
```
bvschaik: Since you're already on the editor map, there's no need to show it again, you can remove this line.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+        window_file_dialog_show(FILE_TYPE_SCENARIO, FILE_DIALOG_SAVE);
+    } else if (window_is(WINDOW_CITY)) {
+        window_city_show();
```
bvschaik: Same reasoning as above: you don't need to show the city.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+        building_construction_clear_type();
+        game_undo_disable();
+        game_state_reset_overlay();
+        window_main_menu_show(1);
+    }
+}
+
+static void load_file(void)
+{
+    if (editor_is_active() && window_is(WINDOW_EDITOR_MAP)) {
```
bvschaik: No need to check for `editor_is_active()`, since you can  only be on the `WINDOW_EDITOR_MAP` window when the editor is active.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+        window_editor_map_show();
```
bvschaik: Since you're already on the editor map, there's no need to show it again, you can remove this line.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+        window_file_dialog_show(FILE_TYPE_SCENARIO, FILE_DIALOG_LOAD);
+    } else if (window_is(WINDOW_CITY)) {
+        building_construction_clear_type();
+        window_city_show();
```
bvschaik: Same reasoning as above: you don't need to show the city.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
     }
 }
 
+static void new_map(void)
```
bvschaik: To be honest I'm not sure if this Ctrl+N hotkey is desirable: it puts you back to the main menu without warning whatsoever when in the city.

src/input/hotkey.c
```diff
@@ -160,10 +166,58 @@ static void input_number(int number)
...
+{
+    if (editor_is_active() && window_is(WINDOW_EDITOR_MAP)) {
+        window_editor_map_show();
+        window_file_dialog_show(FILE_TYPE_SCENARIO, FILE_DIALOG_LOAD);
+    } else if (window_is(WINDOW_CITY)) {
+        building_construction_clear_type();
+        window_city_show();
+        window_file_dialog_show(FILE_TYPE_SAVED_GAME, FILE_DIALOG_LOAD);
+    }
+}
```
bvschaik: Perhaps also allow Ctrl+O on the main menu for loading a saved game?

src/input/hotkey.c
queria: Exactly my feeling when i got to test the ctrl-o/n keys yesterday.
Btw selecting new game from top menu had same effect, no warning and "suddenly" (ehm) in the main menu, maybe that should be improved?

Not sure if even shortcut for this is really needed/usefull. I have doubts about if someone wants to do 'new game/map' that often for it to deserve shortcut?

CommonLoon102: It doesn't :D

bvschaik: New game from the menu is a more explicit action than pressing a random key combination on the keyboard by accident, so I have no problem with that remaining like it is. I do think we can remove the Ctrl+N shortcut, like you mentioned I don't think it's used very often.

bvschaik:
Hi @queria, are you still planning to update this PR after my review comments?

bvschaik:
PR has been partially merged in f4ccc27deced6b23db6ee4ceffc415226bf93e02

PR #345, opened 2020-02-20
Make green mask more color blind friendly

crudelios:
Makes it slightly darker so it can be seen better against the terrain by color blind people like me.

As usual, it's not perfect - some rare types of color blind people will still have difficulty distinguishing the colors - but it's better than before.

### DIFF
--- a/src/graphics/color.h
+++ b/src/graphics/color.h
@@ -21,7 +21,7 @@ typedef uint32_t color_t;
 #define COLOR_SIDEBAR 0xbdb592
 #define COLOR_NO_MASK 0xffffff
 #define COLOR_MASK_RED 0xff0818 // bit lost 0xf863
-#define COLOR_MASK_GREEN 0x18ff18
+#define COLOR_MASK_GREEN 0x08e718
 #define COLOR_TOOLTIP 0x424242
 #define COLOR_SOLDIER 0xf70000
 #define COLOR_ENEMY_CENTRAL 0x7b0000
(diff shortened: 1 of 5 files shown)


bvschaik:
Do you have before/after examples?

crudelios:
Before:

before

After:

after

I can hardly spot the `before` one, and the `after` is much better.

bvschaik:
Is this only a problem when painting terrain in the editor? This same green mask is also used for placing buildings in the game itself, and I'd like to keep the "C3" green there.

This darker green is less visible for me on a northern map... maybe we should tweak the color for the editor. Perhaps a lighter green will stand out more?

CommonLoon102:
This obviously should be a setting. A lot of games have a "color blind" setting, and some of them not even one but 3, because there are different kind of color blindnesses.

crudelios:
I can do as I did with the cost and add a new color that is only shown on central maps and not on buildings.

Regarding adding a setting: if the color change is reduced to only being used on a single terrain type (and the old color is kept for everything else), I don't think would be necessary to add a setting just for that.

CommonLoon102:
The long term solution would be a setting. There will be other cases in the future where altering the original behavior will be needed for colorblind people. And there are different kind of color blindnesses, which complicates the things even further.

Personally, I don't want to have the colors to be changed anywhere, not even a little. I'm not colorblind and I like it as it is. There should be a setting.

crudelios:
How about this color? (note: this is a new `image_draw_blend_alpha` implementation with color set to `COLOR_MASK_GREEN` and transparency set to `0x48`).

city 2020-02-20 20 14 05

PR #358, opened 2020-03-02
Add reservoir, fountain and well range when constructing

### LINKED ISSUE #343: Show future reservoir (and fountain) supply area
In the original it is annoying to count the tiles when building a new reservoir. I have to save the game, imagine where should I put the reservoir, and if my calculation was wrong, I have to load the game and place it again and see if it provides water to that area where I meant it to be. So, it would be very helpful if the game would show it in advance, if the green reservoir image has water filled in it (or even if it doesn't, because you can connect an aqueduct later), the future pipes should be shown as green too. Or just show the covered area somehow else (if not pipes).

Something similar should be done to fountains too. Show the area with some color which will have water if you place the fountain (only if the fountain has water in it).

crudelios:
Example:

city 2020-03-02 08 32 34

Fix #343.

It shows the range whether the reservoir is next to water or not, and whether the fountain has reservoir access or not. Desert climate was also taken into account.

Also slightly improves texture alpha blending performance.


src/widget/city_building_ghost.c
```diff
@@ -351,6 +399,12 @@ static void draw_draggable_reservoir(const map_tile *tile, int x, int y)
             draw_flat_tile(x + X_VIEW_OFFSETS[i], y + Y_VIEW_OFFSETS[i], COLOR_MASK_RED);
         }
     } else {
+        if(!building_construction_in_progress() || draw_later) {
+            if(draw_later) {
```
bvschaik: Space after `if` please.
```suggestion
        if (!building_construction_in_progress() || draw_later) {
            if (draw_later) {
```

crudelios:
Before doing that, I'm trying to improve the performance. It should be slightly faster, but no miracle yet...

bvschaik:
Even without the performance hit I'd rather have a setting for it.

crudelios:
Fair enough. I'll do both.

CommonLoon102:
> 
> 
> Even without the performance hit I'd rather have a setting for it.

The question is, should it be enabled or disabled by default?

crudelios:
Can you test the performace of the new code? I also think I found a slight optimization to all `draw_compressed` routines.

> should it be enabled or disabled by default?

Since we're adding a setting that, when enabled, makes the game look different, I'd say it should be disabled by default.

bvschaik:
> The question is, should it be enabled or disabled by default?

The default settings should look and act like plain Caesar 3, so disabled by default.

I'll have a look at the performance later today when I get back home.

src/graphics/image.c
```diff
@@ -307,22 +312,21 @@ static void draw_compressed_blend_alpha(const image *img, const color_t *data, i
...
             } else {
                 data += b;
-                color_t *dst = graphics_get_pixel(x_offset + x, y_offset + y);
                 if (unclipped) {
-                    x += b;
                     while (b) {
-                        color_t d = *dst;
-                        *dst = MIX(color, d, alpha, 0) | MIX(color, d, alpha, 8) | MIX(color, d, alpha, 16);
-                        dst++;
+                        color_t d = dst[x];
```
bvschaik: Did you run a profiler to see whether this really makes a difference?

bvschaik: I just ran a simple test program to verify this. The old implementation using pointers is 5-10% _faster_ than the new implementation using array indexing, both without (`-O0`) and with (`-O3`) compiler optimizations.

crudelios:
Really? I thought avoiding the constant calls to `graphics_get_pixel` would make things faster.

My visual studio profiler always crashes so I have to go by feeling :/

theradeonxt:
Isn't it because some math done in that MIX   step? Also the way it is done per each byte, it's directly affevted by image size.
Even the get pixel function can be optimmized by compiler.

crudelios:
Even though my profiler doesn't work, I tested this manually by using the windows-specific `QueryPerformanceCounter` and measuring the time between the start and the end of the function `city_view_foreach_tile_in_radius`.

I tested both the new mixing function and the array access independently.

Tests were performed using `mingw-w64` with `-O3` optimization. The same spot was used to display the building ghost.

For simplicity, I tested both the new and the old mixing functions using the original `dst` pointer dereference from the `master` branch.

Here are the times (averaged from a 10 second run each):

* Old alpha mixing function: 2.04ms
* New alpha mixing function: 1.56ms

Next, I changed the `dst` access when using the new alpha mixing function:

* Old style (from `master`): 1.56ms
* Array indexing (from most recent commit): 1.52ms
* New style: one call to `graphics_get_pixels` per line, `dst` incremented: 1.46ms

Curiously enough, I get a performance increase using the new code. However there is indeed more to gain by ditching the array indexing and using pointer incrementation along with a single `graphics_get_pixel` per row.

I'll also test at home where I have more available compilers for Windows.

crudelios:
This latest implementation is the fastest one for me across all compilers. Can you test, please?

src/graphics/color.h
```diff
@@ -22,6 +22,7 @@ typedef uint32_t color_t;
 #define COLOR_NO_MASK 0xffffff
 #define COLOR_MASK_RED 0xff0818 // bit lost 0xf863
 #define COLOR_MASK_GREEN 0x18ff18
+#define COLOR_ALPHA_BLUE 0x665555ff
```
bvschaik: I played around a bit with the colors, since this one looks purple, especially on desert maps.
```suggestion
#define COLOR_ALPHA_BLUE 0x663377ff
```
Current color:
color-old

New color:
color-new

bvschaik:
> This latest implementation is the fastest one for me across all compilers. Can you test, please?

Tested it, looks good 👍  Just one comment about the color.

crudelios:
Out of curiosity I tried changing the number/order of `graphics_get_pixel` calls for `draw_compressed` to be similar to this implementation and the performance was marginally worse. So I don't think I'll be trying anything else regarding performance improvements.

bvschaik:
I just did some testing with using premultiplied alpha, but performance also got worse, so I guess this is as good as we'll get using the current CPU-based setup. Thanks for profiling!

PR #359, opened 2020-03-02
Android build preparation: disable setting windowed mode

crudelios:
Already works for Vita and Switch.

This is a change I'm unsure about as it adds a `definitions.h` file to `platform`, allowing global platform-related settings. It's used both for forcing full-screen and for implementing virtual keyboard support, which is coming next.

This also implements a `disabled` parameter to the top menu. When fullscreen is forced, the `Display settings` menu is disabled (grayed out) and the keyboard shortcuts for resizing the window don't work.

### DIFF
--- a/src/city/view.c
+++ b/src/city/view.c
@@ -1,11 +1,11 @@
 #include "view.h"
 
 #include "core/direction.h"
+#include "graphics/menu.h"
 #include "map/grid.h"
 #include "map/image.h"
 #include "widget/minimap.h"
 
-#define MENUBAR_HEIGHT 24
 #define TILE_WIDTH_PIXELS 60
 #define TILE_HEIGHT_PIXELS 30
 #define HALF_TILE_WIDTH_PIXELS 30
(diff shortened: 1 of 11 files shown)


bvschaik:
Adding platform configuration to the core code: if at all possible, please don't.
Since `game/system.h` is already used to call display-related functions, I'd rather have a function `system_can_resize()` in there which returns 0 in `src/platform/vita/screen.c` and `src/platform/switch/screen.c`, and 1 in `src/platform/screen.c`. You can then call this function in `top_menu_window_show` for both the main game and the editor.

Instead of disabling the display settings, I think it makes more sense to just remove the entry. Although that will probably mean a bit more code since the index of the menu is now closely tied to the position.

crudelios:
So if I understand what you're saying correctly, I fully replace the `platform/definitions.h` with that implementation?

Regarding the menu removal: the only thing I'm not so sure about is the fact that in order to remove the menu, we'll have to check for `system_can_resize` directly in the menu drawing routine. That's a bit spaghetti.

bvschaik:
> So if I understand what you're saying correctly, I fully replace the `platform/definitions.h` with that implementation?

Yes, exactly.

> Regarding the menu removal: the only thing I'm not so sure about is the fact that in order to remove the menu, we'll have to check for `system_can_resize` directly in the menu drawing routine. That's a bit spaghetti.

You do the setup from `static void top_menu_window_show(void)` using an `init()` function (see `init_from_settings()` in `top_menu.c` for reference), so it's not a drawing routine: it's a function for setting up the "window". I see no problem with that.

PR #360, opened 2020-03-02
Android build preparation: Improved touch support

### LINKED ISSUE #269: Touch support improvements
@m-mohr, continuing the  discussion from #195:

> Hmm, so let's think about best gestures. I'm also thinking about how other apps do that.
> 
> So what do we have on a computer with a mouse:
> 
> movement (scrolling): Taken from some maps: Use two fingers and move/scroll. Alternative: One tap, keep pressing and move.
> left mouse click: Single tap
> left mouse pressed for longer time (e.g. building houses): Double tap, keep pressing after the second tap and move.
> right mouse click (exit menus, delete messages, reset tool, ...): Double tap? Or some gesture? In menus: Use hardware back button from Android? Could we use that "x" button left to the messages for tool reset?
> middle mouse scroll (not sure, is that used?)
> more?
> These gestures are what I would expect intuitievely.
> What do you think @crudelios ?

crudelios:
Fix #269.

**Note:** Due to the fact that I implemented double-tap as a way to exit most menus, I had to change the exit code for nearly all the windows in the game. Hence the huge amount of changed files.

Here are some of the new features:

* Cancel build button. Only appears when touch is being used.
* Double tap to exit most windows
* Increased top menu options hitbox to make it easier to select the menu option
* Keep overlay submenu open on main button click. Works for both touch and mouse
* Single tap on city without any construction active to show building info
* Scrolling no longer stops the city when either smooth scrolling is active or touch is used to scroll
* Complete overhaul of construction using touch
* Support for touch on the map editor

**New way to construct using touch**

There are two types of construction: some can be click-dragged, others cannot (they are single buildings).
Examples of constructions that can be click-dragged: roads, houses, walls, reservoirs, clear land.

There are different (but somewhat similar) ways to build a simple and a click-draggable construction.

__For simple builings:__ tap anywhere on the map to show the ghost. Tap again on the same spot to confirm construction or somewhere else to move the ghost. If you press-drag, the building ghost will move with your finger.

__For click-draggable buildings:__ Press to set the starting position. You can then either drag the finger to extend the construction range, or immediately lift the finger and tap somewhere else to extend the construction range. To confirm the construction, tap at the end of the range (the tile you last released the finger at).

If you are not happy with the construction range, instead of confirming it as said above, you can do either of the following:
* Press and drag at the end of the range to change the end of the range
* Tap somewhere else to set the end of the range to that spot
* Press and drag somewhere else to restart the construction range from the pressed spot

When either type of construction is active, if you press-drag next to the edges of the screen, the city will scroll. The closer to the edge your finger is, the faster the scroll will be.

**Cancelling a build**

As mentioned earlier, there is now a button to cancel the build. As an alternative, you can still tap with a second finger to cancel the build. Both options work exactly the same way.

For simple buildings or click-draggable buildings that are not being constructed, clicking the button cancels the selected construction. That means you can navigate the city normally and you'll have to select the proper build button on the sidebar if you want to restart construction.

If you are in the middle of a click-draggable construction, pressing the button will only remove the pending construction. If you don't wish to build anything, you must press the cancel button again.

**Using the map editor**

The map editor works exactly the same way as normal city construction when creating buildings or roads.
For map points, the game works like creating a simple building.
For tiles that use a brush, a simple press/drag and drag will immediately apply the terrain change without having to confirm by clicking again.

### DIFF
--- a/src/window/building_info.c
+++ b/src/window/building_info.c
@@ -521,41 +521,23 @@ static void draw_foreground(void)
     }
 }
 
-static void handle_mouse(const mouse *m)
+static int handle_specific_building_info_mouse(const mouse *m)
 {
-    if (m->right.went_up) {
-        window_city_show();
-        return;
-    }
-    // general buttons
-    if (context.storage_show_special_orders) {
-        int y_offset = window_building_get_vertical_offset(&context, 28);
-        image_buttons_handle_mouse(m, context.x_offset, y_offset + 400, image_buttons_help_close, 2, &focus_image_button_id);
-    } else {
-        image_buttons_handle_mouse(
-            m, context.x_offset, context.y_offset + 16 * context.height_blocks - 40,
-            image_buttons_help_close, 2, &focus_image_button_id);
-    }
-    if (context.can_go_to_advisor) {
-        image_buttons_handle_mouse(
-            m, context.x_offset, context.y_offset + 16 * context.height_blocks - 40,
-            image_buttons_advisor, 1, 0);
-    }
     // building-specific buttons
     if (context.type == BUILDING_INFO_NONE) {
-        return;
+        return 0;
     }
     if (context.type == BUILDING_INFO_LEGION) {
-        window_building_handle_mouse_legion_info(m, &context);
+        return window_building_handle_mouse_legion_info(m, &context);
     } else if (context.figure.drawn) {
-        window_building_handle_mouse_figure_list(m, &context);
+        return window_building_handle_mouse_figure_list(m, &context);
     } else if (context.type == BUILDING_INFO_BUILDING) {
         int btype = building_get(context.building_id)->type;
         if (btype == BUILDING_GRANARY) {
             if (context.storage_show_special_orders) {
-                window_building_handle_mouse_granary_orders(m, &context);
+                return window_building_handle_mouse_granary_orders(m, &context);
             } else {
-                window_building_handle_mouse_granary(m, &context);
+                return window_building_handle_mouse_granary(m, &context);
             }
         } else if (btype == BUILDING_WAREHOUSE) {
             if (context.storage_show_special_orders) {
@@ -564,6 +546,32 @@ static void handle_mouse(const mouse *m)
                 window_building_handle_mouse_warehouse(m, &context);
             }
         }
+    }
+    return 0;
+}
+
+static void handle_mouse(const mouse *m)
+{
+    int handled = 0;
+    // general buttons
+    if (context.storage_show_special_orders) {
+        int y_offset = window_building_get_vertical_offset(&context, 28);
+        handled = image_buttons_handle_mouse(m, context.x_offset, y_offset + 400, image_buttons_help_close, 2, &focus_image_button_id);
+    } else {
+        handled = image_buttons_handle_mouse(
+                      m, context.x_offset, context.y_offset + 16 * context.height_blocks - 40,
+                      image_buttons_help_close, 2, &focus_image_button_id);
+    }
+    if (context.can_go_to_advisor) {
+        handled = image_buttons_handle_mouse(
+                      m, context.x_offset, context.y_offset + 16 * context.height_blocks - 40,
+                      image_buttons_advisor, 1, 0);
+    }
+    if (!handled) {
+        handled = handle_specific_building_info_mouse(m);
+    }
+    if (!handled && (m->right.went_up || (m->is_touch && m->left.double_click))) {
+        window_city_show();
     }
 }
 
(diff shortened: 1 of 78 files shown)


src/window/city.c
```diff
@@ -63,13 +63,27 @@ static void draw_paused_and_time_left(void)
     }
 }
 
+static void draw_cancel_construction(void)
+{
+    if (!mouse_get()->is_touch || !building_construction_type()) {
+        return;
+    }
+    uint8_t cancel_button_text[] = { 'X', 0 };
```
theradeonxt: Maybe simplified?
```suggestion
    uint8_t* cancel_button_text = "X";
```

crudelios: You're looking at an outdated commit, beware

src/window/building_info.c
```diff
@@ -565,6 +547,32 @@ static void handle_mouse(const mouse *m)
...
+    }
+    if (context.can_go_to_advisor) {
+        handled = image_buttons_handle_mouse(
+                      m, context.x_offset, context.y_offset + 16 * context.height_blocks - 40,
+                      image_buttons_advisor, 1, 0);
+    }
+    if (!handled) {
+        handled = handle_specific_building_info_mouse(m);
+    }
+    if (!handled && (m->right.went_up || (m->is_touch && m->left.double_click))) {
```
theradeonxt: I think second part after && can be moved inside the same if above

src/window/editor/map.c
```diff
@@ -22,10 +24,25 @@ static void draw_background(void)
     widget_top_menu_editor_draw();
 }
 
+static void draw_cancel_construction(void)
+{
+    if (!mouse_get()->is_touch || !editor_tool_is_active()) {
```
theradeonxt: Maybe simplified?
```suggestion
   if (!(mouse_get()->is_touch && editor_tool_is_active())) {
```

src/window/building_info.c
crudelios: It can't as `handled` may be changed between both `if`s.

theradeonxt: You're right, ignore this

crudelios: No problem, thanks for the review ;)

src/window/city.c
```diff
@@ -63,13 +64,26 @@ static void draw_paused_and_time_left(void)
...
+}
+
 static void draw_foreground(void)
 {
     widget_top_menu_draw(0);
     window_city_draw();
     widget_sidebar_draw_foreground();
     if (window_is(WINDOW_CITY) || window_is(WINDOW_CITY_MILITARY)) {
         draw_paused_and_time_left();
+        draw_cancel_construction();
```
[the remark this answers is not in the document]
crudelios: No, I just used the already existant `if` ;)

src/window/config.c
```diff
@@ -159,9 +159,13 @@ static void draw_foreground(void)
...
     const mouse *m_dialog = mouse_in_dialog(m);
-    generic_buttons_handle_mouse(m_dialog, 0, 0, checkbox_buttons, NUM_CHECKBOXES, &data.focus_button);
-    generic_buttons_handle_mouse(m_dialog, 0, 0, bottom_buttons, NUM_BOTTOM_BUTTONS, &data.bottom_focus_button);
-    generic_buttons_handle_mouse(m_dialog, 0, 0, &language_button, 1, &data.language_focus_button);
+    int handled = 0;
+    handled |= generic_buttons_handle_mouse(m_dialog, 0, 0, checkbox_buttons, NUM_CHECKBOXES, &data.focus_button);
+    handled |= generic_buttons_handle_mouse(m_dialog, 0, 0, bottom_buttons, NUM_BOTTOM_BUTTONS, &data.bottom_focus_button);
+    handled |= generic_buttons_handle_mouse(m_dialog, 0, 0, &language_button, 1, &data.language_focus_button);
+    if (!handled && (m->right.went_up || (m->is_touch && m->left.double_click))) {
+        window_main_menu_show(0);
```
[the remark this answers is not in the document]
crudelios: I thought you had forgotten to add it, but I should know better. I constantly right-click to exit without changing anything and it feels right to me. Much like clicking the `X` on a settings window in your OS.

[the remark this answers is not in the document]
bvschaik: The only concern I have is that people might accidentally double-tap when changing a lot of settings quickly...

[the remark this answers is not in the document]
crudelios: What if I keep just the right click for this one?

src/window/editor/map.c
crudelios: Yeah, I think that as well.

src/window/config.c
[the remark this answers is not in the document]
bvschaik: That's fine by me.

PR #364, opened 2020-03-04
Show construction size when creating draggable buildings

crudelios:
Feature:

Construction size

This is added as a config setting.


src/widget/city.c
```diff
@@ -58,30 +58,41 @@ void widget_city_draw_for_figure(int figure_id, pixel_coordinate *coord)
...
     if (!building_construction_in_progress()) {
         return;
     }
     if (scroll_in_progress()) {
         return;
     }
+    int size_x, size_y;
     int cost = building_construction_cost();
-    if (!cost) {
+    int size = building_construction_size(&size_x, &size_y);
```
bvschaik: Rename `size` to `has_size` to indicate that it's a boolean.

src/building/construction.c
```diff
@@ -334,6 +335,26 @@ int building_construction_cost(void)
...
+    }
+    *x = data.end.x - data.start.x;
+    *y = data.end.y - data.start.y;
+    if (*x < 0) {
+        *x = -*x;
+    }
+    if (*y < 0) {
+        *y = -*y;
+    }
+    (*x)++;
```
bvschaik: I prefer using local variables to calculate the value and then at the very end assign them to the passed pointers, to prevent this pointer "magic".

PR #365, opened 2020-03-05
Android build preparation: Handle virtual keyboard

crudelios:
This only works for platforms which use a virtual keyboard, which are Switch, Vita and Android.

This allows the virtual keyboard to display automatically when the input box is pressed.

It may already work with Vita and Switch, but it's untested. On the worst case it should behave exactly like until now.

### DIFF
--- a/src/platform/virtual_keyboard.c
+++ b/src/platform/virtual_keyboard.c
@@ -4,39 +4,16 @@
 
 #include "SDL.h"
 
-static int active;
-
-void platform_virtual_keyboard_start(void)
-{
-    if (system_use_virtual_keyboard()) {
-        SDL_StopTextInput();
-    }
-    active = 1;
-}
-
-void platform_virtual_keyboard_resume(void)
-{
-    active = 1;
-}
-
-void platform_virtual_keyboard_stop(void)
-{
-    if (system_use_virtual_keyboard()) {
-        SDL_StopTextInput();
-    }
-    active = 0;
-}
-
 void platform_virtual_keyboard_show(void)
 {
-    if (active && system_use_virtual_keyboard()) {
+    if (system_use_virtual_keyboard()) {
         SDL_StartTextInput();
     }
 }
 
 void platform_virtual_keyboard_hide(void)
 {
-    if (active && system_use_virtual_keyboard()) {
+    if (system_use_virtual_keyboard()) {
         SDL_StopTextInput();
     }
 }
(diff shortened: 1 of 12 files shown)


src/window/new_career.c
```diff
@@ -66,7 +69,9 @@ static void handle_mouse(const mouse *m)
         window_go_back();
     }
```
bvschaik: Shouldn't we `return` here?

src/widget/input_box.h
```diff
@@ -0,0 +1,18 @@
...
+#define WIDGET_INPUT_BOX_H
+
+#include "input/mouse.h"
+
+#define INPUT_BOX_BLOCK_SIZE 16
+
+typedef struct {
+    int x;
+    int y;
+    int w;
```
bvschaik: Suggestion: call `w` and `h`, `width_blocks` and `height_blocks`, to make it clear that the unit is blocks instead of pixels.

src/widget/input_box.c
```diff
@@ -0,0 +1,29 @@
...
+    int selected = is_mouse_inside_input(m, box);
+    if (!platform_virtual_keyboard_showing() && selected) {
+        platform_virtual_keyboard_show();
+    } else if (platform_virtual_keyboard_showing() && !selected) {
+        platform_virtual_keyboard_hide();
+    }
+    return selected;
+}
+
+void input_box_draw(const input_box *box)
```
bvschaik: I prefer having the `draw` function above the `handle_mouse` function: you have to draw first before you can interact with it. Same goes for the header file.

src/platform/virtual_keyboard.c
```diff
@@ -0,0 +1,47 @@
...
+    active = 1;
+}
+
+void platform_virtual_keyboard_stop(void)
+{
+    if (system_use_virtual_keyboard()) {
+        SDL_StopTextInput();
+    }
+    active = 0;
+}
```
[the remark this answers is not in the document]
crudelios: You're right, this is overcomplicated. `start` and `stop` don't seem needed, `show` and `hide` should be enough.

src/window/new_career.c
crudelios: I added a `return` on the latest commit and it doesn't show up in the changes, has that happened to you before?

devnoname120:
I've just tested and it doesn't work on the PS Vita. No regression though, the old behavior is kept.

PR #373, opened 2020-03-13
Drop unnecessary check on unsigned values

janisozaur:
Both figure.x and y are of unsigned char type and cannot be less than 0

https://github.com/bvschaik/julius/blob/b8e0fede7cc88b35e9bf6a0af978458f0aa5650e/src/figure/figure.h#L32-L33


bvschaik:
Thanks. I think that's a left-over from the time when I didn't know which `char`s were signed and which ones unsigned.

Issue #374, opened 2020-03-14
Add Alt+Return to switch between windowed mode and full screen mode


devnoname120:
Currently it's bound to <kbd>F6</kbd> but most games use <kbd>Alt</kbd> <kbd>Return</kbd> so for consistency reasons it would be nice to add/replace the hotkey.

PR #376, opened 2020-03-15
Fullscreen toggle with Alt+Return

devnoname120:
Close https://github.com/bvschaik/julius/issues/374.

### DIFF
--- a/src/platform/keyboard_input.c
+++ b/src/platform/keyboard_input.c
@@ -74,8 +74,9 @@ void platform_handle_key_down(SDL_KeyboardEvent *event)
     switch (event->keysym.sym) {
         case SDLK_RETURN:
         case SDLK_KP_ENTER:
-            if(!is_alt_down(event))
+            if (!is_alt_down(event)) {
                 keyboard_return();
+            }
             hotkey_enter(is_alt_down(event));
             break;
         case SDLK_BACKSPACE:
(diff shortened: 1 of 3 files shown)


src/platform/keyboard_input.c
```diff
@@ -74,8 +74,9 @@ void platform_handle_key_down(SDL_KeyboardEvent *event)
     switch (event->keysym.sym) {
         case SDLK_RETURN:
         case SDLK_KP_ENTER:
-            keyboard_return();
-            hotkey_enter();
+            if(!is_alt_down(event))
+                keyboard_return();
```
bvschaik: Formatting: space after `if` and always use curly braces, even though it's only one statement, please.

Functional: why should Alt+Enter not be recognised as an enter on the keyboard? In other words, why would we need the if at all?

devnoname120: Do you want me to move the check inside `keyboard_return()`? None of the `keyboard_*` functions seemed to do any logic, which is why I didn't put the check in there.

If we remove the if completely, then pressing <kbd>Alt</kbd> <kbd>Return</kbd> to switch to full screen will also confirm some dialogs. I think that this is an unwanted behavior.

For example:
https://github.com/bvschaik/julius/blob/8076f54afa7107ceae177759e8a46a41758e0411/src/window/new_career.c#L78

bvschaik: > If we remove the if completely, then pressing Alt Return to switch to full screen will also confirm some dialogs. I think that this is an unwanted behavior.

Fair point. It's ok like this then :)

devnoname120: By the way, do you use any specific coding style? It would be nice to add a `.clang-format` file at the root of the repository so that IDEs such as VSCode and CLion can use it to automatically format the code.

There are some built-in coding styles, and you can build your own interactively here: https://zed0.co.uk/clang-format-configurator/ (you can use as a base the style that is the closest to your rules).

PR #378, opened 2020-03-19
Military sidebar and other military changes

### LINKED ISSUE #320: Maybe more visibility when controlling legions?
image

Example:
Currently i need to choose one legion, we say its Rabbits, to move or something. BUT, u know, clicking there the game somehow gives us random (i know it isnt random, it depends on "legion layer", and time when current legion came to point, earlier or later) legion.

Idea: maybe game should somehow give us a highlight of choosen legion when it clicked once? We click on that mess of legions and TROOPS are, maybe, come greener? And then we decide, to do something with it, or just right-click it to drop and try another to pick. 
Better, but impossible (i think): to highlight troops in legion when it hovered by cursor.

Thanks.

crudelios:
Also add changes to military behavior.

When the "show military sidebar" option is selected, the sidebar will display information about the currently selected legion.
Picking a destination for the legion to move to will not cancel its selection. The only exception is if you move the legion back to its own fort.

If you click a different legion, instead of the currently selected legion moving there, the clicked legion will be selected.

When the mouse hovers a legion, it will be highlighted to let you know it will be selected if clicked.

Detail change: The cursor is only a sword when it is over the city (and not over your own legions), otherwise it will be the regular arrow. Also, you can select the top menus during military mode now.

Besides the usual code comments, please test the interface and suggest changes. Since I changed some behaviours I'm not sure about everything I've done.

Fix #320.

Images.

Showing a single legion information while opening the military menu:

city 2020-03-23 14 44 27

Showing information about all legions while hovering over a legion (mouse doesn't appear):

city 2020-03-23 14 44 36


crudelios:
I'm opening this for review since there were no other comments.

crudelios:
This has finally been updated and rebased.

New changes:

* The buttons for cycling a legion are no longer the ones displayed on the screenshot, but the normal arrow buttons
* Using the `cycle legion` hotkey (defaults to `L`) when in military mode will properly select the cycled legion
* Because of the sidebar refactor (#444), the military sidebar now slides in and out instead of just appearing

src/widget/sidebar/military.c
```diff
@@ -0,0 +1,732 @@
...
+    {
+        FORMATION_SINGLE_LINE_2,
+        FORMATION_SINGLE_LINE_1,
+        FORMATION_DOUBLE_LINE_2,
+        FORMATION_DOUBLE_LINE_1,
+        FORMATION_MOP_UP
+    }
+};
+
+static void column_formation_button(int x, int y, int selected)
```
crudelios, line 143: All the `<xxx>_formation_button` functions are effectively drawing functions.

Since they're only used for the military sidebar I've kept them in this file.

However, should they go to `src/widget/formation_layout_buttons.c`?

bvschaik:
Without looking at the code yet, some observations:

1. The cursor switching between mouse pointer and sword is very confusing. If I have a row of javelins and I want to move them one tile to the left, the cursor is still a pointer, while clicking does move them
1. Sliding in the sidebar: I'm not a fan to be honest, it's too slow. I think replacing the build button parts immediately would work better. You can then also keep all sidebar info instead of only leaving the game speed.
1. `Go to legion` and `Return to fort` buttons: those do not fit in all languages. We can reuse the icon buttons from the military advisor for those.
1. Bug: in my saved game I had three legions. I clicked the first, could scroll to the next two using the arrows, but then it wrapped around to the first again and didn't update the sidebar. It seems the left arrow skips the 3rd (last) legion, the right arrow skips the 1st.
1. While I can appreciate pixel art, I think this would look nicer (the original buttons scaled by 50% using a simple linear algorithm):
legion-buttons

crudelios:
> 1. The cursor switching between mouse pointer and sword is very confusing. If I have a row of javelins and I want to move them one tile to the left, the cursor is still a pointer, while clicking does move them

I can get rid of that for the legions. Should I keep it for the menus, though?

> 2. Sliding in the sidebar: I'm not a fan to be honest, it's too slow. I think replacing the build button parts immediately would work better.

Sure, that can be changed. 

> You can then also keep all sidebar info instead of only leaving the game speed.

Well, I did it like this because the objectives and unemployment are irrelevant info when in military mode. I can add them back if you want.

> 3. `Go to legion` and `Return to fort` buttons: those do not fit in all languages. We can reuse the icon buttons from the military advisor for those.

The buttons from the military advisor are small and don't look so good, that's why I ended up using text.

I can try either scaling up the images or elipsizing the text (the full text would show as a tooltip), what do you think?

> 4. Bug: in my saved game I had three legions. I clicked the first, could scroll to the next two using the arrows, but then it wrapped around to the first again and didn't update the sidebar. It seems the left arrow skips the 3rd (last) legion, the right arrow skips the 1st.

I'll have a look at this one.

> 5. While I can appreciate pixel art, I think this would look nicer (the original buttons scaled by 50% using a simple linear algorithm):

Is the image a "programmer's mockup"? Because in that case, I'd like the code please :wink:

bvschaik:
Here you go :)
```
static void add_rgb(color_t color, int *r, int *g, int *b)
{
    *r += (color >> 16) & 0xff;
    *g += (color >> 8) & 0xff;
    *b += (color >> 0) & 0xff;
}

static color_t color_average(const image *img, const color_t *data, int x, int y)
{
    int r = 0, g = 0, b = 0;
    int num_colors = 1;
    add_rgb(data[y * img->width + x], &r, &g, &b);
    if (x + 1 < img->width) {
        add_rgb(data[y * img->width + x + 1], &r, &g, &b);
        num_colors++;
        if (y + 1 < img->height) {
            add_rgb(data[(y + 1) * img->width + x + 1], &r, &g, &b);
            num_colors++;
        }
    }
    if (y + 1 < img->height) {
        add_rgb(data[(y + 1) * img->width + x], &r, &g, &b);
        num_colors++;
    }
    return (r / num_colors) << 16 | (g / num_colors) << 8 | (b / num_colors) << 0;
}

void image_draw_scaled_50_percent(int image_id, int x_offset, int y_offset)
{
    const image *img = image_get(image_id);
    const color_t *data = image_data(image_id);
    if (!data || img->draw.type != 1) {
        return;
    }

    int width = img->width / 2;
    int height = img->height / 2;

    const clip_info *clip = graphics_get_clip_info(x_offset, y_offset, width, height);
    if (!clip->is_visible) {
        return;
    }
    for (int y = clip->clipped_pixels_top; y < height - clip->clipped_pixels_bottom; y++) {
        data += clip->clipped_pixels_left;
        color_t *dst = graphics_get_pixel(x_offset + clip->clipped_pixels_left, y_offset + y);
        int x_max = width - clip->clipped_pixels_right;

        for (int x = clip->clipped_pixels_left; x < x_max; x++, dst++) {
            *dst = color_average(img, data, x * 2, y * 2);
        }
    }
}
```

Then just use `image_draw_scaled_50_percent(image_group(GROUP_FORT_FORMATIONS) + button_id, x, y);` in `draw_layout_button`.

crudelios:
Thanks for the code!

What's image draw type `1`?

bvschaik:
> What's image draw type `1`?

The image type of those buttons... programmer's mockup :wink: I have to look it up, but I think 1 was "not compressed, no transparency", though you can also use the flags for compression, and transparency should not be too complex to support.

crudelios:
New look:

New military sidebar aspect

bvschaik:
I finally had the chance to test this properly during a real invasion in Mediolanum:

> If you click a different legion, instead of the currently selected legion moving there, the clicked legion will be selected.

This prevents the player from stacking/partially overlapping legions, which is a popular tactic to deal with heavy invasions. It also caused me to mis-click a few times and my legion was not moved out of the way in time at one point during hit-and-run tactics.

> Picking a destination for the legion to move to will not cancel its selection.

Not sure I like this one: it caused a LOT of right-clicking for me to cancel the military command mode. On a Mac, it's not easy to right-click on a touchpad... How does one 'cancel' military command for touch, by the way?

The "All" button in the "overlay" menu: I'm not sure what it's supposed to do/be useful for: sure, I can move legions, but it's not clear which one I'm moving: I don't remember my legion's names, I just remember "the javelins on the left side of the map". Also, for Augustus with more forts enabled, all 10 forts don't fit in the sidebar. I personally think we can drop this.

crudelios:
> This prevents the player from stacking/partially overlapping legions, which is a popular tactic to deal with heavy invasions. It also caused me to mis-click a few times and my legion was not moved out of the way in time at one point during hit-and-run tactics.

Sure, this can be reverted. The "highlight legions" should allow a user to see if he's about to click another legion, but I reckon that in the heat of battle the user can accidently mis-click.

> Not sure I like this one: it caused a LOT of right-clicking for me to cancel the military command mode. On a Mac, it's not easy to right-click on a touchpad...

This can also be reverted, I was trying to implement something like the Total War series which doesn't immediately cancel the selected legion upon clicking somewhere, allowing for finer control of the legion's position.

> How does one 'cancel' military command for touch, by the way?

To cancel military command for touch, you click the arrow at the upper right corner of the sidebar.

> The "All" button in the "overlay" menu: I'm not sure what it's supposed to do/be useful for: sure, I can move legions, but it's not clear which one I'm moving: I don't remember my legion's names, I just remember "the javelins on the left side of the map".

The idea for the "All" sidebar was to check how many units/morale you have left in a single panel during the heat of the battle. It's not really very practical for much else.

> Also, for Augustus with more forts enabled, all 10 forts don't fit in the sidebar. I personally think we can drop this.

Well, since Augustus has new features and does not worry to much about supporting the low-end 640x480 resolution, the "all" panel could be expanded to fit 10 units in there.

Anyway this can always be dropped if you think it's not worth it.

bvschaik:
This PR has been expanded upon in #516

PR #385, opened 2020-03-21
Work in progress: Drop custom cmake rules to find SDL2, rely on sdl2-config.cmake

### LINKED ISSUE #384: Drop custom cmake rules to find SDL2, rely on sdl2-config.cmake
Since version 2.0.6, SDL2 upstream provides `sdl2-config.cmake`.

Relying on this would allow to drop the custom rules to find SDL2 that are currently maintained as part of the build chain, at the cost of bumping minimum version of dependency on SDL2 to 2.0.6.

vv221:
⚠ This is currently a work in progress. ⚠

---

The objective of this proposed patch is to drop the custom cmake rules used to find and load SDL2, and use instead sdl2-config.cmake that is provided by the SDL2 upstream.
The inspiration is the following update of OpenJK: https://github.com/JACoders/OpenJK/pull/1033

---

Fix #384


vv221:
In the current state of this patch, the builds fail at the step `Linking C executable julius` with:
```
/usr/bin/ld: CMakeFiles/julius.dir/src/platform/arguments.c.o: undefined reference to symbol 'SDL_strtol'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/julius.dir/build.make:4960: julius] Error 1
make[1]: *** [CMakeFiles/Makefile2:98: CMakeFiles/julius.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

crudelios:
This is incompatible with Android build.

vv221:
@crudelios 

You listed good reasons to not switch to `sdl2-config.cmake` (yet?), so I see no issue with the closing of this discussion.

Do not worry about the work shown in this merge request: I spent only a couple minutes on this, mainly to illustrate the idea behind this patch. So I will not feel disappointed if it has been "for nothing" ;)

crudelios:
Thanks for the contribution anyway!

vv221:
I’m closing this merge request and [the related issue](https://github.com/bvschaik/julius/issues/384), see @crudelios [comment](https://github.com/bvschaik/julius/pull/385#issuecomment-602089142) for why it should not be implemented this way.

Issue #393, opened 2020-03-25
Message 'city need workers' when building a fort


LePatator:
As we discussed on discord, i fixed on a local branch this very little bug : 
If you build a fort when your city need workers, a message is displayed saying 'city need workers' which doesn't make sense since a fort doesn't consume workers.

![message](https://user-images.githubusercontent.com/62617403/77535261-9f9b5e80-6e9a-11ea-9b18-77657c8f4d23.png)

So the fix is only to not display this message in this particular case.

![noMessage](https://user-images.githubusercontent.com/62617403/77535336-bd68c380-6e9a-11ea-92eb-f005c03cf3bb.png)

bvschaik:
Hi @LePatator, nice going! To contribute your code, do the following:

1. **Fork** this project to your own Github account
2. Clone the fork to your local computer
3. Apply the code changes and push them to your Github
4. Create a **pull request**

LePatator:
I have a question : the 2 fix i just proposed with the 2 issues i opened could be added in Augustus as well. Is there an automatic process for this ? Or should do the same (fork, branch, pull request) with Augustus code ?

bvschaik:
Keriew normally pulls all Julius changes into Augustus before he creates a release, so nothing needs to be done for that :)

PR #395, opened 2020-03-25
Removed message saying city needs workers when building a fort

LePatator:
This branch corresponds to the fix detailled here : https://github.com/bvschaik/julius/issues/393


bvschaik:
Two general comments:
- Use the imperative mood instead of past tense for commit messages: use "Remove X" instead of "Removed X"
- You can automatically link (and close) issues by typing "Fixes #393" in the commit body

Issue #392, opened 2020-03-24
Display workers needed in fountains and missions


LePatator:
The fountains and the missions consume workers but it is not displayed in their description panel.

I fixed it, now the workers counter is displayed like in the other buildings.
![workers](https://user-images.githubusercontent.com/62617403/77463886-9e205680-6e06-11ea-8f30-8157b7c3ed9e.png)

My fix is on a local branch. If you want i can give it to you, but for now i can't push it to make a pull request (err 403).

crudelios:
I like this, but since fountains and mission posts don't need to spawn citizens, could you get rid of the message "poor access to workers"?

LePatator:
Ok, i removed this messages. I left the one displayed when there is nobody in the city because it still makes sense.
![workers](https://user-images.githubusercontent.com/62617403/77472488-4a1c6e80-6e14-11ea-9935-4d725142d49c.png)

crudelios:
I think you can keep the "too few people alocated by the labor advisor", as it also makes sense to keep :+1:

LePatator:
Ok I did it
![Sans titre](https://user-images.githubusercontent.com/62617403/77532223-b8087a80-6e94-11ea-9bc4-fe9122bfc17d.png)

bvschaik:
Fixed by #396

PR #396, opened 2020-03-25
Fixed workers needed not displayed in description panel for fountain …

LePatator:
…and mission buildings

This branch corresponds to the fix detailled here : https://github.com/bvschaik/julius/issues/392

### DIFF
--- a/src/window/building/common.c
+++ b/src/window/building/common.c
@@ -85,7 +85,7 @@ int draw_employment_info(building_info_context *c, building *b, int y_offset, in
     } else {
         text_id = 19; // too few people allocated
     }
-    if (!text_id && consider_house_covering == 1 && b->houses_covered < 40) {
+    if (!text_id && consider_house_covering && b->houses_covered < 40) {
         text_id = 20; // poor access to employees
     }
     return text_id;
--- a/src/window/building/common.c
+++ b/src/window/building/common.c
@@ -53,21 +53,7 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
     return new_window_y;
 }
 
-void window_building_draw_employment(building_info_context *c, int y_offset)
-{
-    building *b = building_get(c->building_id);
-    int text_id = draw_employment_info(c, b, y_offset, 1);
-    draw_employment_details(c, y_offset, text_id, b);
-}
-
-void window_building_draw_employment_without_house_cover(building_info_context *c, int y_offset)
-{
-    building *b = building_get(c->building_id);
-    int text_id = draw_employment_info(c, b, y_offset, 0);
-    draw_employment_details(c, y_offset, text_id, b);
-}
-
-int draw_employment_info(building_info_context *c, building *b, int y_offset, int consider_house_covering)
+static int draw_employment_info(building_info_context *c, building *b, int y_offset, int consider_house_covering)
 {
     int text_id;
     if (b->num_workers >= model_get_building(b->type)->laborers) {
@@ -91,7 +77,7 @@ int draw_employment_info(building_info_context *c, building *b, int y_offset, in
     return text_id;
 }
 
-void draw_employment_details(building_info_context *c, int y_offset, int text_id, building *b)
+static void draw_employment_details(building_info_context *c, int y_offset, int text_id, building *b)
 {
     y_offset += c->y_offset;
     image_draw(image_group(GROUP_CONTEXT_ICONS) + 14,
@@ -103,14 +89,27 @@ void draw_employment_details(building_info_context *c, int y_offset, int text_id
             c->x_offset + 70 + width, y_offset + 10, FONT_SMALL_BLACK);
         lang_text_draw(69, 0, c->x_offset + 70 + width, y_offset + 10, FONT_SMALL_BLACK);
         lang_text_draw(69, text_id, c->x_offset + 70, y_offset + 26, FONT_SMALL_BLACK);
-    }
-    else {
+    } else {
         int width = lang_text_draw_amount(8, 12, b->num_workers,
             c->x_offset + 60, y_offset + 16, FONT_SMALL_BLACK);
         width += text_draw_number(model_get_building(b->type)->laborers, '(', "",
             c->x_offset + 70 + width, y_offset + 16, FONT_SMALL_BLACK);
         lang_text_draw(69, 0, c->x_offset + 70 + width, y_offset + 16, FONT_SMALL_BLACK);
     }
+}
+
+void window_building_draw_employment(building_info_context *c, int y_offset)
+{
+    building *b = building_get(c->building_id);
+    int text_id = draw_employment_info(c, b, y_offset, 1);
+    draw_employment_details(c, y_offset, text_id, b);
+}
+
+void window_building_draw_employment_without_house_cover(building_info_context *c, int y_offset)
+{
+    building *b = building_get(c->building_id);
+    int text_id = draw_employment_info(c, b, y_offset, 0);
+    draw_employment_details(c, y_offset, text_id, b);
 }
 
 void window_building_draw_description(building_info_context *c, int text_group, int text_id)
--- a/src/window/building/common.c
+++ b/src/window/building/common.c
@@ -53,24 +53,34 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
     return new_window_y;
 }
 
-void window_building_draw_employment(building_info_context *c, int y_offset)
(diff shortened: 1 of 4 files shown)


src/window/building/common.c
```diff
@@ -53,24 +53,34 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
     return new_window_y;
 }
 
-void window_building_draw_employment(building_info_context *c, int y_offset)
+void window_building_draw_employment(building_info_context* c, int y_offset)
```
bvschaik: Please keep to the project's coding/spacing conventions

src/window/building/common.c
```diff
@@ -53,24 +53,34 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
...
+
+void do_window_building_draw_employment(building_info_context* c, int y_offset, int consider_house_covering)
+{
+    building* b = building_get(c->building_id);
     int text_id;
     if (b->num_workers >= model_get_building(b->type)->laborers) {
         text_id = 0;
-    } else if (city_population() <= 0) {
+    }
+    else if (city_population() <= 0) {
```
bvschaik: Please keep to the project's coding/spacing conventions: `}` and the corresponding `else` on one line.

src/window/building/common.c
```diff
@@ -53,24 +53,34 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
...
-void window_building_draw_employment(building_info_context *c, int y_offset)
+void window_building_draw_employment(building_info_context* c, int y_offset)
 {
-    building *b = building_get(c->building_id);
+    do_window_building_draw_employment(c, y_offset, 1);
+}
+
+void do_window_building_draw_employment(building_info_context* c, int y_offset, int consider_house_covering)
+{
+    building* b = building_get(c->building_id);
```
bvschaik: Please keep to the project's coding/spacing conventions: `*` belongs to the variable.

src/window/building/common.h
```diff
@@ -68,6 +68,8 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
 
 void window_building_draw_employment(building_info_context *c, int y_offset);
 
+void do_window_building_draw_employment(building_info_context* c, int y_offset, int consider_house_covering);
```
bvschaik: Keep to the projects code conventions by prefixing the functions with the file they're in.

But more importantly: in this case I would say it's better to create a new function called `void window_building_draw_employment_without_house_cover(building_info_context *c, int y_offset);`
Then, in the `.c` file, create a new function `void draw_employment_info(building_info_context* c, int y_offset, int consider_house_covering);`, which is called by both public functions.

This way, the `consider_house_covering` flag is not exposed to the outside: this function is currently only called from the outside with the flag set to `0`.

src/window/building/common.c
```diff
@@ -53,24 +53,34 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
...
     int text_id;
     if (b->num_workers >= model_get_building(b->type)->laborers) {
         text_id = 0;
-    } else if (city_population() <= 0) {
+    }
+    else if (city_population() <= 0) {
         text_id = 16; // no people in city
-    } else if (b->houses_covered <= 0) {
+    }
+    else if (consider_house_covering == 1 && b->houses_covered <= 0) {
```
bvschaik: The `consider_house_covering` variable is a boolean: just use `if (consider_house_covering)`, instead of checking for `== 1`.
Also, instead of checking for this variable every single `if` statement, you can also add the following code before this if-statement:
```
} else if (!consider_house_covering) {
   text_id = 19; // too few people allocated
} else ... the rest of the if statements
```

src/window/building/utility.c
```diff
@@ -165,7 +167,9 @@ void window_building_draw_mission_post(building_info_context *c)
     window_building_play_sound(c, "wavs/mission.wav");
     outer_panel_draw(c->x_offset, c->y_offset, c->width_blocks, c->height_blocks);
     lang_text_draw_centered(134, 0, c->x_offset, c->y_offset + 10, 16 * c->width_blocks, FONT_LARGE_BLACK);
-    window_building_draw_description_at(c, 106, 134, 1);
+    window_building_draw_description_at(c, 66, 134, 1);
```
bvschaik: Use this code to use the default offset for the description:
```suggestion
    window_building_draw_description(c, 134, 1);
```

src/window/building/common.c
LePatator: Ok done

src/window/building/common.c
LePatator: Done too. In fact theses modifications happen when i copy/paste. I don't use Visual studio often. I will check if there is a parameter to automatically change the coding style, or a file to it.

src/window/building/common.c
LePatator: same as above

src/window/building/utility.c
LePatator: Done.

src/window/building/common.c
LePatator: Ok

src/window/building/common.h
LePatator: Ok it needed a little more code refactoring because we want to avoid duplicated code. But yes i did what you say here.

src/window/building/common.c
```diff
@@ -56,11 +56,26 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
...
+}
+
+void window_building_draw_employment_without_house_cover(building_info_context *c, int y_offset)
+{
+    building *b = building_get(c->building_id);
+    int text_id = draw_employment_info(c, b, y_offset, 0);
+    draw_employment_details(c, y_offset, text_id, b);
+}
+
+int draw_employment_info(building_info_context *c, building *b, int y_offset, int consider_house_covering)
```
bvschaik: Make this function `static` and move it above the other two so you don't have to add a function prototype.

src/window/building/common.c
```diff
@@ -70,9 +85,14 @@ void window_building_draw_employment(building_info_context *c, int y_offset)
...
         text_id = 19; // too few people allocated
     }
-    if (!text_id && b->houses_covered < 40) {
+    if (!text_id && consider_house_covering && b->houses_covered < 40) {
         text_id = 20; // poor access to employees
     }
+    return text_id;
+}
+
+void draw_employment_details(building_info_context *c, int y_offset, int text_id, building *b)
```
bvschaik: Make this function `static` and move it up.

src/window/building/common.c
```diff
@@ -83,7 +103,8 @@ void window_building_draw_employment(building_info_context *c, int y_offset)
             c->x_offset + 70 + width, y_offset + 10, FONT_SMALL_BLACK);
         lang_text_draw(69, 0, c->x_offset + 70 + width, y_offset + 10, FONT_SMALL_BLACK);
         lang_text_draw(69, text_id, c->x_offset + 70, y_offset + 26, FONT_SMALL_BLACK);
-    } else {
+    }
+    else {
```
bvschaik: Formatting: } and else on the same line please.

src/window/building/common.h
```diff
@@ -68,6 +70,12 @@ int window_building_get_vertical_offset(building_info_context *c, int new_window
 
 void window_building_draw_employment(building_info_context *c, int y_offset);
 
+void window_building_draw_employment_without_house_cover(building_info_context *c, int y_offset);
+
+int draw_employment_info(building_info_context *c, building *b, int y_offset, int consider_house_covering);
+
+void draw_employment_details(building_info_context *c, int y_offset, int text_id, building *b);
```
bvschaik: These two functions don't belong in the header since they're only used in the corresponding c file itself. You can then also skip the header.

src/window/building/common.c
LePatator: done

src/window/building/common.c
LePatator: done

src/window/building/common.c
LePatator: done

src/window/building/common.h
LePatator: done

bvschaik:
Also, make sure the build succeeds, right now there are compile errors:
[Windows](https://ci.appveyor.com/project/bvschaik/julius/builds/31859760)
[Other platforms](https://travis-ci.org/github/bvschaik/julius/builds/669379022)

LePatator:
Ok it seems fine

Issue #383, opened 2020-03-21
Wishlist: Provide basic Linux binaries with releases


vv221:
Please consider providing "classic" Linux binaries with releases, in addition to the AppImage.
This would help a lot in including Julius in other projects.

---

I am currently planning to add support for Julius into [./play.it](https://forge.dotslashplay.it/play.it/scripts), a free/libre installer for DRM-free commercial games on Linux, that makes games installation easier by automating the build of native packages from a collection of games installers.

We currently support the Cæsar Ⅲ installers sold on gog.com, building packages that rely on WINE to run on Linux.

I first learnt about Julius a couple hours ago thanks to [a post on French news website LinuxFr.org](https://linuxfr.org/users/ted/journaux/et-vous-vous-jouez-a-quoi#comment-1802864), and tried it on a Debian Sid. The experience has been very good, so I would like to provide it as an alternative to WINE for people wanting to play Cæsar Ⅲ thanks to ./play.it.

With open source engines like [DevilutionX](https://github.com/diasurgical/devilutionX) for Diablo this inclusion is made quasi-trivial thanks to having access to binaries ready to include in our packages for any release. It would be very nice to have something similar with Julius.

vv221:
Sorry for the "bug" label, this is of course not a bug ;)

bvschaik:
The last time I checked (but that was years ago), there is no concept of "universal" Linux binaries: binaries are not compatible across CPU architecture, or even Linux distributions because of differences in libc versions. Please correct me if I'm wrong though.

Julius is already included in some distributions.

vv221:
> The last time I checked (but that was years ago), there is no concept of "universal" Linux binaries

This is still the case, most projects provide only x86_64 and x86_32 binaries (or only x86_64 more recently) as that seems to be enough to cover most use cases.

About the differences in libc versions, I actually don’t know how most project maintainers handle this. It might be as simple as building against a libc old enough that it would run on current LTS distributions.

> Julius is already included in some distributions.

I actually did not notice that, maybe including [a repology badge](https://repology.org/project/julius-caesar3/badges) in the readme might help make it more visible.

bvschaik:
[...]

PR #399, opened 2020-04-02
Prevent compilations linked on higher versions of SDL from crashing on lower versions

crudelios:
Should help with #383.


crudelios:
Closed due to #403.

PR #403, opened 2020-04-02
Add Linux binaries

### LINKED ISSUE #383: Wishlist: Provide basic Linux binaries with releases

bvschaik:
Fixes #383

### DIFF
--- a/src/platform/platform.c
+++ b/src/platform/platform.c
@@ -1,0 +1,10 @@
+#include "platform.h"
+
+#include "SDL.h"
+
+int platform_sdl_version_at_least(int major, int minor, int patch)
+{
+    SDL_version v;
+    SDL_GetVersion(&v);
+    return SDL_VERSIONNUM(v.major, v.minor, v.patch) >= SDL_VERSIONNUM(major, minor, patch);
+}
--- a/src/platform/sound_device.c
+++ b/src/platform/sound_device.c
@@ -4,6 +4,7 @@
 #include "game/settings.h"
 #include "SDL.h"
 #include "SDL_mixer.h"
+#include "platform/platform.h"
 #include "platform/vita/vita.h"
 
 #include <stdlib.h>
@@ -17,8 +18,9 @@
 #define MAX_CHANNELS 150
 
 #if SDL_VERSION_ATLEAST(2, 0, 7) 
-#define USE_SDL_AUDIOSTREAM 
+#define USE_SDL_AUDIOSTREAM
 #endif
+#define HAS_AUDIOSTREAM() (platform_sdl_version_at_least(2, 0, 7))
 
 typedef struct {
     const char *filename;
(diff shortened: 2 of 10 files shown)


src/platform/platform.c
```diff
@@ -0,0 +1,10 @@
+#include "platform.h"
+
+#include "SDL.h"
+
+int platform_version_at_least(int major, int minor, int patch)
```
crudelios, line 5: I know you want to hide SDL from the files, but I still think the name `platform_sdl_version_at_least` would be clearer.

src/platform/sound_device.c
```diff
@@ -17,8 +18,9 @@
 #define MAX_CHANNELS 150
 
 #if SDL_VERSION_ATLEAST(2, 0, 7) 
-#define USE_SDL_AUDIOSTREAM 
+#define USE_SDL_AUDIOSTREAM
 #endif
+#define HAS_AUDIOSTREAM() (platform_version_at_least(2, 0, 7))
```
crudelios, line 23: Do we really need this `#define`? We only use it once on line `65`...

devnoname120:
What is the point of having a Linux executable that is not statically linked? (And as such is extremely sensitive on having exactly the right library versions in the dynamic loader path.)

src/platform/sound_device.c
bvschaik: I created it to keep the 2.0.7 SDL versions together, otherwise there will be one check for 2.0.7 at the top, and one somewhere in the code.

bvschaik:
@devnoname120 see #383.

src/platform/sound_device.c
crudelios: Fair enough.

src/platform/platform.c
bvschaik: Fair point, will rename.

devnoname120:
> @devnoname120 see #383.

Ok, why can't they just use the universal executable that we already have?

PR #404, opened 2020-04-02
Fix touch bugs

### LINKED ISSUE #381: Bugs with improved touch support
While testing the new android build (check #380), I already found the following bugs:

* ~If you start building a road, wall or draggable reservoir in an invalid place (i.e. a tree or a rock), the build may "bug" and behave strangely. To get around the bug, cancel the build (click the upper-right `X` button until it disappears) and reselect it.~
* ~When you finish a mission and Caesar congratulates you, double tapping does not skip the window as I forgot to implement it there. To skip the window, tap once with two fingers for now (old behaviour).~ Fixed by #391.

**Note:** I will be striking the bugs I'm fixing in at home for my own reference, and at the end I'll provide a PR with all the fixes.

If you find any other bugs regarding touch support, please report them here and I'll fix them in bulk.

crudelios:
Fix some double clicks that I missed.

Fix a bug that when starting a road or wall construction on an invalid tile the construction would glitch.

Fix a bug when selecting military units that caused the screen to move.

Not totally touch related, but fixed a bug when scrolling on slow speeds that caused the scrolling to accelarate indefinitely in certain cases.

Fix #381.

### DIFF
--- a/src/input/scroll.c
+++ b/src/input/scroll.c
@@ -55,6 +55,7 @@ static struct {
         int y;
         time_millis start_time;
         time_millis last_time;
+        time_millis stopped_time;
         pixel_offset last_position;
         int decaying;
         float modifier;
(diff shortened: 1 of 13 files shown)


src/input/scroll.c
```diff
@@ -375,7 +368,7 @@ void scroll_get_delta(const mouse *m, pixel_offset *delta, scroll_type type)
     time_millis current_time = time_get_millis();
     time_millis time_delta = current_time - data.speed.last_time;
     data.speed.last_time = current_time;
-    if (time_delta > 17 && time_delta < MAX_SPEED_TIME_WEIGHT) {
+    if (time_delta > (int) FRAME_TIME && time_delta < MAX_SPEED_TIME_WEIGHT) {
```
bvschaik, line 371: I think I missed it last time, but I don't think it's a good idea to rely on a 17ms frame time. For 120Hz monitors this will be 8.3ms, and it will be undefined for systems which don't support vsync.

crudelios: So I have changed the code (the branch has 3 commits with this fixed) but due to the outage, I guess it didn't sync the PR.

Anyway, I removed that `(int) FRAME_TIME`. It seems to be working properly, and now scroll speed is normalized even if the monitor's refresh rate is higher than 60hz.

PR #405, opened 2020-04-02
Zoom preparation: add support for middle mouse button click

crudelios:
Zooming will restore the city scale to default upon clicking on the middle mouse button.

So I'm providing middle mouse button support in a separate PR, so as not to add to much code at the same time.


bvschaik:
What are we going to use this for? Touchpads have no middle mouse button, Mac doesn't have it at all, so this will only be useful for a subset of users.

crudelios:
Most mice's scroll wheels can be clicked, which triggers the middle mouse button. My idea for zoom is to use the scroll wheel to zoom in or out, and clicking the scroll wheel restores it to 100%. It's convenient, but it would be its only use.

anon569:
This is pretty standard in other games but there should be a hotkey for it as well.

PR #408, opened 2020-04-05
Refactor hotkey handling

bvschaik:
Preparation for possibly configurable hotkeys.

### DIFF
--- a/src/game/cheats.c
+++ b/src/game/cheats.c
@@ -26,14 +26,14 @@ void game_cheat_activate(void)
 void game_cheat_money(void)
 {
     if (data.is_cheating) {
-        city_victory_force_win();
+        city_finance_process_cheat();
+        window_invalidate();
     }
 }
 
 void game_cheat_victory(void)
 {
     if (data.is_cheating) {
-        city_finance_process_cheat();
-        window_invalidate();
+        city_victory_force_win();
     }
 }
--- a/src/game/cheats.c
+++ b/src/game/cheats.c
@@ -1,0 +1,39 @@
+#include "cheats.h"
+
+#include "building/type.h"
+#include "city/finance.h"
+#include "city/victory.h"
+#include "graphics/window.h"
+#include "scenario/invasion.h"
+#include "window/building_info.h"
+
+static struct {
+    int is_cheating;
+} data;
+
+void game_cheat_activate(void)
+{
+    if (window_is(WINDOW_BUILDING_INFO)) {
+        data.is_cheating = window_building_info_get_building_type() == BUILDING_WELL;
+    } else if (data.is_cheating && window_is(WINDOW_MESSAGE_DIALOG)) {
+        data.is_cheating = 2;
+        scenario_invasion_start_from_cheat();
+    } else {
+        data.is_cheating = 0;
+    }
+}
+
+void game_cheat_money(void)
+{
+    if (data.is_cheating) {
+        city_victory_force_win();
+    }
+}
+
+void game_cheat_victory(void)
+{
+    if (data.is_cheating) {
+        city_finance_process_cheat();
+        window_invalidate();
+    }
+}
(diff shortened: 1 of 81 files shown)


crudelios:
I'll have a look at this. I'm just about to implement a `joystick_mapping` class, so I'll have a look at this codebase to see what ideas you came up with to keep the implementation similar.

src/game/cheats.c
```diff
@@ -0,0 +1,39 @@
...
+        scenario_invasion_start_from_cheat();
+    } else {
+        data.is_cheating = 0;
+    }
+}
+
+void game_cheat_money(void)
+{
+    if (data.is_cheating) {
+        city_victory_force_win();
```
crudelios, line 29: You mixed up the money and victory cheat! Almost as good as Caesar firing you in the map editor haha :wink:

src/window/city.c
```diff
@@ -99,22 +108,127 @@ static void draw_foreground_military(void)
...
+    window_invalidate();
+}
+
+static void cycle_legion(void)
+{
+    static int current_legion_id = 1;
+    if (window_is(WINDOW_CITY)) {
+        int legion_id = current_legion_id;
+        current_legion_id = 0;
+        for (int i = 1; i <= MAX_LEGIONS; i++) {
```
crudelios, line 136: I know this bug isn't directly caused by your changes, but I think it should be `MAX_FORMATIONS` here, not `MAX_LEGIONS`.

src/window/plain_message_dialog.c
```diff
@@ -47,24 +47,27 @@ static void draw_foreground(void)
...
+}
+
+static void handle_input(const mouse *m, const hotkeys *h)
 {
     if (image_buttons_handle_mouse(mouse_in_dialog(m), 80, 80, buttons, 1, 0)) {
         return;
     }
-    if (input_go_back_requested()) {
-        window_plain_message_dialog_accept();
+    if (input_go_back_requested(m, h)) {
```
crudelios, line 60: Why not just:

```
if (input_go_back_requested(m, h) || h->enter_pressed) {
    window_go_back();
}
```

src/game/cheats.c
bvschaik: I had too much sun yesterday 😂

src/window/city.c
bvschaik: True. I'll fix that directly on master though.

src/window/plain_message_dialog.c
bvschaik: Probably because I was quite tired after refactoring about 70 screens. I'll simplify this.

bvschaik: Probably because I was quite tired after refactoring about 70 screens. I'll simplify this.

bvschaik:
> I'll have a look at this. I'm just about to implement a `joystick_mapping` struct, so I'll have a look at this codebase to see what ideas you came up with to keep the implementation similar.

The basic idea is: `hotkeys.c` should translate key presses into "commands" which are in the `hotkeys` struct. The individual screens will process those commands in their `handle_input` callback (previously called `handle_mouse`). This way, the mapping from key to command can be changed easily later.

I'm not sure how the mapping will look like yet though, as I'm still unsure how to proceed with regards to different keyboard layouts. Should the mapping be for physical keys or virtual ones? And how should the user configure the hotkeys? Just by editing an .ini file or in-game? If in-game, how?

crudelios:
> Should the mapping be for physical keys or virtual ones?

After thinking about this, I think mapping should be for physical keys. It's what makes sense. Users should expect the same key to work regardless of the layout.

> And how should the user configure the hotkeys? Just by editing an .ini file or in-game? If in-game, how?

We could start with a simple ini file then implement a configuration window later. I'll be trying to create a configuration window for joystick so I can adapt it for regular hotkeys.

PR #409, opened 2020-04-08
Dynamic hotkeys

### LINKED ISSUE #379: Speed keys do not repeat
In the original Caesar III, if I press and hold the "[" key, the game speed slows down while the "[" character repeats. In Julius, it seems that the repeat does not occur, I have to press the key repeatedly to get the same effect.

This is not a big deal, just a slight difference in the gameplay.

bvschaik:
Make hotkey handling dynamic, so we can make them configurable in the next iteration.

Note that this does remove support for keyboard layouts other than US for the moment.

Also fixes #379


crudelios:
LGTM :plus1:

PR #411, opened 2020-04-12
Refactor scrollbar to reusable component

bvschaik:


crudelios:
Overall very useful addition (I think @Keriew will like this one for Augustus due to the extra forts requiring scrolling).

Just a small comment: Right now `scrollbar.c` has too many magic numbers for width and height calculation. The numbers `77` and `26`, among some others, are all over the place. Can you add them to a `#define`?

Other than that, LGTM.

bvschaik:
Good point, changed :)

PR #418, opened 2020-04-17
Compile switch and vita builds without using docker

crudelios:
This allows more control of the build. Also, it removes some old warnings.

I'll probably do the same for vita as well.


.ci_scripts/start_docker.sh
```diff
@@ -8,6 +8,8 @@ case "$BUILD_TARGET" in
 	docker run -d --name vitasdk --workdir /build/git -v "${PWD}:/build/git" gnuton/vitasdk-docker:20190626 tail -f /dev/null
 	;;
 "switch")
-	docker run -d --name switchdev --workdir /build/git -v "${PWD}:/build/git" rsn8887/switchdev tail -f /dev/null
+	wget https://github.com/devkitPro/pacman/releases/download/devkitpro-pacman-1.0.1/devkitpro-pacman.deb
```
bvschaik, line 11: A bit weird to install packages in a script file named `start_docker.sh`... Rename this file to `setup_environment.sh`?
Also: do we want to cache this file?

crudelios: Yes, I'm going to change the name when I get vita off docker as well :)

I'm not sure it is worth it to cache the file, as instead of it being fetched from github, it'll instead be fetched from the travis servers...

crudelios:
Closed for further testing and vita docker removal

crudelios:
Done. Reopening.

CMakeLists.txt
```diff
@@ -770,8 +765,9 @@ elseif(SWITCH_BUILD)
         POST_BUILD
         COMMAND ${DEVKITPRO}/tools/bin/nacptool --create "${SHORT_NAME}" "rsn8887" ${SWITCH_VERSION} ${SHORT_NAME}.nacp
         COMMAND ${DEVKITPRO}/tools/bin/elf2nro ${SHORT_NAME} ${SHORT_NAME}.nro --icon=${CMAKE_SOURCE_DIR}/res/switchdata/icon0.jpg --nacp=${SHORT_NAME}.nacp
-        COMMAND rm -rf release/${SHORT_NAME}
-        COMMAND mkdir -p release/${SHORT_NAME}
+        COMMAND rm -rf release
+        COMMAND mkdir release
+        COMMAND cd release && mkdir ${SHORT_NAME} && cd ..
```
bvschaik, line 770: Why do we do this in two steps? `mkdir -p release/${SHORT_NAME}` will also create the parent `release` if it doesn't exist.

crudelios: Because doing it in one step doesn't work in Visual Studio as Windows doesn't support the `-p` parameter for `mkdir`.

This was the only way that I found which worked on both Visual Studio and Linux.

devnoname120:
Note that if I used Docker for the vitasdk it's because people kept merging breaking changes. It's probably the same for Switch need to ask @rsn8887.

Is there anything that the Docker containers prevent you from doing?

crudelios:
Well, it prevents us from getting a more recent version but I get your point.

devnoname120:
OK for info the Vita ones are built every day so updating is as easy as changing the tag: https://hub.docker.com/r/gnuton/vitasdk-docker/tags?page=1&ordering=name

Looks like the Switch one isn't updated though.

PR #419, opened 2020-04-18
Add hotkey configuration screen

bvschaik:
Check it out from the main menu -> Options -> Configure hotkeys:
hotkey-config


crudelios:
This looks really good! I'll have a look :+1

jdmillard:
I was seriously about to ask about such a feature... Then I checked my email.

Mind blown.

bvschaik:
Also please test different keyboard layouts. The letter-based hotkeys should map to the current layout's actual letters by default. But once you have set up anything they should always map to the same physical key.

crudelios:
I've been checking this and I found two issues so far:

First, there's a problem with the displayed hotkeys.

In the portuguese layout, the default hotkeys are being translated fine to my layout. However, the `Increase game speed` key is `´` and the `Chief advisor` key is `'`. These do not show up at all in the hotkey editor:

city 2020-04-20 09 20 20

A possible way to fix this would be to say the key's name (i.e. like `Return`, `Tab`, etc).

For portuguese layout, here are the keys that are not recognized:

`'`, `«`, `´`, `º`, `~`.

Also, the `<` key (which is immediately to the right of the `Shift` key) seems to remove the modifier.

The other issue is, there seems to be no prevention against setting the same hotkey for two incompatible actions. For example I can set `F7` for both setting `640x480` and `800x600`resolution.

bvschaik:
> First, there's a problem with the displayed hotkeys.
> (...)
> A possible way to fix this would be to say the key's name (i.e. like `Return`, `Tab`, etc).

Yes. I'm going to add names for all the ASCII characters which do not have a displayable character. For any other keys that produce characters that cannot be displayed (e.g. Cyrillic in a non-Russian game version), I'm falling back to `? (<key name in US layout>)`. It's not perfect but I think the best we can do.

> Also, the `<` key (which is immediately to the right of the `Shift` key) seems to remove the modifier.

Any "unknown" key on the keyboard is going to do that: I didn't include a mapping for that key. I'll add that one. The problem is naming the thing... on my physical keyboard it's an addtional `\` or `|`, on Mac that's where the `` ` `` and `~` are located.

> Another issue is, the numpad keys don't get properly assigned. `/`, `*` and `+` work like `<` in the way that they simply remove the modifier. `-` shows up as `'` (which, strangely enough, as I said didn't show up at all when I pressed the actual `'` key).

Ah yes. I mapped the `-`, `Enter` and numeric keys of the keypad to the same keys in the alphanumeric layout, and didn't include the `/` and `*` and `+` keys at all. I'll add separate keys for them.
Would it make sense to treat the 0-9 of the numeric keypad differently from the regular 0-9 on the main part of the keyboard?

> The other issue is, there seems to be no prevention against setting the same hotkey for two incompatible actions. For example I can set `F7` for both setting `640x480` and `800x600`resolution.

That was sort of intended (and because I was lazy 😛): you can have "1" set to "open labor advisor", but also have the "1" set to "show army info in the editor" without them conflicting each other. I thought that leaving it up to the user to solve conflicts would be easiest, since I would rather not keep a table of which hotkeys are compatible.

crudelios:
> Would it make sense to treat the 0-9 of the numeric keypad differently from the regular 0-9 on the main part of the keyboard?

I think they should be separate, yeah!

> That was sort of intended (and because I was lazy 😛): you can have "1" set to "open labor advisor", but also have the "1" set to "show army info in the editor" without them conflicting each other. I thought that leaving it up to the user to solve conflicts would be easiest, since I would rather not keep a table of which hotkeys are compatible.

Fair enough. How about adding just a warning that the keys may conflit depending on what they were set to? (and it's up to the user to see if they don't).

bvschaik:
> > Would it make sense to treat the 0-9 of the numeric keypad differently from the regular 0-9 on the main part of the keyboard?
> 
> I think they should be separate, yeah!

Done. Also added key names and improved handling of unknown keys.

> How about adding just a warning that the keys may conflit depending on what they were set to? (and it's up to the user to see if they don't).

Sounds good. I'll look into that.

crudelios:
It's looking better, but now the key name doesn't fit:

city 2020-04-20 18 50 38

bvschaik:
I've given the key name a little more space and have cut off any remaining text: if you go crazy with the modifiers it will never fit anyway...

PR #420, opened 2020-04-20
Fix all warnings

crudelios:
All compiler warnings removed again (hopefully no new ones appear due to this PR 😉)

### DIFF
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -54,7 +54,7 @@ void config_set_string(config_string_key key, const char *value)
     if (!value) {
         string_values[key][0] = 0;
     } else {
-        snprintf(string_values[key], CONFIG_STRING_VALUE_MAX, "%s", value);
+        strncpy(string_values[key], value, CONFIG_STRING_VALUE_MAX - 1);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/core/config.c
```diff
@@ -73,7 +73,7 @@ void config_set_defaults(void)
     for (int i = 0; i < CONFIG_MAX_ENTRIES; ++i) {
         values[i] = default_values[i];
     }
-    strncpy(string_values[CONFIG_STRING_UI_LANGUAGE_DIR], default_string_values[CONFIG_STRING_UI_LANGUAGE_DIR], CONFIG_STRING_VALUE_MAX - 1);
+    strncpy(string_values[CONFIG_STRING_UI_LANGUAGE_DIR], default_string_values[CONFIG_STRING_UI_LANGUAGE_DIR], CONFIG_STRING_VALUE_MAX);
```
bvschaik, line 76: What was the warning here? The `- 1` is on purpose: if the string is longer than `CONFIG_STRING_VALUE_MAX`, the string is copied without trailing `\0`. Using `-1` leaves the last position unwritten so its default is `\0`, and the string will always be terminated.

crudelios: This is the warning:

```
warning: 'strncpy' output may be truncated copying 63 bytes from a string of length 63 [-Wstringop-truncation]
   76 |     strncpy(string_values[CONFIG_STRING_UI_LANGUAGE_DIR], default_string_values[CONFIG_STRING_UI_LANGUAGE_DIR], CONFIG_STRING_VALUE_MAX - 1);
```

One alternative is to use `snprintf`. It's slower but it copies the ending `0` when the size doesn't fit.

bvschaik: Ok. In this case (known src and dst), we can indeed use `strncpy` with `CONFIG_STRING_VALUE_MAX`.

crudelios: I went ahead and replaced with `snprintf` anyway. It's safer as it adds the `0` and the code is not performance critical.

bvschaik: I don't like that on a conceptual level: you don't use snprintf for copying a string.

crudelios: All right, I'll revert it 👍 

But although copying strings isn't the real purpose of `snprintf`, the truth is it works, it's safer than `strncpy` and it's actually used for that effect.

bvschaik: On second thought, `memcpy` might have been even more appropriate here, but let's stick with strncpy now ;)

PR #422, opened 2020-04-22
Modularizing y offset values in extra info panel 

bbright024:
image

Above: Sidebar in master at time of PR


src/widget/sidebar.c
```diff
@@ -332,27 +333,49 @@ static void draw_extra_info_objective(int x_offset, int y_offset, int text_group
...
     }
+
+    // Objective value info on extra_info panel (culture, prosperity, peace, etc)
     if (extra_info_height >= EXTRA_INFO_HEIGHT_RATINGS) {
-        draw_extra_info_objective(x_offset, y_offset + 110, 53, 1, &data.extra_info.culture, 0);
-        draw_extra_info_objective(x_offset, y_offset + 142, 53, 2, &data.extra_info.prosperity, 0);
-        draw_extra_info_objective(x_offset, y_offset + 174, 53, 3, &data.extra_info.peace, 0);
-        draw_extra_info_objective(x_offset, y_offset + 206, 53, 4, &data.extra_info.favor, 0);
-        draw_extra_info_objective(x_offset, y_offset + 238, 4, 6, &data.extra_info.population, 1);
+        draw_extra_info_objective(x_offset, y_current_line, 53, 1, &data.extra_info.culture, 0);
```
crudelios, line 365: Perhaps you can make `draw_extra_info_objective` return `y_line_space`?

This way all the following lines could be replaced with `y_current_line += draw_extra_info_objective(...);`.

src/widget/sidebar.c
```diff
@@ -332,27 +333,49 @@ static void draw_extra_info_objective(int x_offset, int y_offset, int text_group
 static void draw_extra_info_panel(int x_offset, int extra_info_height)
 {
     int y_offset = FILLER_Y_OFFSET;
-
+    int y_current_line = y_offset;
+    // Space between lines, when line is purely text (no buttons)
+    int y_line_space = EXTRA_INFO_HEIGHT_LINE_SPACE;
```
crudelios, line 338: Assigning a constant value to `y_line_space` seems needless TBH. Just use `EXTRA_INFO_HEIGHT_LINE_SPACE` instead.

bbright024: the name is so long :smile: that's why i didn't; i'll shorten the name of the defined value i suppose.
@crudelios is there a style checker in the CI pipeline? in prior projects i've always had to deal with line length

src/widget/sidebar.c
bbright024: yeah, been thinking about how best to do that.  another idea i had is to have `draw_extra_info_objective` modify `y_line_space` itself before returning.

so, 3 options, and the one chosen will influence how future side panel items are added:
1) return value from callee, applied to line marker in caller 
2) side effect in the callee, caller doesnt care
3) caller controls all (current model) 

option 2 would make future work easier, but with a higher risk of bugs.  that's why i've kept option 3 so far; i don't want to significantly alter the style of what you guys have done so far without a discussion

src/widget/sidebar.c
crudelios: Well, the names do tend to be verbose, but I always felt it makes the code clearer.

>  is there a style checker in the CI pipeline?

Not in the CI pipeline, but we have a very strict style checker called @bvschaik :wink: But adding an automated style checker would be a good idea... Let's see if I can find one for C!

src/widget/sidebar.c
crudelios: You're right about choice `3` being the clearest. I don't really like option `2`, functions shouldn't have unknown side effects. Obviously option `1` isn't as clear either, but at least the return value can be ignored.

src/widget/sidebar.c
bbright024: > Let's see if I can find one for C!
what have i done... :laughing: 

another reason i did this is because of line 335 -  
```
int y_offset = FILLER_Y_OFFSET;
```
y_offset doesnt change throughout draw_extra_info_panel.   ill remove y_offset, if no-one has any objections

src/widget/sidebar.c
bbright024: well if i did a bunch of work to implement option 1, and the programmer chose to ignore the return value, then the programmer would still have to do option 3, or else the line wouldn't change.

in addition, not implementing (1) lets us keep future options open for a return value - e.g. error codes or something else.

gonna resolve this and stick with the current implementation; it's ugly code, but it's safe code.

src/widget/sidebar.c
```diff
@@ -329,30 +334,53 @@ static void draw_extra_info_objective(int x_offset, int y_offset, int text_group
     text_draw_number(obj->target, '(', ")", x_offset + 11 + width, y_offset + 16, font);
 }
 
+
```
bvschaik, line 337: Two lines between functions is one too many 😉

src/widget/sidebar.c
```diff
@@ -329,30 +334,53 @@ static void draw_extra_info_objective(int x_offset, int y_offset, int text_group
...
     graphics_draw_vertical_line(x_offset, y_offset, y_offset + extra_info_height, COLOR_WHITE);
     graphics_draw_vertical_line(x_offset + SIDEBAR_EXPANDED_WIDTH - 1, y_offset, y_offset + extra_info_height, COLOR_SIDEBAR);
     inner_panel_draw(x_offset + 1, y_offset, SIDEBAR_EXPANDED_WIDTH / 16, panel_blocks);
 
-    lang_text_draw(45, 2, x_offset + 11, y_offset + 10, FONT_NORMAL_WHITE);
-    text_draw_percentage(data.extra_info.game_speed, x_offset + 60, y_offset + 36, FONT_NORMAL_GREEN);
-    arrow_buttons_draw(x_offset, y_offset, arrow_buttons_speed, 2);
+    int y_current_line = y_offset + EXTRA_INFO_TOP_PADDING;
+
+    // Game speed modifier in extra_info panel
```
bvschaik, line 350: I don't think the comments really add something to the code.

src/widget/sidebar.c
bvschaik: I certainly agree on adding an automatic style checker. If I can be automated, please do 😉 

> what have i done... 😆

Just an extra new line where one shouldn't be 😉 
I'm quite strict on brace positions, spacing and other formatting things; @crudelios knows from my comments on previous pull requests 😆 

> y_offset doesnt change throughout draw_extra_info_panel. ill remove y_offset, if no-one has any objections

No objection here.

src/widget/sidebar.c
bbright024: yeah, i put those comments in because when i was looking through this function last night, it took a bit of trial and error to determine which calls were impacting what parts of the panel.

also, while they're minor comments, they should help in the future with organization.  for example, if the game speed modifier part was moved into a separate function or wrapped in a config check, the comments i put in help organize which lines are required to make sure it all works properly.

bbright024:
note to self; fix issue with arrow button sounds on click

src/widget/extra_info.h
```diff
@@ -0,0 +1,28 @@
+#ifndef WIDGET_EXTRA_INFO_H
```
bvschaik, line 1: I'd rather call this `sidebar_filler`, and use it for drawing both the extra info and the relief images below it.

The interface can then become much simpler:
- one function for drawing the collapsed filler, arguments: x, y, width, height
- one function for drawing the expanded filler, arguments: x, y, width, height
- one function for handling mouse (the button clicks), arguments: mouse, x, y, width, height

src/widget/extra_info.h
```diff
@@ -0,0 +1,28 @@
...
+    int height;
+    int game_speed;
+    int unemployment_percentage;
+    int unemployment_amount;
+    objective culture;
+    objective prosperity;
+    objective peace;
+    objective favor;
+    objective population;
+} extra_info;
```
bvschaik, line 21: If structs are not used anywhere else besides the corresponding `.c` file, put them in the `.c` file to not pollute the global namespace with them.

src/widget/extra_info.h
```diff
@@ -0,0 +1,28 @@
...
+
+void draw_extra_info_panel(int x_offset, int extra_info_height);
+void draw_extra_info_buttons(int x_offset, int is_collapsed);
+int calculate_extra_info_height(int is_collapsed);
+int update_extra_info(int height, int is_background);
+int extra_info_height_game_speed_check(int height);
```
bvschaik, line 27: Stick to the naming convention of FILENAME_function_name. After applying my naming suggestion, they would have to be prefixed with `sidebar_filler_`.

src/widget/sidebar.c
```diff
@@ -1,4 +1,5 @@
 #include "sidebar.h"
+#include "extra_info.h"
```
bvschaik, line 2: For any includes which are not the corresponding header: please put them in alphabetical order in the list below, including directory prefix. Reasoning: if you move this file, all includes will still work OK.

src/widget/sidebar.c
```diff
@@ -40,16 +41,9 @@
...
-#define SIDEBAR_EXPANDED_WIDTH 162
-#define FILLER_Y_OFFSET 474
-
-#define EXTRA_INFO_HEIGHT_GAME_SPEED 64
-#define EXTRA_INFO_HEIGHT_UNEMPLOYMENT 112
-#define EXTRA_INFO_HEIGHT_RATINGS 272
-
 #define SIDEBAR_SLIDE_STEPS 94
 
+
```
bvschaik, line 46: No double empty lines please.

src/widget/sidebar.h
```diff
@@ -13,4 +14,27 @@ int widget_sidebar_handle_mouse_build_menu(const mouse *m);
...
+    int progress;
+    int focus_button_for_tooltip;
+    extra_info extra_info_vals; // see widget/extra_info.h for typedef
+} sidebar_data;
+
+// this global replaces the old global 'data' confined to sidebar.c;
+// temporarily placing in sidebar.h while breaking out extra_info into extra_info.c
+// future work will have extra_info.c functions that use sidebar_data_vals use struct pointers,
+// rather than the global.
+sidebar_data sidebar_data_vals;
```
bvschaik, line 38: Please, no globals. I just spent two years getting rid of all global variables :wink:

src/widget/extra_info.h
bbright024: ok, sounds good - ill make some open issues for these functions, as i want this pr to do as little logic changes as possible.  ill tackle em after the new files are safely in master

bvschaik: Actually, please do it in this PR, since I'm not going to merge this one with the global variables and the non-conforming naming convention still in here.
And please don't create issues for code cleanup, just submit PRs. I'd like to keep the issues for, well, issues related to gameplay etc.

src/widget/extra_info.h
bbright024: totally agree with ya, unfortunately this extra_info struct is a field in the data struct in sidebar.c 

https://github.com/bvschaik/julius/blob/master/src/widget/sidebar.c#L142

```
static struct {
    time_millis slide_start;
    int progress;
    int focus_button_for_tooltip;
    struct {
        int height;
        int game_speed;
        int unemployment_percentage;
        int unemployment_amount;
        objective culture;
        objective prosperity;
        objective peace;
        objective favor;
        objective population;
    } extra_info;
} data;
```

however it looks like the old sidebar.c doesn't use extra_info anymore after i split it all up, so ill probably be able to remove the extra_info field from the old data struct.

src/widget/extra_info.h
bbright024: :+1: will do

src/widget/extra_info.h
bbright024: ok - renamed and opened a new issue https://github.com/bvschaik/julius/issues/425

bbright024: ok, will do, ill close out the issue i made

src/widget/sidebar.c
```diff
@@ -76,7 +79,6 @@ static void button_empire(int param1, int param2);
 static void button_mission_briefing(int param1, int param2);
 static void button_rotate_north(int param1, int param2);
 static void button_rotate(int clockwise, int param2);
-static void button_game_speed(int is_down, int param2);
```
bbright024, line 79: note - `button_game_speed(int is_down, int param2)` is now in sidebar_filler.c - i know it looks odd to split it out from the other `button_` sidebar functions, but to me doing this makes sense when organization is looked at from the perspective of "original game" vs "julius additions".

thoughts?

bbright024:
ok - pretty sure i got everything requested/desired in splitting up sidebar.c and sidebar_filler.c

this was a really good refresher in C - been a few years.

@bvschaik @crudelios any other thoughts?  i think i got all the style stuff sorted out - `astyle` is a neat tool, it fixed some spacing stuff i missed.  

http://astyle.sourceforge.net/astyle.html

src/widget/sidebar.c
bvschaik: It doesn't look odd at all. Buttons, click handlers, etc, should go in the code that uses it. In the end, `sidebar.c` shouldn't know anything about buttons for game speed, and `sidebar_filler.c` should be an entirely self-sufficient piece of code. Just like the minimap, which handles its own clicks.

src/widget/sidebar.c
```diff
@@ -514,9 +324,14 @@ int widget_sidebar_handle_mouse_build_menu(const mouse *m)
     }
 }
 
+int widget_sidebar_get_expanded_width(void)
+{
+    return SIDEBAR_EXPANDED_WIDTH;
+}
+
```
bvschaik, line 331: We should pass this data as parameter instead of `sidebar` calling `sidebar_filler`, and `sidebar_filler` having to call back to `sidebar` to get the width: such circular dependencies should be avoided.

src/widget/sidebar.h
```diff
@@ -2,6 +2,7 @@
 #define WIDGET_SIDEBAR_H
 
 #include "input/mouse.h"
+#include "widget/sidebar_filler.h"
```
bvschaik, line 5: Why do we need this include?

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,289 @@
...
+#include "graphics/screen.h"
+#include "graphics/text.h"
+#include "graphics/window.h"
+#include "map/orientation.h"
+#include "scenario/property.h"
+#include "scenario/criteria.h"
+#include "sound/effect.h"
+#include "widget/city.h"
+#include "widget/minimap.h"
+#include "widget/sidebar.h"
```
bvschaik, line 34: Sidebar_filler should ideally not have a dependency back on `sidebar`

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,289 @@
...
+    int unemployment_percentage;
+    int unemployment_amount;
+    objective culture;
+    objective prosperity;
+    objective peace;
+    objective favor;
+    objective population;
+} extra_info;
+
+extra_info extra_info_vals;
```
bvschaik, line 70: Make it static to prevent polluting the global scope and call it `data` (project convention).
For such file-specific data structs: don't define a typedef for the struct, but do the definition in one go:
```
static struct {
   ...
} data;
```
Also, put it below the function prototypes and arrow buttons.

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,289 @@
...
+
+static void set_extra_info_objectives(void);
+static int update_extra_info_value(int value, int *field);
+static void draw_extra_info_objective(int x_offset, int y_offset, int text_group, int text_id, objective *obj, int cut_off_at_parenthesis);
```
bvschaik, line 74: Try to rearrange the functions in such a way that function prototypes are not needed, since they cause code duplication: when you change the function you also need to change the declaration.

src/widget/sidebar_filler.h
```diff
@@ -0,0 +1,19 @@
...
+void sidebar_filler_draw_extra_info_panel(int x_offset, int extra_info_height);
+void sidebar_filler_draw_extra_info_buttons(int x_offset, int is_collapsed);
+void sidebar_filler_draw_sidebar_filler(int x_offset, int y_offset, int is_collapsed);
+
+int sidebar_filler_calculate_extra_info_height(int is_collapsed);
+int sidebar_filler_update_extra_info(int height, int is_background);
+int sidebar_filler_extra_info_height_game_speed_check(void);
+
+// Also used in widget/sidebar.c 
+#define FILLER_Y_OFFSET 474
```
bvschaik, line 17: Do not put this in the header, but pass it to the `sidebar_filler` function as `y_offset` parameter.

src/widget/sidebar_filler.h
```diff
@@ -0,0 +1,19 @@
+#ifndef WIDGET_SIDEBAR_FILLER_H
+#define WIDGET_SIDEBAR_FILLER_H
+
+#include "graphics/arrow_button.h"
+
+arrow_button *sidebar_filler_get_arrow_buttons_speed(void);
```
bvschaik, line 6: These shouldn't be exposed: code that uses the filler should not know that arrow buttons are used internally.

src/widget/sidebar.c
```diff
@@ -40,15 +41,17 @@
...
-#define EXTRA_INFO_HEIGHT_UNEMPLOYMENT 112
-#define EXTRA_INFO_HEIGHT_RATINGS 272
+typedef struct {
+    time_millis slide_start;
+    int progress;
+    int focus_button_for_tooltip;
+} sidebar_data;
 
-#define SIDEBAR_SLIDE_STEPS 94
+sidebar_data sidebar_data_vals;
```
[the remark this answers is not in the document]
bbright024: ok, done - i had to leave 

```
typedef struct {                                                                                                                                                                                                   
    int value;                                                                                                                                                                                                     
    int target;                                                                                                                                                                                                    
} objective;                                                                                                                                                                                                       
 ```

above the function declarations though, because one of the functions uses objective in a parameter

File: src/widget/sidebar.c (hunk 12 above)
Comment (bbright024) [reply src/widget/sidebar.c:329]: hmmm what about adding a `widgets/sidebar_constants.h` file for these kind of values?  while there may be some future where we have to use an x value that is different than SIDEBAR_EXPANDED_WIDTH, such a future seems unlikely, and i think it makes the code less clear to have sidebar_width passed around when it's purely the value of that constant.

File: src/widget/sidebar.h (hunk 13 above)
Comment (bbright024) [reply src/widget/sidebar.h:5]: cause of the fun i had with constants; see prior comment

File: src/widget/sidebar_filler.c (hunk 14 above)
Comment (bbright024) [reply src/widget/sidebar_filler.c:34]: cause of the fun i had with constants; see prior comment

File: src/widget/sidebar_filler.c (hunk 16 above)
Comment (bbright024) [reply src/widget/sidebar_filler.c]: yeah, good point - i think i see what you've been doing.  you're only using prototypes when the file's static variables use them, right?

e.g.  `static arrow_button arrow_buttons_speed[]` has a function pointer to button_game_speed

File: src/widget/sidebar.c [hunk 20]
```diff
@@ -648,21 +454,21 @@ static void draw_sliding_foreground(void)
     draw_overlay_text(x_offset_expanded + 4);
     draw_build_image(x_offset_expanded + 6, 1);
 
-    draw_sidebar_filler(x_offset_collapsed, FILLER_Y_OFFSET, 1);
+    sidebar_filler_draw_sidebar_filler(x_offset_collapsed, FILLER_Y_OFFSET, 1);
 
     draw_sidebar_remainder(x_offset_expanded, 0);
-    draw_extra_info_buttons(x_offset_expanded, 0);
+    sidebar_filler_draw_extra_info_buttons(x_offset_expanded, 0);
```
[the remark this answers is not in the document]
bbright024: hmmm, draw_extra_info_buttons for sure, but draw_sidebar_remainder would have to happen with or without sidebar_filler stuff happening, since the sidebar always goes to the bottom of the screen.  

can i leave that one in sidebar.c?

thinking about it more... i wonder if all calls to sidebar_filler functions in sidebar.c should be done from within draw_sidebar_remainder in sidebar.c, since the filler is inserted inside the remainder area.  gonna try this out, let me know what you think!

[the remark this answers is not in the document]
bbright024: ohh its already mostly like that, just the slide function was doing it differently.  throwing another arg to draw_remainder to indicate a slide is occurring should clean this up.

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,290 @@
...
+    objective culture;
+    objective prosperity;
+    objective peace;
+    objective favor;
+    objective population;
+} data;
+
+int sidebar_filler_enabled(void)
+{
+  return data.height;
```
crudelios, line 79: Please fix the identation.

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,290 @@
...
+}
+
+int sidebar_filler_handle_mouse(const mouse *m, int x_offset, int y_offset)
+{
+    int click = 0;
+    click |= arrow_buttons_handle_mouse(m, x_offset, FILLER_Y_OFFSET, arrow_buttons_speed, 2);
```
crudelios, line 85: I guess for now you can just `return arrow_buttons_hanle_mouse(...)`

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,290 @@
...
+    } else {
+        arrow_buttons_draw(x_offset, FILLER_Y_OFFSET, arrow_buttons_speed, 2);
+    }
+
+    graphics_reset_clip_rectangle();
+}
+
+int sidebar_filler_extra_info_height_game_speed_check(void)
+{
+  return data.height >= EXTRA_INFO_HEIGHT_GAME_SPEED;
```
crudelios, line 264: Identation problem as well.

src/widget/sidebar.c
bbright024: eh, i went ahead and removed the sidebar_constants.h after everything else was moved over, and added sidebar_width/height as params to sidebar_filler.  definitely cleaner, thanks again for the input :-) resolving

src/widget/sidebar_filler.c
bbright024: yeah, i agree that it's more code than presently needed; i did it this way because it sets the style for future button additions to the extra info panel.

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,295 @@
...
+int sidebar_filler_get_filler_height(void)
+{
+    return data.height;
+}
+
+int sidebar_filler_handle_mouse(const mouse *m, int x_offset, int y_offset)
+{
+    int click = 0;
+    click |= arrow_buttons_handle_mouse(m, x_offset, y_offset, arrow_buttons_speed, 2);
+    return click;
```
bvschaik, line 91: This can also be one line. I've seen your comment that we can easily build on this when adding new buttons, but we don't have those yet. This is the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle 😉

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,295 @@
...
+}
+
+static int sidebar_filler_calculate_extra_info_height(int y_offset, int is_collapsed)
```
bvschaik, line 94: For internal functions there's no need to prefix, this can just be called `calculate_extra_info_height`. Same goes for the other `static` functions below.

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,295 @@
...
+{
+    if (is_collapsed || !config_get(CONFIG_UI_SIDEBAR_INFO)) {
+        data.height = 0;
+    } else {
+        int available_height = screen_height() - y_offset;
```
bvschaik, line 99: Using `screen_height()` implicitly assumes that the sidebar flows to the edge of the screen. That might not be true in the future (and wasn't true in the past). Better pass the available height instead.

src/widget/sidebar_filler.h
```diff
@@ -0,0 +1,13 @@
+#ifndef WIDGET_SIDEBAR_FILLER_H
+#define WIDGET_SIDEBAR_FILLER_H
+
+#include "input/mouse.h"
+
+void sidebar_filler_draw_filler_foreground(int x_offset, int y_offset, int sidebar_width, int is_collapsed);
+void sidebar_filler_draw_filler_background(int x_offset, int y_offset, int sidebar_width, int is_collapsed);
+
+int sidebar_filler_get_filler_height(void);
```
bvschaik, line 9: Naming: the function name contains "filler" twice now. Just `sidebar_filler_draw_foreground` will do and is in line with the rest of the project's naming. Same goes for the other two functions.

src/widget/sidebar.c
```diff
@@ -118,8 +119,19 @@ static struct {
...
     int focus_button_for_tooltip;
+    int height;
 } data;
 
+static void set_sidebar_height(int override)
+{
+    if (!override || override < 0) {
+        data.height = screen_height() - TOP_MENU_HEIGHT;
+    }
+    else {
```
crudelios, line 130: Just the newline :wink:

bbright024: d'oh! fixed, rebased, pushed

src/widget/sidebar_filler.h
```diff
@@ -0,0 +1,13 @@
+#ifndef WIDGET_SIDEBAR_FILLER_H
+#define WIDGET_SIDEBAR_FILLER_H
+
+#include "input/mouse.h"
+
+void sidebar_filler_draw_foreground(int x_offset, int y_offset, int sidebar_width, int is_collapsed);
+void sidebar_filler_draw_background(int x_offset, int y_offset, int sidebar_width, int is_collapsed, int sidebar_height);
```
bvschaik, line 7: Rename `sidebar_width` to just `width`, and `sidebar_height` to `available_height` to make it more clear what the variables should be (see my comment in `sidebar.c`).
Also, place the width and height arguments next to each other, since they're related.

src/widget/sidebar_filler.c
```diff
@@ -0,0 +1,258 @@
...
+{
+    return arrow_buttons_handle_mouse(m, x_offset, y_offset, arrow_buttons_speed, 2);
+}
+
+static int calculate_extra_info_height(int y_offset, int is_collapsed, int sidebar_height)
+{
+    if (is_collapsed || !config_get(CONFIG_UI_SIDEBAR_INFO)) {
+        data.height = 0;
+    } else {
+        int available_height = sidebar_height - y_offset;
```
bvschaik, line 72: Now you implicitly assume that the sidebar starts at the top of the screen. Just pass `available_height` directly from `sidebar.c` instead of calculating it here.

PR #424, opened 2020-04-24
Enable translation of config screens and popups

bvschaik:
Thanks @crudelios for the first translation: Portuguese.


crudelios:
The code looks fine and seems to be working 👍

PR #427, opened 2020-04-25
German translation

m-mohr:
German translation as requested in #415.

I couldn't compile so I'm not sure which texts are overly long as German tends to be quite verbose, but I tried to keep them short. Happy to fix any texts that are too long.

For some phrases I chose to not translate "word by word", but instead chose a similar meaning to make the translation shorter as otherwise it would probably not fit in the textbox.

Feel free to squash the commits.


m-mohr:
Nothing exceeding the available space? Wow :-D

bvschaik:
I assumed that you'd already made sure of that 😉

PR #431, opened 2020-04-28
Fix reservoir range showing outside the map

### LINKED ISSUE #430: Reservoir range gives artifacts when hovering near map edge
This only seems to happen on enormous maps at the north and south edges.
reservoir-map-edge

@crudelios is going to pick this up.

crudelios:
Fixes #430

### DIFF
--- a/src/city/view.c
+++ b/src/city/view.c
@@ -555,11 +555,14 @@ void city_view_foreach_valid_map_tile(map_callback *callback1, map_callback *cal
 
 static int move_grid_offset_by_delta(int grid_offset, int x, int y)
 {
+    if (!map_grid_is_valid_offset(grid_offset)) {
+        return -1;
+    }
     int grid_x = grid_offset % GRID_SIZE;
     int grid_y = grid_offset / GRID_SIZE;
     if (grid_x + x < 0 || grid_x + x >= GRID_SIZE ||
         grid_y + y < 0 || grid_y + y >= GRID_SIZE) {
-        return 0;
+        return -1;
     }
     return grid_offset + map_grid_delta(x, y);
 }
--- a/src/city/view.c
+++ b/src/city/view.c
@@ -553,6 +553,17 @@ void city_view_foreach_valid_map_tile(map_callback *callback1, map_callback *cal
     }
 }
 
+static int move_grid_offset_by_delta(int grid_offset, int x, int y)
+{
+    int grid_x = grid_offset % GRID_SIZE;
+    int grid_y = grid_offset / GRID_SIZE;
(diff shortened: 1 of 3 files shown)


src/city/view.c
```diff
@@ -581,17 +592,17 @@ void city_view_foreach_tile_in_range(int grid_offset, int size, int radius, map_
...
-            do_valid_callback(x + x_delta * tile, y - y_offset + y_delta * tile, offset_north + tile, callback);
-            do_valid_callback(x - x_delta * tile, y - y_offset + y_delta * tile, offset_north + tile * GRID_SIZE, callback);
-            do_valid_callback(x + x_delta * tile, y + y_offset - y_delta * tile, offset_south - tile * GRID_SIZE, callback);
-            do_valid_callback(x - x_delta * tile, y + y_offset - y_delta * tile, offset_south - tile, callback);
+            do_valid_callback(x + x_delta * tile, y - y_offset + y_delta * tile, move_grid_offset_by_delta(grid_offset, tile + offset_north, offset_north), callback);
+            do_valid_callback(x - x_delta * tile, y - y_offset + y_delta * tile, move_grid_offset_by_delta(grid_offset, offset_north, tile + offset_north), callback);
+            do_valid_callback(x + x_delta * tile, y + y_offset - y_delta * tile, move_grid_offset_by_delta(grid_offset, offset_south, offset_south - tile), callback);
+            do_valid_callback(x - x_delta * tile, y + y_offset - y_delta * tile, move_grid_offset_by_delta(grid_offset, offset_south - tile, offset_south), callback);
```
bvschaik, line 605: Please break up the lines so they're not too long: breaking just before the `move_...` function calls should be good.

src/city/view.c
```diff
@@ -553,6 +553,17 @@ void city_view_foreach_valid_map_tile(map_callback *callback1, map_callback *cal
...
+    if (grid_x + x < 0 || grid_x + x >= GRID_SIZE ||
+        grid_y + y < 0 || grid_y + y >= GRID_SIZE) {
+        return 0;
+    }
+    return grid_offset + map_grid_delta(x, y);
+}
+
 static void do_valid_callback(int x, int y, int grid_offset, map_callback *callback)
 {
     if (map_image_at(grid_offset) >= 6) {
```
bvschaik, line 569: Shouldn't we check for `grid_offset != 0` as well?

crudelios: In the current usage it makes no difference, as `grid_offset` is guaranteed to be valid.

bvschaik: I know, but it saves a call to `map_image_at` and is future-proof, in case we start to return for example `-1` for invalid offsets.

PR #434, opened 2020-04-30
Refactor city/empire scroll

crudelios:
The code changes are more or less these:

- Make right-click drag and touch scroll share the same code
- Create a speed struct that handles scroll acceleration, deceleration and keeps constant speed regardless of actual FPS
- Make `scroll_get_delta` handle every scroll, instead of touch and drag changing position separately from the rest
- Improve end-of-touch detection to make decaying slightly more responsive.

Scrolling can still be improved. Mainly, I still need to change the way `data.speed.modifier` is used to allow dynamic scroll speeds based on joystick axis input. However, since I'll be short on time for the next few weeks, I'm submitting this as-is.

Due to the extreme changes, there may be some bugs left. Please test with both smooth scrolling on and off.


src/input/scroll.c
```diff
@@ -242,131 +192,87 @@ void scroll_set_custom_margins(int x, int y, int width, int height)
...
+    if (data.is_scrolling) {
+        time_millis current_time = time_get_millis();
+        time_millis delta_time = current_time - data.last_time;
+
+        // SDL touch events sometimes incorrectly set movement to zero before the finger is released
+        // This gives a time tolerance before setting the movement to 0
+        if (!data.drag.is_touch ||
+            data.drag.delta.x != 0 || data.drag.delta.y != 0 ||
+            delta_time > DRAG_STOPPED_TIME_TOLERANCE) {
+            data.last_time = current_time;
```
crudelios, line 249: I had a better look at this, this isn't an SDL bug, but my own, in `platform/touch.c`. So don't merge until I properly fix this, please.

crudelios:
All the problems should be fixed now, so this is ready for review :+1:

src/core/speed.h
```diff
@@ -0,0 +1,73 @@
...
+
+typedef struct {
+    time_millis start_time;
+    time_millis total_time;
+    time_millis last_speed_check;
+    int speed_difference;
+    int desired_speed;
+    int current_speed;
+    int cumulative_delta;
+} speed;
```
bvschaik, line 35: Call this `speed_type` instead. That way you can name the function arguments/variables just `speed` instead of `sp`, making the code clearer.

src/core/speed.h
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Changes the speed
+ * @param sp Speed structure to act on
+ * @param new_speed The new speed
+ * @param total_time the time for the acceleration/deceleration, or SPEED_CHANGE_IMMEDIATE
+ */
+void speed_change(speed *sp, int new_speed, time_millis total_time);
+
+/**
+ * Immediately the speed (positive speed becomes negative and vice-versa)
```
bvschaik, line 52: "Immediately the speed" is not a sentence 😉

src/core/speed.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+void speed_clear(speed *sp);
+
+/**
+ * Changes the speed
+ * @param sp Speed structure to act on
+ * @param new_speed The new speed
+ * @param total_time the time for the acceleration/deceleration, or SPEED_CHANGE_IMMEDIATE
+ */
+void speed_change(speed *sp, int new_speed, time_millis total_time);
```
bvschaik, line 49: Changes the speed in what way? I think something like `speed_set_target_speed` would be a better name.

src/platform/mouse.c
```diff
@@ -17,16 +18,18 @@ void system_mouse_get_relative_state(int *x, int *y)
 void system_mouse_set_relative_mode(int enabled)
 {
     if (enabled) {
+        if (data.enabled) {
```
bvschaik, line 21: Suggestion: do the following check at the start of the function, so you don't need to have another check both here and in the `else if` below:
```
if (enabled == data.enabled) {
    return;
}
```

src/input/scroll.c
```diff
@@ -4,23 +4,26 @@
 #include "core/config.h"
 #include "core/direction.h"
 #include "core/time.h"
+#include "core/speed.h"
```
bvschaik, line 7: S comes before T 😉

src/core/speed.h
crudelios: How about just `speed_set_target`? Otherwise the word speed is repeated...

bvschaik: True, that's also fine.

src/platform/mouse.c
```diff
@@ -17,20 +17,19 @@ void system_mouse_get_relative_state(int *x, int *y)
...
 void system_mouse_set_relative_mode(int enabled)
 {
-    if (enabled) {
-        if (data.enabled) {
-            return;
-        }
+    if (enabled == data.enabled) {
+        return;
+    }
+    if(enabled) {
```
bvschaik, line 23: One more `if(` :wink:

crudelios: I blame Visual Studio, which doesn't do as it's told!

PR #438, opened 2020-05-06
Allow touching anywhere on the building ghost to confirm construction

crudelios:
Useful little improvement :)

### DIFF
--- a/src/widget/city.c
+++ b/src/widget/city.c
@@ -1,9 +1,11 @@
 #include "city.h"
 
 #include "building/construction.h"
+#include "building/properties.h"
 #include "city/finance.h"
 #include "city/view.h"
 #include "city/warning.h"
+#include "core/direction.h"
 #include "core/string.h"
 #include "figure/formation_legion.h"
 #include "game/settings.h"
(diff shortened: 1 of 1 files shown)


src/widget/city.c
```diff
@@ -293,41 +328,46 @@ static void handle_first_touch(map_tile *tile)
...
+    int size = building_properties_for_type(type)->size;
+    if (type == BUILDING_WAREHOUSE) {
+        size = 3;
+    }
+
     if (touch_was_click(first) && first->has_ended && data.capture_input &&
-        data.selected_grid_offset == tile->grid_offset) {
-        build_start(tile);
-        build_move(tile);
+        has_confirmed_construction(data.selected_tile.grid_offset, tile->grid_offset, size)) {
```
bvschaik, line 364: How does this work for hippodromes and forts?

crudelios: Only the first section of each building (hippodrome and fort) can be clicked, clicking anywhere else moves the ghost. I know it's not perfect for those buildings, but it seemed a bit overkill to add the code for the entire ghost in those cases.

bvschaik: Fair enough.

PR #449, opened 2020-05-13
More changes to city/empire scrolling

crudelios:
* Fix bug when adjusting to grid while moving up or left
* Slightly refactor scrolling code by removing external calls to scroll_drag_move
* Reduce scroll lag by one frame
* Add support for joystick arrow input

These should be all the remaining changes to be able to proceed with joystick support.

### DIFF
--- a/src/input/scroll.h
+++ b/src/input/scroll.h
@@ -20,12 +20,11 @@ void scroll_restore_margins(void);
 int scroll_get_delta(const mouse *m, pixel_offset *delta, scroll_type type);
 
 void scroll_drag_start(int is_touch);
-void scroll_drag_move(void);
 int scroll_drag_end(void);
 
-void scroll_arrow_left(int is_down);
-void scroll_arrow_right(int is_down);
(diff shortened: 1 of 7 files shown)


src/input/scroll.h
```diff
@@ -20,12 +20,11 @@ void scroll_restore_margins(void);
...
 int scroll_drag_end(void);
 
-void scroll_arrow_left(int is_down);
-void scroll_arrow_right(int is_down);
-void scroll_arrow_up(int is_down);
-void scroll_arrow_down(int is_down);
+void scroll_arrow_left(int value);
+void scroll_arrow_right(int value);
+void scroll_arrow_up(int value);
+void scroll_arrow_down(int value);
```
bvschaik, line 28: What should be passed as value here? Do we need constants/enum for those?

crudelios: > What should be passed as value here?

The joystick axis input, which starts from `0`, has a dead zone of `2000` and then goes up to `32768`.

PR #463, opened 2020-05-24
Added Russian translation for building prices in the GUI

agruzdev:
Might need additional translation to Korean and other languages, anyway for now there is no noticable difference for them


crudelios:
Thank you for this PR, however this is not strictly needed.

Even though the word `Dn` is hardcoded into the file, it is actually part of the language file (either `c3.eng` or `c3.rus`) that comes with Caesar 3, so we already have the translations available.

What we need is to change the string `Dn` to the proper string based on the translation file.

agruzdev:
@crudelios Yes, it's not a big deal to fix and make through the language file. I'm not sure, do such language files exist for the other laguages like german or korean? I see hardcoded names for the files  c3.eng and c3.rus, not for others though. Will it be an universal solution?

crudelios:
> Will it be an universal solution?

It will. All the other languages use the `c3.eng` file, Russian is the only exception.

agruzdev:
@crudelios Alright, thank you for the clarification. Let me do it in the proper way then.

PR #464, opened 2020-05-24
Fixed price translation in the building menu

agruzdev:
Taking price postfix from the language file

### DIFF
--- a/src/graphics/text.c
+++ b/src/graphics/text.c
@@ -1,5 +1,6 @@
 #include "text.h"
 
+#include "core/lang.h"
 #include "core/string.h"
 #include "core/time.h"
 #include "graphics/graphics.h"
@@ -8,6 +9,7 @@
 #include <string.h>
 
 #define ELLIPSIS_LENGTH 4
+#define TEXT_BUFFER_LENGTH 100
 
 static uint8_t tmp_line[200];
 
@@ -261,7 +263,7 @@ int text_draw(const uint8_t *str, int x, int y, font_t font, color_t color)
     return current_x - x;
 }
 
-static void number_to_string(uint8_t *str, int value, char prefix, const char *postfix)
+static int number_to_string(uint8_t *str, int value, char prefix, const char *postfix)
 {
     int offset = 0;
     if (prefix) {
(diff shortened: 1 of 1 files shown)


src/graphics/text.c
```diff
@@ -273,53 +275,58 @@ static void number_to_string(uint8_t *str, int value, char prefix, const char *p
...
 }
 
 int text_draw_money(int value, int x_offset, int y_offset, font_t font)
 {
-    uint8_t str[100];
-    number_to_string(str, value, '@', " Dn");
+    uint8_t str[NUMBER_BUFFER_LENGTH];
+    int money_len = number_to_string(str, value, '@', " ");
+    const uint8_t *postfix = lang_get_string(6, 0);
+    if (postfix && money_len < NUMBER_BUFFER_LENGTH - 1) {
```
crudelios, line 300: You don't really need the second part of the `if`.

`string_copy` will truncate the string if it is too long, which is better since at least is shows some text.

So, my suggestion is to just do the following:

```suggestion
    if (postfix) {
```

crudelios:
You didn't have to replace all magic `100` with the `#define` here, so thanks for that!

src/graphics/text.c
agruzdev: I added it, becuase string_copy() doesn't work very robust when maxlength == 0, in that case it tries to rewrite symbol before the 'src' pointer
So this condition checks that there is at least 1 free symbol.
I do understand that there will be corect place in src[-1], but I felt like not relying on it

crudelios: I get what you mean, but the truth is `string_copy` already makes a few assumptions to save CPU cycles (like `src` and `dst` never being null pointers) which in the current type-safe world may send some shivers down the spine. In fact, if you mess too much with the translation files, you *will* crash the game.

However, since we more or less control the strings that get fed into `string_copy`, it should be safe to use.

agruzdev: Alright, if we take such assumptions, than we can get rid of that check

crudelios:
Lgtm 👍

PR #468, opened 2020-05-28
Make game speed independent of fps

crudelios:
Before, at speeds at or above 100%, the game played at 1 (or 2 at 200%, 3 at 300%, etc...) in-game ticks per frame. This caused 100% game speed to be much faster on 144hz monitors and slower when fps was low.

This PR fixes that, and now the game always runs at the same speed regardless of fps, with a maximum limit of 20 ticks per frame.

This also effectively frameskips when fps is low, mantaining the in-game speed consistent.

There are some slight differences in speed, that aren't really noticeable on speeds below 100% but are palpable in the 200% and 300% game speeds (for the old code, the speeds are based on a 60fps framerate):

```
         |  MS per tick - avg
   Game  |--------------------
  speed  | Old code | New code
------------------------------
   10%   | 502.00   | 502.47
   20%   | 352.00   | 344.15
   30%   | 242.00   | 235.72
   40%   | 162.00   | 161.45
   50%   | 112.00   | 110.58
   60%   |  82.00   |  75.74
   70%   |  57.00   |  51.87
   80%   |  37.00   |  35.53
   90%   |  22.00   |  24.33
  100%   |  16.67   |  16.67
  200%   |   8.33   |  11.41
  300%   |   5.55   |   7.82
  400%   |   4.16   |   5.35
  500%   |   3.33   |   3.66
```

### DIFF
--- a/src/game/file.c
+++ b/src/game/file.c
@@ -96,7 +96,7 @@ static void clear_scenario_data(void)
     building_construction_clear_type();
     city_data_init();
     city_message_init_scenario();
-    game_state_init();
+    game_state_init(setting_game_speed());
     game_animation_init();
     sound_city_init();
     building_menu_enable_all();
@@ -238,6 +238,8 @@ static void initialize_saved_game(void)
     map_tiles_determine_gardens();
 
     city_message_clear_scroll();
+
+    game_state_set_speed(setting_game_speed(), 1);
 
     game_state_unpause();
 }
--- a/src/game/file_editor.c
+++ b/src/game/file_editor.c
@@ -55,7 +55,7 @@ void game_file_editor_clear_data(void)
     city_data_init();
     city_data_init_scenario();
     city_message_init_scenario();
-    game_state_init();
+    game_state_init(70); // speed at 70%, nice speed for flag animations
     game_animation_init();
     sound_city_init();
     building_menu_enable_all();
--- a/src/game/game.c
+++ b/src/game/game.c
@@ -2,6 +2,7 @@
 
 #include "building/construction.h"
 #include "building/model.h"
+#include "core/calc.h"
 #include "core/config.h"
 #include "core/hotkey_config.h"
 #include "core/image.h"
@@ -9,7 +10,6 @@
 #include "core/locale.h"
 #include "core/log.h"
 #include "core/random.h"
-#include "core/time.h"
 #include "editor/editor.h"
 #include "figure/type.h"
 #include "game/animation.h"
@@ -32,11 +32,7 @@
 #include "window/logo.h"
 #include "window/main_menu.h"
 
-static const time_millis MILLIS_PER_TICK_PER_SPEED[] = {
-    0, 20, 35, 55, 80, 110, 160, 240, 350, 500, 700
-};
-
-static time_millis last_update;
+#define MAX_TICKS_PER_FRAME 20
 
 static void errlog(const char *msg)
 {
@@ -112,7 +108,7 @@ int game_init(void)
     }
 
     sound_system_init();
-    game_state_init();
+    game_state_init(setting_game_speed());
     window_logo_show(missing_fonts ? MESSAGE_MISSING_FONTS : (is_unpatched() ? MESSAGE_MISSING_PATCH : MESSAGE_NONE));
 
     return 1;
@@ -175,11 +171,12 @@ int game_reload_language(void)
 
 static int get_elapsed_ticks(void)
 {
-    if (game_state_is_paused()) {
+    int ticks_this_frame = game_state_get_ticks();
+    if (game_state_is_paused() ||
+        building_construction_in_progress() ||
+        (scroll_in_progress() && !scroll_is_smooth())) {
         return 0;
     }
-    int game_speed_index = 0;
-    int ticks_per_frame = 1;
     switch (window_get_id()) {
         default:
             return 0;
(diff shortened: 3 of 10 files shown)


bvschaik:
Call me old-fashioned, but using floating-point arithmetic in a place where integers will suffice kinda bothers me from a performance perspective. The speed code is quite complex, mainly used for transitioning the speed from one value to another, and now we're going to use it for relatively simple increments.

Also, this code relies on the counter ticking on when the map is not in view, which introduces a deviation from Caesar 3: upon unpausing or returning to the map from another window, in C3 the map ALWAYS advances a tick. With this code, it may or may not, depending on how far along we are. This is most noticeable when setting the speed to 10% and unpause-pause quickly: most of the time it does not advance a tick when doing that.

I've toyed around with it, and found that this seems to work. I cannot test it on higher-frequency monitors though, can you perhaps have a look?
```
diff --git a/src/game/game.c b/src/game/game.c
index 5b93da56..976b88f0 100644
--- a/src/game/game.c
+++ b/src/game/game.c
@@ -33,10 +33,16 @@
 #include "window/main_menu.h"
 
 static const time_millis MILLIS_PER_TICK_PER_SPEED[] = {
-    0, 20, 35, 55, 80, 110, 160, 240, 350, 500, 700
+    0, 702, 502, 352, 242, 162, 112, 82, 57, 37, 22, 16
+};
+static const time_millis MILLIS_PER_HYPER_SPEED[] = {
+    0, 16, 8, 5, 3, 2
 };
 
-static time_millis last_update;
+static struct {
+    int last_check_was_valid;
+    time_millis last_update;
+} data;
 
 static void errlog(const char *msg)
 {
@@ -175,11 +181,12 @@ int game_reload_language(void)
 
 static int get_elapsed_ticks(void)
 {
+    int last_check_was_valid = data.last_check_was_valid;
+    data.last_check_was_valid = 0;
     if (game_state_is_paused()) {
         return 0;
     }
-    int game_speed_index = 0;
-    int ticks_per_frame = 1;
+    int millis_per_tick = 1;
     switch (window_get_id()) {
         default:
             return 0;
@@ -187,17 +194,22 @@ static int get_elapsed_ticks(void)
         case WINDOW_CITY_MILITARY:
         case WINDOW_SLIDING_SIDEBAR:
         case WINDOW_OVERLAY_MENU:
-        case WINDOW_BUILD_MENU:
-            game_speed_index = (100 - setting_game_speed()) / 10;
-            if (game_speed_index >= 10) {
+        case WINDOW_BUILD_MENU: {
+            int speed = setting_game_speed();
+            if (speed < 10) {
                 return 0;
-            } else if (game_speed_index < 0) {
-                ticks_per_frame = setting_game_speed() / 100;
-                game_speed_index = 0;
+            } else if (speed <= 100) {
+                millis_per_tick = MILLIS_PER_TICK_PER_SPEED[speed / 10];
+            } else {
+                if (speed > 500) {
+                    speed = 500;
+                }
+                millis_per_tick = MILLIS_PER_HYPER_SPEED[speed / 100];
             }
             break;
+        }
         case WINDOW_EDITOR_MAP:
-            game_speed_index = 3; // 70%, nice speed for flag animations
+            millis_per_tick = MILLIS_PER_TICK_PER_SPEED[7]; // 70%, nice speed for flag animations
             break;
     }
     if (building_construction_in_progress()) {
@@ -208,12 +220,18 @@ static int get_elapsed_ticks(void)
     }
 
     time_millis now = time_get_millis();
-    time_millis diff = now - last_update;
-    if (diff < MILLIS_PER_TICK_PER_SPEED[game_speed_index] + 2) {
+    time_millis diff = now - data.last_update;
+    data.last_check_was_valid = 1;
+    if (!last_check_was_valid) {
+        // returning to map from another window or pause: always force a tick
+        data.last_update = now;
+        return 1;
+    }
+    if (diff < millis_per_tick) {
         return 0;
     }
-    last_update = now;
-    return ticks_per_frame;
+    data.last_update = now - (diff % millis_per_tick);
+    return diff / millis_per_tick;
 }
 
 void game_run(void)
```

crudelios:
I won't be home for the next two weeks. My monitor is also 60fps, so the way I check is to disable vsync on `platform/screen.c` and introduce a delay (using `SDL_Delay`) on every frame to simulate a certain fps.

If you want to test you can use that "technique", otherwise I'll only be able to do it in two weeks!

attrition:
If you can tell me what to look for I can check what happens on high fps monitors. I have 144hz to test with.

bvschaik:
> If you can tell me what to look for I can check what happens on high fps monitors. I have 144hz to test with.

Thanks!
Could you get [this build](https://bintray.com/bvschaik/julius-branches/windows/improve-game-speed-1.4.1.88-662bd220#files) and see how long it takes for 1 month to pass on 100% game speed? It should be around 13 seconds. On regular Julius builds, it's most likely running about twice as fast on a 144Hz monitor.
Also, going from 90% to 100% should be a slight increase in speed, just like going from 80% to 90%, not a big jump, like it probably is now.

attrition:
> 
> 
> > If you can tell me what to look for I can check what happens on high fps monitors. I have 144hz to test with.
> 
> Thanks!
> Could you get [this build](https://bintray.com/bvschaik/julius-branches/windows/improve-game-speed-1.4.1.88-662bd220#files) and see how long it takes for 1 month to pass on 100% game speed? It should be around 13 seconds. On regular Julius builds, it's most likely running about twice as fast on a 144Hz monitor.
> Also, going from 90% to 100% should be a slight increase in speed, just like going from 80% to 90%, not a big jump, like it probably is now.

13 seconds @ 144hz @ 100%. 90 to 100% was a small uptick in speed as expected.

attrition:
Also as a sanity check, the latest stable build of Julius passed a month in about 5-6 seconds at 100% at 144hz.

crudelios:
Can you also test/compare at 200% please? @bvschaik what time do you get on your 60fps monitor to compare? If it checks out I'll close tbis PR so you can add your changes.

bvschaik:
I actually tweaked the speeds for > 100% slightly in my branch since they will probably only be used to speed up the game significantly (for testing those keep-em-stable-for-a-century cities).

The following table lists the current code's calculated speeds at 60 and 144Hz with vsync enabled. This is assuming the computer is powerful enough to run 1 game tick within one screen refresh. I measured the ones for 100 and 200% on my 60Hz screen and they match up. The last column is what my new code does, regardless of screen refresh rate.

    Game  |     Month (seconds)
    Speed | 60Hz  | 144Hz | New code
    -------------------------------
     100% | 13.33 |  5.56 | 12.80
     200% |  6.67 |  2.78 |  6.40
     300% |  4.44 |  1.85 |  4.00
     400% |  3.33 |  1.39 |  2.40
     500% |  2.67 |  1.11 |  1.60

attrition:
@ 144hz:
400% was just shy of 3 seconds and 500% was just about 2 seconds. It looks to be working correctly from your given chart.

bvschaik:
@attrition Thanks for testing, the code is now in master.

PR #475, opened 2020-06-03
Get rid of specific Vita/Switch touch input modes

crudelios:
Adapt them to the main code instead.

In doing so, two and three finger drag have been disabled since they didn't seem very usable.

Also fixed some inconsistencies with mouse cursor location and made Julius touch mode remove the cursor on the Vita/Switch.


src/input/touch.c
```diff
@@ -214,9 +299,30 @@ int touch_to_mouse(void)
...
-    mouse_set_from_touch(first, get_latest_touch());
+    switch (data.mode) {
+        case TOUCH_MODE_TOUCHPAD:
+            handle_mouse_touchpad();
+            break;
+        case TOUCH_MODE_DIRECT:
+            handle_mouse_direct();
+            break;
+        default:
+            mouse_set_from_touch(first, get_latest_touch());
```
bvschaik, line 315: Also add a `break;` here:
```suggestion
            mouse_set_from_touch(first, get_latest_touch());
            break;
```

PR #477, opened 2020-06-04
Multibyte input

bvschaik:
For Korean and Chinese versions of the game, this allows users to type in their native language, instead of being restricted to Latin characters. Only characters displayable by the game are supported.

Also adds Chinese translation for the configuration screen and improves Latin character rendering in Chinese.


crudelios:
The code fprmatting LGTM, and I like that you abstracted keyboard text input handling inside `input_box`.

You need to fix the merge conflict. I still want to test deleting/adding multibyte chars to the middle of a text field, and handling texts larger than the input box.

crudelios:
Typing, deleting and moving through text seems to work, however I found a small bug with text size calculation:

city 2020-06-06 09 55 12

On the last file, the ellipsis should be close to the right of the window, but they are centered instead. That's probably because there is a bug with `text_ellipsize`.

bvschaik:
Thanks, I'll have a look at the ellipsize code.

crudelios:
It's working fine now 😃

PR #488, opened 2020-06-14
Change smooth scrolling default to enabled

devnoname120:
Smooth scrolling is not a gameplay change so I think that it's worth enabling by default.


bvschaik:
We already discussed this a while ago in chat.

The "it's not a gameplay change" argument can also be applied to all the other "UI changes" configuration options, that's no reason to enable any of them by default.

I want Julius to have the same feel as Caesar 3 out-of-the-box.

PR #492, opened 2020-06-19
Add scrollbar to config screen

bvschaik:


crudelios:
I just tested, the scrollbar works well, the language slider works fine but the mouse y position is misaligned by two checkboxes' height (48px).

bvschaik:
Good one. The y-offset misalignment actually depended on your resolution... and since I only tested this at 640x480 I never noticed. It's fixed now.

crudelios:
LGTM now!

PR #494, opened 2020-06-23
Refactor platform code

crudelios:
The aim of this commit is to further merge Vita/Switch code with base Julius code and help pave the way for the joystick support branch.

Add platform-dependant callbacks on game init and for each frame instead of using a specific ifdef for each platform.

Refactor virtual keyboard code, allowing Vita and Switch keyboards to automatically appear when pressing the textbox. It also allows the current text to appear when the textbox is clicked.

When the first iteratiion of joystick support is merged, the current text will also show when the proper gamepad button is pressed on Switch/Vita.


src/platform/julius.c
```diff
@@ -311,14 +307,13 @@ static void main_loop(void)
...
 #ifdef __vita__
-        vita_handle_analog_sticks();
-        vita_handle_virtual_keyboard();
         while (vita_poll_event(&event)) {
 #elif defined(__SWITCH__)
-        switch_handle_analog_sticks();
-        switch_handle_virtual_keyboard();
         while (switch_poll_event(&event)) {
 #else
         while (SDL_PollEvent(&event)) {
```
bvschaik, line 319: Can we also `#define` a poll event function so we don't have the preprocessor switch here?

src/platform/switch/switch_input.c
```diff
@@ -89,14 +91,37 @@ static uint8_t map_switch_button_to_sdlmousebutton[SWITCH_NUM_BUTTONS] =
...
+    if (vkbd_requested) {
+        switch_start_text_input();
+        vkbd_requested = 0;
+    }
+    switch_handle_analog_sticks();
+}
+
+void platform_show_virtual_keyboard(const uint8_t *text)
+{
+    vkbd_text = text ? (const char *)text : "";
```
bvschaik, line 118: Casting an internal string to C-string is not a good idea for non-ascii text.

src/core/encoding.h
```diff
@@ -78,13 +78,13 @@ int encoding_get_utf8_character_bytes(const char input);
...
-void encoding_utf16_to_utf8(uint16_t *input, uint8_t *output);
+void encoding_utf16_to_utf8(const uint16_t *input, uint8_t *output);
 
 /**
  * Converts an UTF-8 input input to UTF-16 output
  * @param input Input to convert, encoded using UTF-8
  * @param output Output buffer to store the UTF-16 encoded input
  */
-void encoding_utf8_to_utf16(uint8_t *input, uint16_t *output);
+void encoding_utf8_to_utf16(const uint8_t *input, uint16_t *output);
```
bvschaik, line 88: I know we merged this earlier, but I just realized this signature is confusing. We use `char*` for UTF-8 strings everywhere, and `uint8_t*` for the internally encoded strings. Here we pass UTF-8 in `uint8_t` instead of `char`. This is going to cause weird issues if we mix the two.
Also applies to the previous function.

src/platform/julius.c
crudelios: My plan is to completely get rid of platform-specific poll events, that's why I didn't do it here :wink:

If you want I can do as you requested, but it'll be a temporary change.

bvschaik: Then it's ok as-is, no need for a temporary change.

src/platform/switch/switch_input.c
crudelios: Hummm. We have a `string_from_ascii`, should I add a `string_to_ascii`? There's still the conversion problem though, I don't really know how Vita/Switch handle the encoding.

PR #496, opened 2020-07-05
add hotkeys for commonly needed tools/buildings

attrition:
# Overview
Coming back to C3 after many years (via Julius), I was surprised to see that hotkeys for tools like clearing and roads had no hotkeys. I made it about halfway through the campaign before beginning this effort. What is a "commonly used" tool/building is a little debatable but more can be added as we go.

This PR adds hotkeys for the following:
- clearing tool
- houses
- road
- walls
- gatehouse
- prefecture
- engineers' post
- granary
- warehouse
- market
- gardens
- plaza
- reservoir
- aquaduct
- fountain

# Approach
1) When the user presses a hotkey for a building, we pack the hotkey_definition with the BUILDING_[_enumeration text_]

2) When handling the hotkey in city.c we can then just unpack the building id and set the current construction type to match


attrition:
The Travis build failed on the Nintendo Switch build:
`The command "./.ci_scripts/setup_environment.sh" failed and exited with 1 during .`

I don't believe I have any control over this, let me know what can be done

crudelios:
The switch build has been giving trouble lately, so for now please make sure only the other builds work.

src/core/hotkey_config.c
```diff
@@ -26,6 +26,15 @@ static const char *ini_keys[] = {
     "decrease_game_speed",
     "rotate_map_left",
     "rotate_map_right",
+    "build_vacant_house",
+    "build_clear",
```
crudelios, line 30: Can you change to `build_clear_land`, please? It's... clearer that way!

src/core/hotkey_config.h
```diff
@@ -15,6 +15,15 @@ typedef enum {
     HOTKEY_DECREASE_GAME_SPEED,
     HOTKEY_ROTATE_MAP_LEFT,
     HOTKEY_ROTATE_MAP_RIGHT,
+    HOTKEY_BUILD_CLEAR,
```
crudelios, line 18: Same here. `HOTKEY_BUILD_CLEAR_LAND` please.

src/translation/english.c
```diff
@@ -52,6 +52,7 @@ static translation_string all_strings[] = {
     {TR_HOTKEY_HEADER_OVERLAYS, "Overlays"},
     {TR_HOTKEY_HEADER_BOOKMARKS, "City map bookmarks"},
     {TR_HOTKEY_HEADER_EDITOR, "Editor"},
+    {TR_HOTKEY_HEADER_BUILD, "Build hotkeys"},
```
crudelios, line 55: How about `Construction hotkeys` instead?

src/window/hotkey_config.c
```diff
@@ -55,6 +55,16 @@ static hotkey_widget hotkey_widgets[] = {
     {HOTKEY_CYCLE_LEGION, TR_HOTKEY_CYCLE_LEGION},
     {HOTKEY_ROTATE_MAP_LEFT, TR_HOTKEY_ROTATE_MAP_LEFT},
     {HOTKEY_ROTATE_MAP_RIGHT, TR_HOTKEY_ROTATE_MAP_RIGHT},
+    {HOTKEY_HEADER, TR_HOTKEY_HEADER_BUILD},
+    {HOTKEY_BUILD_CLEAR, TR_HOTKEY_BUILD_CLEAR},
```
crudelios, line 59: `TR_HOTKEY_BUILD_CLEAR_LAND` here too :wink:

src/core/hotkey_config.c
```diff
@@ -26,6 +26,22 @@ static const char *ini_keys[] = {
...
+    "build_prefecture",
+    "build_engineers_post",
+    "build_doctor",
+    "build_granary",
+    "build_warehouse",
+    "build_market",
+    "build_wall",
+    "build_gatehouse",
+    "build_reservoir",
+    "build_aquaduct",
```
bvschaik, line 43: In English it's aqu**e**duct:
```suggestion
    "build_aqueduct",
```

src/core/hotkey_config.h
```diff
@@ -15,6 +15,22 @@ typedef enum {
...
+    HOTKEY_BUILD_PREFECTURE,
+    HOTKEY_BUILD_ENGINEERS_POST,
+    HOTKEY_BUILD_DOCTOR,
+    HOTKEY_BUILD_GRANARY,
+    HOTKEY_BUILD_WAREHOUSE,
+    HOTKEY_BUILD_MARKET,
+    HOTKEY_BUILD_WALL,
+    HOTKEY_BUILD_GATEHOUSE,
+    HOTKEY_BUILD_RESERVOIR,
+    HOTKEY_BUILD_AQUADUCT,
```
bvschaik, line 32: AQUEDUCT here as well, and also check the translation constants.

src/translation/english.c
```diff
@@ -71,6 +72,22 @@ static translation_string all_strings[] = {
...
+    {TR_HOTKEY_BUILD_ENGINEERS_POST, "Build engineers' post"},
+    {TR_HOTKEY_BUILD_DOCTOR, "Build doctor"},
+    {TR_HOTKEY_BUILD_GRANARY, "Build granary"},
+    {TR_HOTKEY_BUILD_WAREHOUSE, "Build warehouse"},
+    {TR_HOTKEY_BUILD_MARKET, "Build market"},
+    {TR_HOTKEY_BUILD_WALL, "Build wall"},
+    {TR_HOTKEY_BUILD_GATEHOUSE, "Build gatehouse"},
+    {TR_HOTKEY_BUILD_RESERVOIR, "Build reservoir"},
+    {TR_HOTKEY_BUILD_AQUADUCT, "Build aquaduct"},
+    {TR_HOTKEY_BUILD_FOUNTAIN, "Build fountain"},
```
bvschaik, line 90: I was thinking, we already have the building names in the language files. If we say just "Prefecture" instead of "Build prefecture", we don't need translations into 11 languages for these. (it doesn't have to change in this PR, just something to consider for later)

src/window/city.c
```diff
@@ -18,6 +18,7 @@
 #include "map/bookmark.h"
 #include "map/grid.h"
 #include "scenario/criteria.h"
+#include "scenario/building.h"
```
bvschaik, line 21: Alphabetical order please :)
```suggestion
#include "scenario/building.h"
#include "scenario/criteria.h"
```

src/translation/english.c
attrition: That's a great point, and I think it's not confusing since it's under a clear header. The header isn't translated though.

bvschaik: FYI: I just implemented this in 21b48c183eb2a9d6a72c9f5c474e3b5a26d87397.
We can request translations for the header once the feature is stable.

PR #497, opened 2020-07-06
Enter construction mode with same building as cursor is currently over

attrition:
When pressing the designated hotkey, check the building at the currently selected map tile and set the construction mode to that type of building.

The rationale behind this change is that while hotkeys for buildings are nice, there are too many buildings to make this practical. This method of cloning what is under the cursor can help when expanding an already developed city, where if you need to expand some industry you can hover over the raw material building, press [hotkey], place more of the same building. Then hover over the workshop and [hotkey] and place those.

There are some edge cases around houses, forts, triumphal arches and native buildings which I have handled and wrote out in some comments.

The current implementation has one downside, which is the that the following aren't picked up by the clone hotkey because of how I retrieve the current building (`get_building`):  
- road
- plaza
- aqueduct 
- wall

If we want to expand the behaviour to those features as well then I may need a suggestion on how to do that.


src/widget/city.c
```diff
@@ -493,6 +493,11 @@ void widget_city_handle_input_military(const mouse *m, const hotkeys *h, int leg
     }
 }
 
+int widget_city_building_at_current_tile(void)
+{
+    return map_building_at(data.current_tile.grid_offset);
```
attrition, line 498: This may require more guards, feedback welcome

src/translation/english.c
```diff
@@ -64,6 +64,7 @@ static translation_string all_strings[] = {
     {TR_HOTKEY_RESIZE_TO_1024, "Resize window to 1024x768"},
     {TR_HOTKEY_SAVE_SCREENSHOT, "Save screenshot"},
     {TR_HOTKEY_SAVE_CITY_SCREENSHOT, "Save full city screenshot"},
+    {TR_HOTKEY_BUILD_CLONE, "Clone selected building"},
```
bvschaik, line 67: I think a better description would be "Clone building under cursor". "Selected" might confuse users.

src/window/city.c
```diff
@@ -155,6 +161,100 @@ static void cycle_legion(void)
     }
 }
 
+/**
```
bvschaik, line 164: All these added functions is rather a lot of logic to put in `window/city.c`, which should only be concerned with handling input events.
I suggest creating a new C file `building/clone.c` that contains this logic, with one public function `building_clone_from_grid_offset(int grid_offset)`, which has the implementation of `get_clone_type_from_grid_offset()`.

src/window/city.c
```diff
@@ -155,6 +161,100 @@ static void cycle_legion(void)
...
 
+/** 
+    Takes a building and retrieve its proper type for cloning.
+    For example, given a fort, return the enumaration value corresponding to
+    the specific type of fort rather than the general value
+
+    @param building Building to examine
+    @return the building_type value to clone, or BUILDING_NONE if not cloneable
+*/
+static short get_clone_type_from_building(building* building)
```
bvschaik, line 172: Pointer `*` on the other side.

src/window/city.c
```diff
@@ -155,6 +161,100 @@ static void cycle_legion(void)
...
 }
 
+/** 
+    Takes a building and retrieve its proper type for cloning.
+    For example, given a fort, return the enumaration value corresponding to
+    the specific type of fort rather than the general value
+
+    @param building Building to examine
+    @return the building_type value to clone, or BUILDING_NONE if not cloneable
+*/
```
bvschaik, line 171: For documentation comments, please use the indentation you also see in header files: start every line with ` * ` rather than indenting by 4 spaces.

attrition:
> Functionality is OK, please just move some code around to keep files small.

Thanks for taking a look, please let me know if there are any other changes needed!

bvschaik:
> Thanks for taking a look, please let me know if there are any other changes needed!

Check your discord messages please, I'd like to double-check something before I merge.

attrition:
Thank you!

PR #498, opened 2020-07-06
Fix construct housing translation string showing as "Hospital"

attrition:
Just a typo fix


bvschaik:
Actually it was not a typo but my stupidity :wink:
345edcd71bd588bf2caac1d1d5747dfd600c7f44 fixes it.

PR #500, opened 2020-07-07
make resource icons on the empire screen buttons for quick trading access

attrition:
### Rationale
I find myself switching between the empire and the trade advisor screen often, sometimes forgetting what I had meant to check in either screen on the way. So I built this PR to make the trade resources on the empire screen become buttons that launch the resources setting dialog.

### Preview
julius-resource-buttons-short

### Impurity
I believe this change does not conflict with Julius' mission to maintain a 1:1 gameplay+save/load experience with C3 as it's just a UI change. That being said, it is a fair departure from the original work. 

### Drawbacks
1) At the time of submission, there is no option to control this feature. Due to the scope of the changes (the layout, drawing and input handling of the buttons), the change is not simple to gate behind a setting. It is possible to do, but would require a fair amount of copy/paste to handle the old and new modes.

~~2) Literally drawing the background. When working on this I noticed a bug in Julius not present in C3, which is bits of the help menu remaining on screen when closing the help popup from the resource setting dialog box:~~
julius-bg-bug

~~Of course I had the same issue from the empire screen, the help dialog would remain artifacted behind the resource setting window, on top of the empire map.~~

~~There were two approaches I considered for this and I didn't really like either of them, so I just picked one:~~
~~- When creating the resource dialog, you may pass in a background drawing function, as the resource window now can't guarantee the advisor window is in the background to draw that~~
~~- If NULL is passed, the existing logic of drawing the advisor background is taken via the `draw_background` function~~
~~- But in this case we also draw the full advisors background and not just the trade advisor portion, to fix the help menu bug above~~

~~One alternative approach would have been for the resource setting menu to not provide any default `draw_background` function and require the advisor screen and empire screen to pass in their callbacks. This removes the NULL sent by the trade advisor screen but also means setting up a bespoke background drawing function of its own to pass as a callback.~~

~~I have implemented the first method but will use whichever method is deemed best.~~

Cleaned up now, thanks to crudelios

3) There is one button for each resource type, even if such resource is not present in the selected city. We check at the input handling stage if a resource button needs to be handled. It is possible to instead have a smaller cap of buttons and then modify their parameters to represent just the resources for a given city, but there isn't much to be gained by the complexity.

### DIFF
--- a/src/window/advisor/trade.c
+++ b/src/window/advisor/trade.c
@@ -113,7 +113,7 @@ static void button_empire(int param1, int param2)
 
 static void button_resource(int resource_index, int param2)
 {
-    window_resource_settings_show(city_resource_get_available()->items[resource_index], 0 /*NULL*/);
+    window_resource_settings_show(city_resource_get_available()->items[resource_index]);
 }
 
 static int get_tooltip_text(void)
--- a/src/window/advisor/trade.c
+++ b/src/window/advisor/trade.c
@@ -113,7 +113,7 @@ static void button_empire(int param1, int param2)
 
 static void button_resource(int resource_index, int param2)
 {
-    window_resource_settings_show(city_resource_get_available()->items[resource_index]);
+    window_resource_settings_show(city_resource_get_available()->items[resource_index], NULL);
 }
 
 static int get_tooltip_text(void)
--- a/src/window/advisor/trade.c
+++ b/src/window/advisor/trade.c
@@ -113,7 +113,7 @@ static void button_empire(int param1, int param2)
 
 static void button_resource(int resource_index, int param2)
 {
-    window_resource_settings_show(city_resource_get_available()->items[resource_index], NULL);
+    window_resource_settings_show(city_resource_get_available()->items[resource_index], 0 /*NULL*/);
 }
 
 static int get_tooltip_text(void)
--- a/src/window/empire.c
+++ b/src/window/empire.c
@@ -65,7 +65,7 @@ static generic_button generic_button_trade_resource[] = {
     {0, 0, 101, 27, button_show_resource_window, button_none, RESOURCE_POTTERY, 0}
 };
 static generic_button generic_button_open_trade[] = {
-    {30, 61, 440, 26, button_open_trade, button_none, 0, 0}
+    {30, 56, 440, 26, button_open_trade, button_none, 0, 0}
 };
 
 static struct {
@@ -552,7 +552,9 @@ static int is_mouse_hit(tooltip_context *c, int x, int y, int size)
 static int get_tooltip_resource(tooltip_context *c)
 {
     const empire_city *city = empire_city_get(data.selected_city);
-    if (city->type != EMPIRE_CITY_TRADE) {
+    // we only want to check tooltips on our own closed cities.
+    // open city resource tooltips are handled by their respective buttons directly
+    if (city->type != EMPIRE_CITY_TRADE || city->is_open) {
         return 0;
     }
     int object_id = empire_selected_object() - 1;
--- a/src/window/empire.c
+++ b/src/window/empire.c
@@ -542,16 +542,16 @@ static void handle_input(const mouse *m, const hotkeys *h)
     }
 }
 
-static int is_mouse_hit(tooltip_context* c, int x, int y, int size)
+static int is_mouse_hit(tooltip_context *c, int x, int y, int size)
 {
     int mx = c->mouse_x;
     int my = c->mouse_y;
     return x <= mx && mx < x + size && y <= my && my < y + size;
 }
 
-static int get_tooltip_resource(tooltip_context* c)
+static int get_tooltip_resource(tooltip_context *c)
 {
-    const empire_city* city = empire_city_get(data.selected_city);
+    const empire_city *city = empire_city_get(data.selected_city);
     if (city->type != EMPIRE_CITY_TRADE) {
         return 0;
     }
--- a/src/window/empire.c
+++ b/src/window/empire.c
@@ -26,6 +26,7 @@
 #include "window/city.h"
 #include "window/message_dialog.h"
 #include "window/popup_dialog.h"
+#include "window/resource_settings.h"
 #include "window/trade_opened.h"
 
 #define MAX_WIDTH 2032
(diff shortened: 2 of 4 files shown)


src/window/advisor/trade.c
```diff
@@ -113,7 +113,7 @@ static void button_empire(int param1, int param2)
 
 static void button_resource(int resource_index, int param2)
 {
-    window_resource_settings_show(city_resource_get_available()->items[resource_index]);
+    window_resource_settings_show(city_resource_get_available()->items[resource_index], NULL);
```
attrition, line 116: See `Drawbacks #2`

src/window/empire.c
```diff
@@ -45,6 +47,23 @@ static image_button image_button_return_to_city[] = {
...
+    {0, 0, 27, 27, button_show_resource_window, button_none, 7, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 8, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 9, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 10, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 11, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 12, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 13, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 14, 0},
+    {0, 0, 27, 27, button_show_resource_window, button_none, 15, 0}
+};
```
attrition, line 66: See `Drawbacks #3`

src/window/empire.c
```diff
@@ -128,45 +153,45 @@ static void draw_trade_resource(resource_type resource, int trade_max, int x_off
 static void draw_trade_city_info(const empire_object *object, const empire_city *city)
```
attrition, line 153: Minor layout changes were required to fit buttons vs the old resource icons. However the changes touched a large number of lines. See `Drawbacks #1`

src/window/empire.c
```diff
@@ -495,56 +538,6 @@ static int is_mouse_hit(tooltip_context *c, int x, int y, int size)
     return x <= mx && mx < x + size && y <= my && my < y + size;
 }
 
-static int get_tooltip_resource(tooltip_context *c)
```
attrition, line 498: Removing this code was sort of a "win", complexity-wise. The `generic_button` input handling knows which resource button was focused, if any. The tooltip code can then retrieve the focused resource type and display that text directly.

src/window/resource_settings.c
```diff
@@ -180,7 +185,7 @@ static void handle_input(const mouse *m, const hotkeys *h)
         return;
     }
     if (input_go_back_requested(m, h)) {
-        window_advisors_show();
+        window_go_back();
```
attrition, line 188: We can use the more generic `go-back` process instead of showing the advisor dialog. This seems to work perfectly fine in my testing and allows me to return to the empire screen instead

src/window/resource_settings.c
```diff
@@ -216,11 +221,11 @@ static void button_toggle_stockpile(int param1, int param2)
     city_resource_toggle_stockpiled(data.resource);
 }
 
-void window_resource_settings_show(resource_type resource)
+void window_resource_settings_show(resource_type resource, void (*override_draw_background_fn)(void))
 {
     window_type window = {
         WINDOW_RESOURCE_SETTINGS,
-        draw_background,
+        (override_draw_background_fn ? override_draw_background_fn : draw_background),
```
attrition, line 228: This is ugly

src/window/advisor/trade.c
```diff
@@ -113,7 +113,7 @@ static void button_empire(int param1, int param2)
 
 static void button_resource(int resource_index, int param2)
 {
-    window_resource_settings_show(city_resource_get_available()->items[resource_index]);
+    window_resource_settings_show(city_resource_get_available()->items[resource_index], 0 /*NULL*/);
```
attrition, line 116: See `Drawbacks #2`

src/window/resource_settings.c
```diff
@@ -55,6 +55,11 @@ static void init(resource_type resource)
 
 static void draw_background(void)
 {
+    // added this draw call to fix a bug where the Help window is larger than the
+    // trade dialog itself, causing bits of it to remain on the screen when going
+    // back to the dialog. Bug was not present in original C3.
+    window_advisors_draw_dialog_background();
```
crudelios, line 61: Unsing `window_draw_underlying_window` here should fix your background issues without requiring hacky callbacks.

src/window/resource_settings.c
crudelios: Yeah, this is the correct way to do it :wink:

src/window/empire.c
```diff
@@ -413,6 +437,7 @@ static void draw_foreground(void)
     }
     draw_city_name(city);
     draw_panel_buttons(city);
+    draw_object_info();
```
attrition, line 440: This was in draw_background, but with the new button handling it needed to be handled on the foreground

attrition:
Updated to cover all feedback. The original intent of having the buttons clickable prior to opening a trade is to see if I am importing or exporting that resource already to another city, but it isn't high value so I removed it as requested.

src/window/empire.c
attrition: This is back again to handle the tooltips for a closed city only.

src/window/empire.c
```diff
@@ -488,60 +542,42 @@ static void handle_input(const mouse *m, const hotkeys *h)
...
-static int is_mouse_hit(tooltip_context *c, int x, int y, int size)
+static int is_mouse_hit(tooltip_context* c, int x, int y, int size)
 {
     int mx = c->mouse_x;
     int my = c->mouse_y;
     return x <= mx && mx < x + size && y <= my && my < y + size;
 }
 
-static int get_tooltip_resource(tooltip_context *c)
+static int get_tooltip_resource(tooltip_context* c)
```
crudelios, line 552:
```suggestion
static int get_tooltip_resource(tooltip_context *c)
```

Move the `*` to the right place, please.

src/window/empire.c
```diff
@@ -488,60 +542,42 @@ static void handle_input(const mouse *m, const hotkeys *h)
...
 {
-    const empire_city *city = empire_city_get(data.selected_city);
+    const empire_city* city = empire_city_get(data.selected_city);
```
crudelios, line 554:
```suggestion
    const empire_city *city = empire_city_get(data.selected_city);
```

Same.

attrition:
oops disregard last push

attrition:
Thank you!

PR #514, opened 2020-09-05
Android build

bvschaik:
Continuation of #380


src/game/file.c
```diff
@@ -227,6 +227,7 @@ static void initialize_saved_game(void)
 
     sound_city_init();
 
+    building_construction_clear_type();
```
bvschaik, line 230: Why do we need this?

crudelios: Because android keeps some things in memory on exit, if that code isn't there the last used construction will still be active upon restarting the game.

src/platform/keyboard_input.c
```diff
@@ -262,6 +264,20 @@ void platform_handle_key_down(SDL_KeyboardEvent *event)
         case SDLK_END:
             keyboard_end();
             break;
+        case SDLK_AC_BACK:
+            event->keysym.scancode = SDL_SCANCODE_ESCAPE;
+            break;
+#ifdef __ANDROID__
```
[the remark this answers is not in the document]
crudelios: Well, though it hasn't worked for the user, it did work for my phone. But if you don't like the hack it can be removed.

src/platform/virtual_keyboard.c
```diff
@@ -1,5 +1,6 @@
 #include "game/system.h"
 
+#include "android/android.h"
```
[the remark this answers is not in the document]
crudelios: The include is needed because it defines `PLATFORM_USE_VIRTUAL_KEYBOARD`, which is then used by this file.

src/platform/julius.c
```diff
@@ -34,10 +34,18 @@
 #include "platform/vita/vita_input.h"
 #endif
 
+#ifdef __ANDROID__
```
bvschaik, line 37: Include guard is not necessary, since the header also does the check.

src/platform/keyboard_input.c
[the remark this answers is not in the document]
bvschaik: In my opinion it's a little too hacky.

src/platform/virtual_keyboard.c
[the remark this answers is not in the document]
bvschaik: Yes, please add a comment. If someone decides to clean up imports this will break.

src/game/file.c
bvschaik: Isn't that also a problem when loading scenarios then?

crudelios: If I remember well this code works for both cities and scenarios.

bvschaik: Only for campaign scenarios (which are basically saved games from mission1.pak), custom scenarios don't call this function when starting.

crudelios: Yeah, I knew it did work but it was a while ago that I fixed it so I didn't really remember it too well. Thanks for checking 👍

PR #516, opened 2020-10-16
Military sidebar take 2

### LINKED ISSUE #320: Maybe more visibility when controlling legions?

bvschaik:
Continuation from Crudelios' work in #378.

Fixes #320


crudelios:
Tested, everything seems to be working, only a small spacing issue:

nospacing

bvschaik:
Oh, good one. Additionally, I'd like the military sidebar texts to start at the same X offset as the ones for the extra info below it. Right now the military text starts a little too close to the border to my case. I'll fix those as well.

And I just found another issue: collapse the sidebar, select a legion, watch the panel slide in view. Then, right-click and the panel slides out of view. While it's doing so, the contents revert to the "0 soldiers" state.

crudelios:
> And I just found another issue: collapse the sidebar, select a legion, watch the panel slide in view. Then, right-click and the panel slides out of view. While it's doing so, the contents revert to the "0 soldiers" state.

Good one! At least it doesn't crash, as it probably loses the selected ID reference.

crudelios:
The code itself looks good, but there are two issues:

- The arrows to cycle legions disappear when the sidebar is expanding/collapsing when a military unit is selected.
- The german translation is still clipping for "Soldier's health", do we leave it like this or do we eclipsize the text?

bvschaik:
I didn't work on the German yet. I want to try ellipsize, but if that looks bad I will leave it like this.

The arrow buttons: that was already the case. I'm impartial to this. The regular sidebar also hides some elements when sliding.

crudelios:
> I didn't work on the German yet. I want to try ellipsize, but if that looks bad I will leave it like this.

Brazilian will also need it because somehow the translators though that adding plenty of `++++` to the health info would be a good idea (spoiler: it isn't).

> The arrow buttons: that was already the case. I'm impartial to this.

It was? I failed to notice it then somehow. My OCD doesn't like them disappearing actually.

> The regular sidebar also hides some elements when sliding.

Not anymore, I fixed that when I changed the sliding code.

crudelios:
Working fine, LGTM 👍

PR #518, opened 2020-10-16
Update german.c

m-mohr:
German translations as requested by #415. Also includes the translation for #517,

Didn't check whether the translation exceeds the space on the screen, but tried to be short.


bvschaik:
All strings fit perfectly, thanks!

PR #521, opened 2020-10-21
Ellipsize the config text when it doesn't fit

crudelios:
Also make a tooltip appear over the ellipsized text to expand it

(Also fixed a warning because it was nagging me)

Let me know what you think about the tooltip, I think it looks good but it made me implement a bunch of code changes.

### DIFF
--- a/src/graphics/tooltip.c
+++ b/src/graphics/tooltip.c
@@ -290,23 +290,6 @@ static void draw_senate_tooltip(tooltip_context *c)
     }
 }
 
-static void draw_config_checkbox_tooltip(tooltip_context *c)
-{
-    if (!c->text) {
-        return;
-    }
-    int x = screen_dialog_offset_x() + 50;
-    int y = c->mouse_y - ((c->mouse_y - screen_dialog_offset_y() - 60) % 24);
-    int width = text_get_width(c->text, FONT_NORMAL_PLAIN) + 10;
-    int height = 20;
-
-    save_window_under_tooltip_to_buffer(x, y, width, height);
-
-    graphics_draw_rect(x, y, width, height, COLOR_BLACK);
-    graphics_fill_rect(x + 1, y + 1, width - 2, height - 2, COLOR_WHITE);
-    text_draw(c->text, x + 5, y + 5, FONT_NORMAL_PLAIN, 0);
-}
-
 static void draw_tooltip(tooltip_context *c)
 {
     if (c->type == TOOLTIP_BUTTON) {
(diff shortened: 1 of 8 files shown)


bvschaik:
Regarding the tooltip: I see this ellipsizing as a temporary measure until all strings have been updated to fit the available space, so the tooltip is not necessary in my opinion.

crudelios:
I agree it's temporary, but since we don't know how long it will take for everyrhing to be translated, what do you think about keeping the tooltip temporarily? Then when we get smaller translations for everything we revert the code.

bvschaik:
I dislike that. I've seen "temporary" changes being kept around for years...
Having them ellipsized without tooltip will also be an incentive for people to provide updated translations for their language.

crudelios:
Makes sense. I'll get rid of it then.

src/graphics/tooltip.c
```diff
@@ -312,7 +312,7 @@ void tooltip_handle(const mouse *m, void (*func)(tooltip_context *))
         reset_timer();
         return;
     }
-    tooltip_context context = {m->x, m->y, 0, 0, 0, 0, 0, 0};
+    tooltip_context context = {m->x, m->y, 0, 0, 0, 0, 0, 0, 0, 0};
```
bvschaik, line 315: Do we still need the extra zeroes here?
(aside: do we need the zeroes at all? I seem to remember that when you initialize at least one variable of a struct, the rest of it is set to 0 automatically)

crudelios: > (aside: do we need the zeroes at all? I seem to remember that when you initialize at least one variable of a struct, the rest of it is set to 0 automatically)

Answer: Indeed, no.

From [Stack Overflow](https://stackoverflow.com/questions/37548287/partially-initializing-a-c-struct):

> This is guaranteed as per section 6.7.8 of the C standard (with highlighting specific to this case):
> 
>
> > 10 If an object that has automatic storage duration is not initialized explicitly, its value is indeterminate. If an object that has static storage duration is not initialized explicitly, then:
> > 
> > —if it has pointer type, it is initialized to a null pointer;
> >
> > —if it has arithmetic type, it is initialized to (positive or unsigned) zero;
> > 
> > —if it is an aggregate, every member is initialized (recursively) according to these rules;
> > 
> > —if it is a union, the first named member is initialized (recursively) according to these rules.
> > 
> > ...
> > 
> > 21 If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration.

I'll get rid of those zeroes.

PR #529, opened 2020-11-12
Joystick support

crudelios:
This PR starts implementing joystick support.

Right now joysticks are not mappable, but the implementation is there to allow mapping in the future.

The new code allows for much better integration of both Vita and Switch code. Specifically, the virtual keyboard integration should be improved and Switch should have multi-gamepad support (untested).

The idea in the future is to integrate this code with the SDL Game Controller API, allowing for seamless integration of a very high number of controllers.

The code was tested on my own Vita, and on yuzu emulator for Switch and seems to be working properly.

The reason I'm releasing it as-is is not only because it's already a big change, but also because it reduces Vita and Switch-specific code.


src/game/system.h
```diff
@@ -98,7 +98,7 @@ void system_keyboard_set_input_rect(int x, int y, int width, int height);
  * @param text The text to display on the virtual keyboard
  * @param max_length The maximum length of the text
  */
-void system_keyboard_show(const uint8_t *text, int max_length);
+void system_keyboard_show(uint8_t *text, int max_length);
```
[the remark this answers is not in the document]
crudelios: Because we directly change the `text` variable on Vita and Switch with the result of the virtual keyboard window's final text

src/widget/input_box.c
```diff
@@ -5,8 +5,11 @@
 #include "graphics/text.h"
 #include "input/keyboard.h"
 
+static input_box *active_box;
+
 void input_box_start(input_box *box, uint8_t *text, int length, int allow_punctuation)
 {
+    active_box = box;
```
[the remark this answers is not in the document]
crudelios: Humm. I do need to find out if an input box is being used or not.

What do you think about making `input_box.c` own all input boxes (it would be a `static` variable) instead of each window having its own `input_box` (current implementation)?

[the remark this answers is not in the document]
bvschaik: I migrated the code away from that so I'd not like going back 😉 The input boxes used to be lumped together but that was hard to maintain. I see the keyboard code is also not "safe" with respect to the `text` variable. I'll have a look at the whole setup to make it simpler and less error-prone.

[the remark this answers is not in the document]
crudelios: Thank you, that would be appreciated. My idea was to have Switch and Vita's virtual keyboard controlling each input box's text, and also only allowing the virtual keyboard to appear when there is an active input box.

crudelios:
Since this PR needs some rework, I'm closing it for now.

crudelios:
This is now ready for review again

src/input/joystick.c
```diff
@@ -0,0 +1,714 @@
...
+
+    input->value = 0;
+    input->element = JOYSTICK_ELEMENT_NONE;
+
+    for (int i = 0; i < JOYSTICK_MAX_CONTROLLERS; ++i) {
+        const joystick_info *joystick = &data.joystick[i];
+        if (!joystick->connected) {
+            continue;
+        }
+        if(get_input_for_mapping(joystick, &joystick->model->first_mapping[action], input) ||
```
bvschaik, line 356: Space after `if` 😉

src/input/joystick.c
```diff
@@ -0,0 +1,714 @@
...
+typedef int joystick_axis;
+typedef int joystick_button;
+
+enum {
+    DIRECTION_UP = 0,
+    DIRECTION_LEFT = 1,
+    DIRECTION_DOWN = 2,
+    DIRECTION_RIGHT = 3,
+    NUM_DIRECTIONS = 4
+};
```
bvschaik, line 26: Any reason why this is not a `typedef enum { ...} joystick_button`?

src/input/joystick.h
```diff
@@ -0,0 +1,101 @@
...
+    JOYSTICK_HAT_DOWN = 4,
+    JOYSTICK_HAT_RIGHT = 8
+} joystick_hat_position;
+
+typedef enum {
+    JOYSTICK_TRACKBALL_X_POSITIVE = 0,
+    JOYSTICK_TRACKBALL_X_NEGATIVE = 1,
+    JOYSTICK_TRACKBALL_Y_POSITIVE = 2,
+    JOYSTICK_TRACKBALL_Y_NEGATIVE = 3,
+} joystick_trackball_position;
```
bvschaik, line 42: These constants are not used outside the `.c` file, and the enum type is not used anywhere.

src/input/joystick.h
```diff
@@ -0,0 +1,101 @@
...
+} mapping_element;
+
+typedef struct {
+    int connected_joysticks;
+    char name[JOYSTICK_MAX_NAME];
+    char guid[JOYSTICK_MAX_GUID];
+    int total_axis;
+    int total_buttons;
+    int total_trackballs;
+    int total_hats;
```
bvschaik, line 86: These 4 variables are only set and never used.

src/input/joystick.h
```diff
@@ -0,0 +1,101 @@
...
+    mapping_element first_mapping[MAPPING_ACTION_MAX];
+    mapping_element second_mapping[MAPPING_ACTION_MAX];
```
bvschaik, line 88: What's the meaning of `first_mapping` and `second_mapping`? Is it like the hotkeys with alternative keys or is it something else? Is there a compelling reason not to use a single mapping variable?

src/input/joystick.h
```diff
@@ -0,0 +1,101 @@
...
+} joystick_model;
```
bvschaik, line 89: Ideally I'd not expose this struct in the header, since it's only used internally in the corresponding `joystick.c`. See also my comments on the next two functions.

src/input/joystick.c
```diff
@@ -0,0 +1,714 @@
...
+    hat->right = (position & JOYSTICK_HAT_RIGHT) ? 1 : 0;
+}
+
+static void update_trackball(joystick_trackball *trackball, int delta_x, int delta_y)
+{
+    trackball->delta_x += delta_x;
+    trackball->delta_y += delta_y;
+}
+
+static void joystick_reset_all_button_states(joystick_info *joystick)
```
bvschaik, line 136: Preferably don't prefix `static` functions with the name of the file (`joystick_`): that way it's easier to see from the name whether the function is public or not.
Also, I think these one-liner functions can be inlined in `joystick_reset_state`.

src/input/joystick.c
```diff
@@ -0,0 +1,714 @@
...
+            update_hat(&joystick->hat[element_id], value1);
+            break;
+        case JOYSTICK_ELEMENT_AXIS:
+            joystick->axis[element_id] = (abs(value1) > 1300) ? value1 : 0;
+            break;
+        case JOYSTICK_ELEMENT_TRACKBALL:
+            update_trackball(&joystick->trackball[element_id], value1, value2);
+            break;
+        default:
+            log_info("Trying to update wrong joystick element", 0, element);
```
bvschaik, line 237: Also end this case with a `break;`

src/input/joystick.c
```diff
@@ -0,0 +1,714 @@
...
+{
+    for (int i = 0; i < JOYSTICK_MAX_CONTROLLERS; ++i) {
+        if (!data.joystick[i].connected) {
+            return &data.joystick[i];
+        }
+    }
+    return 0;
+}
+
+static joystick_info *get_joystick_from_id(int joystick_id)
```
bvschaik, line 174: Suggested name: `get_joystick_by_id()`

src/input/joystick.c
```diff
@@ -0,0 +1,714 @@
...
+        }
+    } else if (translated_element == JOYSTICK_ELEMENT_BUTTON || translated_element == JOYSTICK_ELEMENT_HAT) {
+        input->value = (input->value != 0) ? 1 : 0;
+    }
+    input->element = translated_element;
+}
+
+static int rescale_axis(mapped_input *inputs)
+{
+    //radial and scaled dead_zone
```
bvschaik, line 396: One space after `//` please. Also format like normal sentences, with capital letter at the start.

src/platform/joystick.c
```diff
@@ -0,0 +1,231 @@
...
+        return model;
+    }
+    model = joystick_get_first_unused_model();
+    if (!model) {
+        return 0;
+    }
+    snprintf(model->guid, JOYSTICK_MAX_GUID, "%s", guid);
+    const char *name = SDL_JoystickName(joystick);
+    if (!name) {
+        snprintf(model->name, JOYSTICK_MAX_NAME, "Joysyick %d", instance_id);
```
bvschaik, line 123: `Joysyick`? 😂

bvschaik (changes requested):

src/input/joystick.h
crudelios: Exactly. Is for the cases where there may be more than a single buttom for the same action. Both vita and switch have some repeated actions to different buttons, so I dod it this way.

bvschaik: Can we put them in just one array? That's also how the regular hotkeys are implemented: even though an action might have two hotkeys assigned to it (or even more if you edit the config file by hand), it's just one list in memory, which makes the code for handling hotkeys easier. Only on the config screen are they put in a struct with "main" and "alternative" hotkey.

src/input/joystick.c
crudelios: Because the button is independent from the directions. The button can be an `A` button, a `X` button, etc... Directions are used in entirely different places, mainly to calculate the axis direction.

src/input/joystick.h
crudelios: These may be needed later, but for now I'm removing them.

src/input/joystick.c
crudelios: Sure, but why is there a need to end the last case with a `break;`?

bvschaik: For consistency's sake and defensive programming. See also: [here](https://softwareengineering.stackexchange.com/a/201786) and [here](https://stackoverflow.com/a/26139061).

crudelios: Makes sense! Thanks for the explanation

src/input/joystick.h
crudelios: I need to expose this in the header because both `platform/joystick.c` and `input/joystick.c` need full access to `joystick_model`.

`platform` because it fills the `model` with the values from SDL (and also from mapping, which could later be also obtained via SDL), and `input` needs to access the model information, mainly for mapping purposes.

src/input/joystick.c
```diff
@@ -0,0 +1,720 @@
...
+static void translate_input_for_element(mapped_input *input, joystick_element translated_element)
+{
+    if (input->element == translated_element) {
+        return;
+    }
+    if (translated_element == JOYSTICK_ELEMENT_AXIS) {
+        switch(input->element)
+        {
+            case JOYSTICK_ELEMENT_TRACKBALL:
+                input->value *= 5; // TODO is this a high enough value? I doubt it
```
bvschaik, line 378: Do we need to resolve this TODO before merging?

crudelios: No, at least not until sometime has an actual trackball to test!

src/input/joystick.c
```diff
@@ -0,0 +1,720 @@
...
+    return 0;
+}
+
+static void translate_input_for_element(mapped_input *input, joystick_element translated_element)
+{
+    if (input->element == translated_element) {
+        return;
+    }
+    if (translated_element == JOYSTICK_ELEMENT_AXIS) {
+        switch(input->element)
```
bvschaik, line 375: Found a last-minute formatting inconsistency: `switch(` without space before the `(`

src/input/joystick.c
bvschaik: Hmm, ok. Can we at least move the `5` to a constant and add a comment there instead? I'm hesitant to merge code with TODOs in them.

crudelios: Sure!

If it helps your OCD, remember this is just phase one - we can't test trackballs anyway, because there's no mapping support for custom controllers!

And we'll need someone with an actual trackball to see if the ratio of `5` is under or oversensitive.

bvschaik (approved):

PR #530, opened 2020-11-13
Add scale options to config screen

### LINKED ISSUE #520: Add scaling options to the config window
Is there a way to change to another full screen resolution?

I appreciate that julius supports 1920x1080 resolution, but I think either a zoom feature or a possibility to select another full-screen solution like 1280x720 or 1366x768 would be very good for our eyes :)

bvschaik:
This PR adds support for changing the display and cursor scale on the fly on the Julius config screen, which were previously only available as command line arguments.

Display scale and cursor scale options

These options remain available as command line options; specifying them on the command line overwrites the values set in the config screen.

Fixes #520


src/platform/screen.c
```diff
@@ -38,27 +38,47 @@ static int scale_pixels_to_logical(int pixel_value)
...
+
+    if (!pixel_width || !pixel_height) {
+        return;
+    }
+
     // Assure width is at least 640 and height is at least 480
     float width_reference = scale_pixels_to_logical(pixel_width) / (float) MINIMUM.WIDTH;
     float height_reference = scale_pixels_to_logical(pixel_height) / (float) MINIMUM.HEIGHT;
     float minimum_reference = SDL_min(width_reference, height_reference);
     if (minimum_reference < 1.0f) {
```
crudelios, line 60: I think you can use `get_max_scale_percentage` here now, right?

crudelios (commented):
I checked the code (looks good, I only have one suggestion) and tested it as well. It's working just fine, nice job! 👍

src/platform/screen.c
bvschaik: Yep, fixed.

PR #538, opened 2020-11-17
Stop playing city and speech sounds when the game is paused

### LINKED ISSUE #537: Stop playing city and speech sounds when the game is paused
When the game is paused, time basically stops, and it makes no sense that the city still produces sounds and citizens still speak when clicked. Only the sound effects and the music should be playing in pause state.

(It would be even better to stop their sprite animations, but that's another issue.)

gergelypap:
Fixes #537

### DIFF
--- a/src/game/game.c
+++ b/src/game/game.c
@@ -24,6 +24,7 @@
 #include "scenario/property.h"
 #include "scenario/scenario.h"
 #include "sound/city.h"
+#include "sound/speech.h"
 #include "sound/system.h"
 #include "translation/translation.h"
 #include "window/editor/map.h"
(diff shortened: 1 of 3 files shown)


src/game/game.c
```diff
@@ -182,6 +183,10 @@ void game_run(void)
 void game_draw(void)
 {
     window_draw(0);
+    if (game_state_is_paused()) {
+        sound_city_set_volume(0);
+        sound_speech_set_volume(0);
+    }
```
crudelios, line 189: Regardless of actually agreeing or disagreeing to the proposed change, this code will turn off the volume until you manually restore it, even after you unpause the game, so you can remove it.

gergelypap: I have tested, and the sounds came back after unpausing.

bvschaik:
Closing this PR, see #537 for details.

PR #539, opened 2020-11-18
Merge screen and cursor code for Vita and Switch

crudelios:
This is the last step necessary for VIta/Switch integration

The screen code is centralized on `platform/screen.c` for all platforms. Nearly all the code is shared now.

This PR also has some new features:

Switch now changes resolution between docked and in handheld mode. More specifically, it uses the scaling to set a custom resolution of 960x540 on handheld mode and 1280x720 in docked mode (full hd seemed too much resolution for a TV screen).

Implemented a "software mode" for the mouse cursor: since Vita and Switch don't have a mouse cursor (rather, the cursor is drawn as a texture), I went ahead and created a general purpose code that draws the cursor as a texture. Right now that code is hardcoded to only be used on Vita and Switch, but it works for any system (for example, we can add a command line option)

Note: this code will have merge conflicts with #529. Whichever is merged first, the other one will have to be rebased.


src/game/system.h
```diff
@@ -125,6 +127,24 @@ void system_mouse_get_relative_state(int *x, int *y);
  */
 void system_move_mouse_cursor(int delta_x, int delta_y);
 
+/**
+ * Tells whether to use the software or hardware mouse cursor
+ * @return True when the software cursor should be used, false otherwise
+ */
+int system_use_software_cursor(void);
```
bvschaik, line 134: Only use `game/system.h` for code outside `platform` that needs to call something that's implemented in `platform`, to keep the integration layer as thin as possible. These three new functions are only used inside`platform` code, so move them to a private header there.

src/graphics/graphics.c
```diff
@@ -24,22 +22,13 @@ static struct {
 static struct {
     int x;
     int y;
-} translation = {0, 0};
+} translation;
 
 static clip_info clip;
 
-#ifdef __vita__
```
bvschaik, line 31: Nice! No more platform checks in the main Julius code.

bvschaik (changes requested):

src/graphics/graphics.c
crudelios: Yes, precisely one of the reasons I've been wanting to do this for a long time...

src/game/system.h
crudelios: Well, live and learn! There are still some coding standards intrincancies I never noticed 😉

bvschaik (approved):

PR #543, opened 2020-11-19
Add directory structure cache for Vita and Switch

crudelios:
Improves performance when handling many files (opening the options menu went from taking 30 seconds to about 1 second)

### DIFF
--- a/src/platform/file_manager.c
+++ b/src/platform/file_manager.c
@@ -4,6 +4,7 @@
 #include "core/log.h"
 #include "core/string.h"
 #include "platform/android/android.h"
+#include "platform/file_manager_cache.h"
 #include "platform/vita/vita.h"
 
 #include <dirent.h>
@@ -20,7 +21,6 @@
 #define fs_dir_close _wclosedir
 #define fs_dir_read _wreaddir
 #define dir_entry_name(d) wchar_to_utf8(d->d_name)
-#define dir_to_utf8(d) wchar_to_utf8(d)
 typedef const wchar_t * dir_name;
 
 static const char *wchar_to_utf8(const wchar_t *str)
@@ -52,7 +52,6 @@ static wchar_t *utf8_to_wchar(const char *str)
 #define fs_dir_close closedir
 #define fs_dir_read readdir
 #define dir_entry_name(d) ((d)->d_name)
-#define dir_to_utf8(d) (d)
 typedef const char * dir_name;
 #endif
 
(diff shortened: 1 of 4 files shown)


crudelios:
I think there are some improvements left: separating the directory and file cache structs may improve performance when opening the Options menu.

I'll have a look later today.

crudelios:
Finished the optimizations.

Stat was not working, meaning every file was always being checked, even when we only wanted to check for dirs. This made opening the `Options` menu reeeeally slow.

So for Vita, we don't use stat anymore, rather we try to open the files as a directory. To improve performance, only files that don't have an extension are checked to see if they are directories.

src/platform/file_manager.c
```diff
@@ -83,10 +85,159 @@ typedef const char * dir_name;
 #include <unistd.h>
 #endif
 
+#ifdef __vita__
```
bvschaik, line 88: Can we move the caching parts to a separate file - `file_manager_cache.c` perhaps? It's quite a lot of code, only necessary for vita, and it's separate functionality ([single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)).

bvschaik (changes requested):

src/platform/file_manager.c
crudelios: I was thinking whether I should do that actually. Guess I'll do it then :wink:

crudelios:
Changes applied.

Also added the cache to Switch. Turns out adding the cache to switch makes both booting the game and loading the options screen about twice as fast (at least on the yuzu emulator)

src/platform/file_manager_cache.h
```diff
@@ -0,0 +1,28 @@
+#ifndef FILE_MANAGER_CACHE_H
+#define FILE_MANAGER_CACHE_H
+
+#if defined(__vita__) || defined(__SWITCH__)
```
bvschaik, line 4: Suggestion: create define for USE_FILE_CACHE:
```suggestion
#if defined(__vita__) || defined(__SWITCH__)
#define USE_FILE_CACHE
#endif
```
In the other places you can then just use `#ifdef USE_FILE_CACHE` instead of repeating the vita/switch define checks. This makes it easier to add caching to other platforms if that's ever necessary.

src/platform/file_manager_cache.c
```diff
@@ -0,0 +1,166 @@
...
+static int stat_status;
+
+const dir_info *platform_file_manager_cache_get_dir_info(const char *dir)
+{
+    dir_info *info = base_dir_info;
+    while (info) {
+        if (strcmp(info->name, dir) == 0) {
+            return info;
+        }
+        if (!info->next) {
```
bvschaik, line 29: This check is not necessary: `info = info->next` will set info to null, and the next while loop evaluation will stop the loop.

src/platform/file_manager_cache.c
```diff
@@ -0,0 +1,166 @@
...
+        if (stat_status == STAT_WORKS) {
+            stat(file_item->name, &file_info);
+            if (!S_ISREG(file_info.st_mode)) {
+                type = TYPE_DIR;
+            }
+        } else {
+            // When stat does not work, we check if a file is a directory by trying to open it as a dir
+            // For performance reasons, we only check for a directory if the name has no extension
+            // This is effectively a hack, and definitely not full-proof, but the performance gains are well worth it
+            if (!c) {
```
bvschaik, line 96: Reuse of a one-letter variable... whose intent is not clear from just the code. Can we replace this with `!*file_item->extension` or something?

src/platform/file_manager.c
```diff
@@ -103,7 +106,25 @@ int platform_file_manager_list_directory_contents(
...
 #ifdef __ANDROID__
-    return android_get_directory_contents(current_dir, type, extension, callback);
+    int match = android_get_directory_contents(current_dir, type, extension, callback);
+#elif defined(__vita__) || defined(__SWITCH__)
+    const dir_info *d = platform_file_manager_cache_get_dir_info(current_dir);
+    if (!d) {
+        return LIST_ERROR;
+    }
+    int match = LIST_NO_MATCH;
+    for(file_info *f = d->first_file; f; f = f->next) {
```
bvschaik, line 116: Space after `for`

bvschaik (changes requested):

src/platform/file_manager_cache.c
crudelios: Forgotten remnants of past code...

bvschaik (approved):

PR #545, opened 2020-11-20
Show bridge cost on hover

### LINKED ISSUE #541: Building bridges should show the cost
When building roads or reservoirs, dragging the mouse shows the cost of building:

image

image

This is not the case with bridges:

image

Bridges are quite expensive, taking 100 denarii per tile for a Ship Bridge, and 40 denarii for a Low Bridge, so I think it would be nice to see how much would it cost. The bridge in the picture costs 1400 for example.

**Edit:** I am now seeing that the cost is only shown when the mouse click is pressed and hold, for every building. I'm guessing this is the reason bridges don't show the cost, because they are not "draggable". However, I tend to now favor showing costs for all buildings even when just hovering the mouse and not clicking. Of course, I can "drag away" to see the cost and not build the bridge, but it is not really intuitive. 

What do you think?

crudelios:
Fix #541

### DIFF
--- a/src/building/construction.c
+++ b/src/building/construction.c
@@ -263,6 +263,12 @@ static int place_reservoir_and_aqueducts(
     return 1;
 }
 
+void building_construction_set_cost(int cost)
+{
+    data.in_progress = cost != 0;
+    data.cost = cost;
+}
+
 void building_construction_set_type(building_type type)
 {
     data.type = type;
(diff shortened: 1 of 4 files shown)


bvschaik:
Bug found:

1. Let the game run
1. Select bridge from the build menu
1. Hover over a suitable location

Notice that the game now pauses like it does when you try to drag roads.

crudelios:
Fixed, thanks 👍

bvschaik:
Found another bug 😉  (They're getting more obscure now)

Steps:
1. Select bridge
2. Find a spot where it can be placed
3. Position the map in such a way that you can move your mouse from the good spot to the sidebar without hitting another tile
4. Select any build, a school or something
5. Hover over the map - school will display the cost of the bridge

Solution is pretty simple: in `building_construction_set_type()` we should also reset `data.cost` to 0.

crudelios:
That's a really obscure one, well spotted!

bvschaik (approved):

PR #546, opened 2020-11-20
Improvements to menu handling

### LINKED ISSUE #534: UI interaction when menu overlays are open
I often find myself trying to "click away" after I opened the **Overlays** dropdown menu in the top right corner. My intuition based on modern software UIs tells that the Overlays menu should automatically close, and the new action should behave as expected.

Example:

1. Open Overlays menu.
2. Decide that you want build a house instead, click on house icon.
3. Nothing happens, because you first need to right click / escape out of the Overlays menu.

Also, scrolling is not possible when the Overlays is opened.

image

I checked, and the original Caesar III has the same behaviour, however, this minor UI modernization would nicely fit into Julius. My proposal:

- Clicking any sidebar button or top menu item when the Overlays menu is open, should automatically close the Overlays menu and activate the clicked button.
- Scrolling when the Overlays menu is open should be possible.

What do you think?

crudelios:
Clicking outside overlays/military/build menu closes the menu (fixes #534)

Clicking a build menu icon button for a menu that is already open closes it

### DIFF
--- a/src/widget/sidebar/city.c
+++ b/src/widget/sidebar/city.c
@@ -293,22 +293,26 @@ static void button_build(int submenu, int param2)
 
 static void button_undo(int param1, int param2)
 {
+    window_build_menu_show(SUBMENU_NONE);
     game_undo_perform();
     window_invalidate();
 }
 
 static void button_messages(int param1, int param2)
 {
+    window_build_menu_show(SUBMENU_NONE);
     window_message_list_show();
 }
 
 static void button_help(int param1, int param2)
 {
+    window_build_menu_show(SUBMENU_NONE);
     window_message_dialog_show(param2, window_city_draw_all);
 }
 
 static void button_go_to_problem(int param1, int param2)
 {
+    window_build_menu_show(SUBMENU_NONE);
     int grid_offset = city_message_next_problem_area_grid_offset();
     if (grid_offset) {
         city_view_go_to_grid_offset(grid_offset);
--- a/src/widget/sidebar/city.c
+++ b/src/widget/sidebar/city.c
@@ -293,26 +293,26 @@ static void button_build(int submenu, int param2)
 
 static void button_undo(int param1, int param2)
 {
-    window_build_menu_show(SUBMENU_NONE);
+    window_build_menu_hide();
     game_undo_perform();
     window_invalidate();
 }
 
 static void button_messages(int param1, int param2)
 {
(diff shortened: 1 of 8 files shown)


bvschaik:
Bug found:
1. Open the overlay menu
1. Click the overlay button again

I expect the overlay menu either to stay open like it used to do, or for it to close, but instead it closes and immediately reopens.

src/widget/sidebar/city.c
```diff
@@ -293,22 +293,26 @@ static void button_build(int submenu, int param2)
 
 static void button_undo(int param1, int param2)
 {
+    window_build_menu_show(SUBMENU_NONE);
```
bvschaik, line 296: Instead of "show none", it's more readable to create a new function `window_build_menu_hide()` that internally calls `window_build_menu_show(SUBMENU_NONE)`. That also gets rid of the SUBMENU_NONE define in `build_menu.h`

src/widget/sidebar/editor.c
```diff
@@ -169,6 +169,7 @@ int widget_sidebar_editor_handle_mouse_attributes(const mouse *m)
 
 static void button_attributes(int show, int param2)
 {
+    window_editor_build_menu_show(MENU_NONE);
```
bvschaik, line 172: Same reasoning here: create a new function `window_editor_build_menu_hide()`.

src/window/military_menu.c
```diff
@@ -53,14 +58,29 @@ static void draw_foreground(void)
     data.active_buttons = num_legions;
 }
 
+static int click_outside_menu(const mouse *m, int x_offset)
+{
+    return m->left.went_down &&
```
bvschaik, line 63: Use `went_up` instead of down? That fixes the bug I reported.

src/window/overlay_menu.c
```diff
@@ -137,18 +145,27 @@ static void handle_submenu_focus(void)
     }
 }
 
+static int click_outside_menu(const mouse *m, int x_offset)
+{
+    return m->left.went_down &&
```
bvschaik, line 150: Use `went_up` instead of down? That fixes the bug I reported.

bvschaik (changes requested):

src/window/military_menu.c
crudelios: Yeah, oversight of mine...

bvschaik (approved):

PR #553, opened 2020-11-23
Changes to github actions

crudelios:
- Add Linux build with SDL 2.0.0 and SDL_mixer 2.0.0 (named `Linux x64 (old SDL versions)`)
- Add a `SKIP_BINTRAY` variable. When set, skips deployment to bintray.

I'm also thinking about adding a `RUN_TESTS` variable that runs `make test`. I don't think we need to run the tests for both Linux Appimage and Linux old SDL. What do you think?

Also, as you see I went ahead of what I originally set out to do. Let me know if you don't like the artifact/bintray deployment changes.

### DIFF
--- a/.ci_scripts/bintray_upload.sh
+++ b/.ci_scripts/bintray_upload.sh
@@ -61,11 +61,6 @@ case "$DEPLOY" in
   then
     DEPLOY_FILE=julius-$VERSION-android.apk
     cp "${build_dir}/julius.apk" "deploy/$DEPLOY_FILE"
-  elif [ -f "${build_dir}/julius-debug.apk" ]
-  then
-    DEPLOY_FILE=julius-debug-$VERSION-android.apk
-    cp "${build_dir}/julius-debug.apk" "deploy/$DEPLOY_FILE"
-    export SKIP_BINTRAY=true
   fi
   ;;
 *)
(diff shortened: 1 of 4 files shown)


bvschaik:
Github Actions artifacts are not internal, but visible to all logged-in users (I just checked with a new account). Only anonymous users cannot access them. I suggest not having artifacts for things we don't already publish to Bintray (with the possible exception of the other windows builds). I especially don't want Android debug APKs laying around.

Regarding the tests: they're not strictly needed, no, but they do not add much in the way of build times. For all checks to complete we have to wait 6 minutes for the Android build either way 😉

crudelios:
Okay, I'll remove the artifact then. I just don't like the warnings.

Unrelated question: There may be a way to cache gradle's sdl build. The only problem is that it may also cache unchanged julius files, I'm not sure.

That would obviously go to a different PR, but do you think it's worth it?

bvschaik:
> Unrelated question: There may be a way to cache gradle's sdl build. The only problem is that it may also cache unchanged julius files, I'm not sure.
> 
> That would obviously go to a different PR, but do you think it's worth it?

Knowing Android and Gradle a bit, I don't think that's a good idea. Keeping around compiled files from previous builds will either not help because Gradle decides to rebuild anyway, or worse: will confuse Gradle to the point that it will break builds randomly.

bvschaik (approved):

PR #556, opened 2020-12-17
Do not display hippodrome footprints if it cannot be built

HyperJeanJean:
This was originally proposed as a fix for Augustus, but as @crudelios suggested, here is a PR for Julius as well.

AFAIK, the fix is purely visual, it should not have any impact on save game compatibility.

As this bug was originally present in the original C3, I will totally understand if you don't want to merge this fix.

Here is a screenshot of the bug as seen in C3:

image


bvschaik (approved):
These kind of things are nice to have as improvement, thanks! :)

PR #559, opened 2020-12-23
Upgrade SDL to 2.0.14

bvschaik:
Works fine on Mac and Android.
@crudelios Could you perhaps test some other platforms?


crudelios:
Works fine on windows and Android too, I'd already tested :wink:

PR #561, opened 2020-12-29
Add ability to view minimap from CCK selection

bvschaik:
A personal pet peeve: there's no way to preview the maps in the CCK selection screen. This PR fixes that:
cck-minimap-button

The little minimap-like button is a toggle between the regular scenario conditions and the minimap:
cck-minimap-result

### DIFF
--- a/src/widget/scenario_minimap.c
+++ b/src/widget/scenario_minimap.c
@@ -108,9 +108,6 @@ static void draw_minimap_tile(int x_view, int y_view, int grid_offset)
             switch (map_property_multi_tile_size(grid_offset)) {
                 case 1: image_draw(image_id, x_view, y_view); break;
                 case 2: image_draw(image_id + 1, x_view, y_view - 1); break;
-                case 3: image_draw(image_id + 2, x_view, y_view - 2); break;
-                case 4: image_draw(image_id + 3, x_view, y_view - 3); break;
-                case 5: image_draw(image_id + 4, x_view, y_view - 4); break;
             }
         }
     } else {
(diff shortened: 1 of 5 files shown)


src/widget/scenario_minimap.c
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    int terrain = map_terrain_get(grid_offset);
+
+    if (terrain & TERRAIN_BUILDING) {
+        // Native huts/fields
+        if (map_property_is_draw_tile(grid_offset)) {
+            int image_id = image_group(GROUP_MINIMAP_BUILDING);
+            switch (map_property_multi_tile_size(grid_offset)) {
+                case 1: image_draw(image_id, x_view, y_view); break;
```
crudelios, line 109: Instead of `case N:`, you could use:

```
int offset = map_property_multi_tile_size(grid_offset) - 1;
image_draw(image_id + offset, x_view, y_view - offset);
```

crudelios:
Upon closing and reopening the CCK window, should it still display the map or revert to the scenario info?

src/widget/scenario_minimap.c
bvschaik: That will lead to problems when buildings ever get bigger than 5x5 (Augustus...). In fact, for the scenario minimap we'll only ever need 1x1 and 2x2 so I'll get rid of the other cases.

bvschaik:
> Upon closing and reopening the CCK window, should it still display the map or revert to the scenario info?

Good point. I think it should reset to the scenario info. I'll fix this.

crudelios:
LGTM 👍

PR #564, opened 2020-12-31
Updated Korean Translation 1

sickchicken:
yg3988 and I will compare, update and double-check each other's texts. First, I will upload mine, and he will fork updated version and fill some texts which need in-game context(he knows, I don't)
Not sure if I'm doing it right, but I'll try anyway.


crudelios:
This is not correct. You must add a PR to Keriew's fork of Julius, not the base Julius. When submitting the PR, at the top part (before the description input), you must select Keriew/augustus - master instead of bvschaik/julius - master.

PR #573, opened 2021-01-08
Added a hotkey to snow house merge locations on cleared tiles

kekekeks:
In the original game it's possible to know if a certain tile is a valid merge point by repeatedly clearing it (for no cost since it's already clean) and looking at the land pattern.

This PR introduces a quality of life hotkey to just show valid merge locations on the map for clear tiles. 

This is purely a visual change and doesn't give the player any extra advantage except for not needing to repeatedly clear tiles and memorize merge locations.

screen


kyklish:
**Valid merge point** - can you explain on example, please? I don'n understand.

kekekeks:
Houses up to large insulae can only merge if the top tile of the would be 2x2 is a valid merge point determined by the map random seed. Starting from grand insulae houses are forcefully merged into 2x2 regardless of the position.

bvschaik:
Hi @kekekeks, thanks for the PR, but I do not think this functionality should be in Julius for the following reasons:

- It's an unpredictable gameplay mechanic. Showing these spots, thereby making it predictable, feels like cheating to me.  
- Using it requires the player to know exactly what the algorithm is that's causing the houses to merge, and in which direction. Without that knowledge, it's just a bunch of dots on the screen that don't make sense. I think that less than 1% of players are aware of the merging rules and even less of them are interested in knowing the merge points in advance.

By the way, I had a brief look at the code: please follow the [coding conventions](https://github.com/bvschaik/julius/wiki/Coding-conventions) for any future work.

kekekeks:
@bvschaik After playtesting for several days those dots turned out to be really, *really* useful for measuring distances and identifying buildable tiles. Do you mind a PR that just shows grey dots at empty tiles?

bvschaik:
We already have a measuring tool: the "Show draggable construction size" option.
As for displaying buildable tiles: isn't the "empty land" graphic enough of a clue?

kekekeks:
>isn't the "empty land" graphic enough of a clue?

изображение

I've failed to identify that single tile I can't build on here until it was time to merge into 3x3

PR #575, opened 2021-01-09
Julius check hotkey duplicate

kyklish:
Suggestion to not allow duplicated hotkeys.
From this:
Hotkey Duplicated Issue
to this:
Hotkey Duplicated Solution Centered Custom
Final result:
Hotkey Duplicated Solution Extra Centered
**Known issues**: translate needed.


src/window/hotkey_duplicate.c
```diff
@@ -0,0 +1,90 @@
+#include "hotkey_duplicate.h"
```
bvschaik, line 1: I think it's better to expand the `plain_message_dialog` with an extra (optional) text parameter rather than to duplicate the code for this dialog.

src/window/hotkey_duplicate.c
```diff
@@ -0,0 +1,90 @@
...
+static struct {
+    const uint8_t *title;
+    const uint8_t *message;
+    const uint8_t *action_name;
+} data;
+
+static int init(translation_key title, translation_key message, const uint8_t * action_name)
+{
+    // QUESTION
+    // do we need this here? copy paste from WINDOW_PLAIN_MESSAGE_DIALOG
```
bvschaik, line 27: Answer: if we expand `plain_message_dialog` instead, we don't need it here because this file won't be necessary.
If we keep this dialog, I think it's also not needed, since it's a specific dialog, not a generic one like the `plain_message_dialog`.

src/translation/translation.h
```diff
@@ -97,6 +97,8 @@ typedef enum {
     TR_HOTKEY_SET_BOOKMARK_4,
     TR_HOTKEY_EDITOR_TOGGLE_BATTLE_INFO,
     TR_HOTKEY_EDIT_TITLE,
+    TR_HOTKEY_DUPLICATE_TITLE,
+    TR_HOTKEY_DUPLICATE,
```
bvschaik, line 101: Rename to `TR_HOTKEY_DUPLICATE_MESSAGE` to indicate it's the message for the dialog.

src/translation/english.c
```diff
@@ -104,7 +104,9 @@ static translation_string all_strings[] = {
     {TR_HOTKEY_SET_BOOKMARK_3, "Set bookmark 3"},
     {TR_HOTKEY_SET_BOOKMARK_4, "Set bookmark 4"},
     {TR_HOTKEY_EDITOR_TOGGLE_BATTLE_INFO, "Toggle battle info"},
-    {TR_HOTKEY_EDIT_TITLE, "Press new hotkey"}
+    {TR_HOTKEY_EDIT_TITLE, "Press new hotkey"},
+    {TR_HOTKEY_DUPLICATE_TITLE, "Hotkey is duplicated"},
```
bvschaik, line 108: Suggested text: `Hotkey already used`

src/translation/english.c
```diff
@@ -104,7 +104,9 @@ static translation_string all_strings[] = {
...
+    {TR_HOTKEY_DUPLICATE, "This key combination assigned to action:"}
```
bvschaik, line 109: Suggested text:
`This key combination is already assigned to the following action:`

src/window/hotkey_config.c
```diff
@@ -283,10 +285,54 @@ static void handle_input(const mouse *m, const hotkeys *h)
     }
 }
 
+static const uint8_t *hotkey_action_name_for(hotkey_action action)
+{
+    const uint8_t *name = 0;
+    for (int i = 0; i < (NUM_VISIBLE_OPTIONS + scrollbar.max_scroll_position); i++) {
```
bvschaik, line 291: Parentheses `()` are not necessary

src/window/hotkey_config.c
```diff
@@ -283,10 +285,54 @@ static void handle_input(const mouse *m, const hotkeys *h)
...
+            if (widget->name_translation != TR_NONE) {
+                name = translation_for(widget->name_translation);
+            } else {
+                name = lang_get_string(widget->name_text_group, widget->name_text_id);
+            }
+            break;
+        }
+    }
+    // QUESTION
+    // is it OK return NULL pointer and crash game if we did not find "name"?
```
bvschaik, line 303: Answer: yes. If we reuse the "plain_message_dialog" instead of having a custom dialog, like I suggested in another comment, it will simply not display the message when this function returns NULL.

bvschaik (changes requested):
Functionality works well!

One suggestion for the duplicate dialog and a couple minor improvements. Overall it's looking good.

src/window/hotkey_duplicate.c
kyklish: Does pure C allow optional parameters in function declaration?

bvschaik: Yes, a second function is best in this case.
Suggestion: name the function `window_plain_message_dialog_show_with_extra` the parameter `extra`?

kyklish: Done.

bvschaik (approved):

PR #585, opened 2021-02-09
Improve icons on Linux

crudelios:
- Add Icon to taskbar (missing 32x32 file before)
- ~Add icon to window title~


crudelios:
Closing for now as in Windows the icon becomes blurry. I'll make the change linux-only and reopen this.

bvschaik:
Doesn't Linux pick up the icons in `share` automatically after doing an install? I guess that also depends on which desktop environment the user is running.

crudelios:
Yes it does, except for the window title's icon. That one requires a specific `pixbuf` (or, in our case, `SDL_SetWindowIcon` deals with that for us.

Also, at least my distro requires 32x32 icon size in the `share` directory to show an icon in the taskbar. The other icons we install work for the "start menu", but not for the taskbar.

bvschaik:
Suggestion: just add the 32x32 icon, and leave the window title icon? It's not a huge issue, and I don't quite like having that binary blob in the code.

bvschaik (approved):

PR #586, opened 2021-02-15
File dialog improvements

bvschaik:
This PR adds the following improvements to the file dialog:

- Typing in the input box now scrolls the list to the first file that starts with the typed text
- When loading a saved game, the placeholder filename "My Rome" is no longer pre-filled - it's only used for the "save file" dialog
- Multi-byte characters are taken into account when sorting

### DIFF
--- a/src/core/string.c
+++ b/src/core/string.c
@@ -1,16 +1,5 @@
 #include "core/string.h"
 
-#include <ctype.h>
-
-#if _MSC_VER
-// Of course MSVC is the only compiler that doesn't have strcasecmp...
-#include <string.h>
-#define strcasecmp _stricmp
-#define strncasecmp _strnicmp
-#else
-#include <strings.h>
-#endif
-
 int string_equals(const uint8_t *a, const uint8_t *b)
 {
     while (*a && *b && *a == *b) {
@@ -142,14 +131,4 @@ int string_from_int(uint8_t *dst, int value, int force_plus_sign)
     }
 
     return total_chars;
-}
-
-int string_compare_case_insensitive(const char *a, const char *b)
-{
-    return strcasecmp(a, b);
-}
-
-int string_compare_case_insensitive_prefix(const char *full_string, const char *prefix, int prefix_len)
-{
-    return strncasecmp(full_string, prefix, prefix_len);
 }
--- a/src/core/string.c
+++ b/src/core/string.c
@@ -2,6 +2,28 @@
 
 #include <ctype.h>
 
+#if _MSC_VER
+// Of course MSVC is the only compiler that doesn't have strcasecmp...
+#include <string.h>
+#define strcasecmp _stricmp
+#define strncasecmp _strnicmp
+#else
+#include <strings.h>
+#endif
+
+int string_equals(const uint8_t *a, const uint8_t *b)
+{
+    while (*a && *b && *a == *b) {
+        ++a;
+        ++b;
+    }
+    if (*a == 0 && *b == 0) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
+
 void string_copy(const uint8_t *src, uint8_t *dst, int maxlength)
 {
     int length = 0;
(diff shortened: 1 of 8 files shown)


src/core/string.c
```diff
@@ -2,6 +2,28 @@
...
 #include <ctype.h>
 
+#if _MSC_VER
+// Of course MSVC is the only compiler that doesn't have strcasecmp...
+#include <string.h>
+#define strcasecmp _stricmp
+#define strncasecmp _strnicmp
+#else
+#include <strings.h>
+#endif
```
crudelios, line 12: Adding a `MSVC` define to a `core` file makes me sad 😭

src/core/string.c
```diff
@@ -124,33 +146,10 @@ int string_from_int(uint8_t *dst, int value, int force_plus_sign)
...
-    while (*a && *b && *a == *b) {
-        ++a;
-        ++b;
-    }
-    if (*a == 0 && *b == 0) {
-        return 1;
-    } else {
-        return 0;
-    }
+    return strncasecmp(full_string, prefix, prefix_len);
```
crudelios, line 154: I know you did this to avoid code duplication, but I really don't like the use of the non-standard `strcasecmp` and `strncasecmp`...

src/window/file_dialog.c
```diff
@@ -78,25 +80,64 @@ static input_box file_name_input = {144, 80, 20, 2, FONT_NORMAL_WHITE, 0, data.t
...
+    data.message_not_exist_start_time = 0;
+    data.double_click = 0;
+    data.focus_button_id = 0;
+
+    if (strlen(data.file_data->last_loaded_file) > 0) {
         encoding_from_utf8(data.file_data->last_loaded_file, data.typed_name, FILE_NAME_MAX);
         file_remove_extension(data.typed_name);
+    } else if (dialog_type == FILE_DIALOG_SAVE) {
+        // Suggest default filename
+        string_copy(lang_get_string(9, type == FILE_TYPE_SCENARIO ? 7 : 6), data.typed_name, FILE_NAME_MAX);
```
crudelios, line 131: Just occured to me. Instead of a simple `My Rome`, why not display the name of the city and maybe the date as the default filename?

crudelios (commented):
Allright, just tested, everything seems to work fine.

Some comments/suggestions follow.

src/core/string.c
bvschaik: I did this because the existing implementation of `string_compare_case_insensitive` doesn't work for non-ascii (multibyte) characters, always placing them above the files starting with ascii characters.

Now that I think of it: it's also weird that these two functions are the only ones in this file that operate on `char *`, while the others are for `uint8_t *`. Since all usages of `string_compare_case_insensitive*` are related to file names, I think it makes sense to move these functions to `file_manager` and handle MSVC/Windows in the proper way there. What do you think?

src/window/file_dialog.c
bvschaik: Then the File dialog needs to know about cities and empire maps, and this is not applicable to scenarios in the editor. Besides, "My Rome" is only used if the user has not loaded a file of the same type before. So if you load saved game "Massilia 120AD", then start a new CCK scenario and try to save it, it will still suggest "Massilia 120AD" as file name.
I'd rather keep the "My Rome" (or "Scenario1") than complicate this screen with more logic.

src/core/string.c
crudelios: Lovely idea! But change it to `compare_file_name` out something...

bvschaik: I just moved the functions to `file_manager`, looks much cleaner!

Could you test this on windows with both the mingw and msvc builds?
On my Mac it lists a file starting with "ä" under "a" now, which is pretty nice, but that may just be a platform thing.

src/window/file_dialog.c
```diff
@@ -78,25 +81,64 @@ static input_box file_name_input = {144, 80, 20, 2, FONT_NORMAL_WHITE, 0, data.t
 static file_type_data saved_game_data = {"sav"};
 static file_type_data scenario_data = {"map"};
 
-static int double_click = 0;
+static int find_first_file_with_prefix(const char *prefix)
+{
+    int len = strlen(prefix);
```
crudelios, line 86:
```suggestion
    size_t len = strlen(prefix);
```

Please change otherwise MSVC throws a warning.

src/core/string.c
crudelios: Both mingw and msvc compile and run.

However, unlike mac, `a` won't match `á`. For some reason, `à` is also not matching `à`, although `é` is matching `é`.

crudelios: Another comment: When deleting chars (using backspace) the file manager keeps updating the filter and thus moving. Shouldn't the filter only apply when the user inputs chars, but not when removing them?

bvschaik: Good suggestion. I wasn't sure what the user would expect when deleting characters.

crudelios:
The code looks good!

I'll test it tomorrow morning.

crudelios:
Tested and working, ready to merge! 👍

PR #587, opened 2021-02-17
Update vita toolchain

crudelios:
Update vita toolchain, which uses a more recent/better integrated SDL build.

Allows getting rid of the `vita2d` hacks and improves performace on Vita by 5% to 10%.

### DIFF
--- a/src/platform/screen.h
+++ b/src/platform/screen.h
@@ -18,7 +18,8 @@ void platform_screen_center_window(void);
 void platform_screen_recreate_texture(void);
 #endif
 
-void platform_screen_render(void);
+void platform_screen_render(int update_screen);
+void platform_screen_clear(void);
 
 void platform_screen_generate_mouse_cursor_texture(int cursor_id, int scale, const color_t *cursor_colors);
 
--- a/src/platform/screen.h
+++ b/src/platform/screen.h
@@ -18,8 +18,9 @@ void platform_screen_center_window(void);
 void platform_screen_recreate_texture(void);
 #endif
 
-void platform_screen_render(int update_screen);
 void platform_screen_clear(void);
+void platform_screen_update(void);
+void platform_screen_render(void);
 
 void platform_screen_generate_mouse_cursor_texture(int cursor_id, int scale, const color_t *cursor_colors);
 
--- a/src/platform/vita/vita_keyboard.c
+++ b/src/platform/vita/vita_keyboard.c
@@ -2,16 +2,15 @@
 
 #include "core/calc.h"
 #include "core/encoding.h"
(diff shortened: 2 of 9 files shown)


src/platform/screen.h
```diff
@@ -18,7 +18,8 @@ void platform_screen_center_window(void);
 void platform_screen_recreate_texture(void);
 #endif
 
-void platform_screen_render(void);
+void platform_screen_render(int update_screen);
```
bvschaik, line 21: Can we please eliminate this boolean flag by creating two functions?
Perhaps `platform_screen_draw_and_render(void)` (original code) and `platform_screen_render(void)` with just the `SDL_RenderPresent()`.

For rationale, see Clean Code by Robert C. Martin, or [this article](https://martinfowler.com/bliki/FlagArgument.html) by Martin Fowler.

src/platform/vita/vita_keyboard.c
```diff
@@ -109,27 +110,20 @@ const uint8_t *vita_keyboard_get(const uint8_t *initial_text, int max_length)
...
         if (ime_result == IME_DIALOG_RESULT_FINISHED) {
             encoding_from_utf8(ime_text_utf8, final_text, max_length);
         } else if (ime_result != IME_DIALOG_RESULT_CANCELED) {
             done = 0;
         }
-        vita2d_end_drawing();
-        vita2d_common_dialog_update();
-        vita2d_swap_buffers();
-        sceDisplayWaitVblankStart();
+        platform_screen_render(0);
```
bvschaik, line 126: Just wondering: what's the reason for not wanting to redraw the screen at this point?

bvschaik (changes requested):
Generally looking good. Nice to be rid of the vita-specific code.
One requested improvement and one question.

src/platform/screen.h
crudelios: I completely agree. I was thinking about doing just that when I had some time 😉 

I was thinking about `platform_screen_update` and `platform_screen_render`. Normal code would call both functions, the vita keyboard code would just need to call `render`.

src/platform/vita/vita_keyboard.c
crudelios: The normal redraw gets the drawing from our texture which is controlled by `graphics_canvas`. When drawing the virtual keyboard, we don't want/need the texture from there but rather the internal rendering from vita's virtual keyboard.  So we only need to call renderer but not to update the texture.

src/platform/screen.h
bvschaik: Sounds good 👍

PR #593, opened 2021-03-17
Show current active requests on the sidebar

crudelios:
This requires some new strings, which I added to translations.

I could not find suitable translations for all the strings, so let me know what you think about adding these.


crudelios:
This has been up for a while and is frankly outdated when compared to augustus.

Should I close this?

bvschaik:
> Should I close this?

Perhaps that's for the best, yes. I can't find the energy these days to spend much time on Julius.

PR #594, opened 2021-03-30
Add Emscripten build

crudelios:
So Julius can be played on a browser by the 0.8 people that are interested in it.


crudelios:
Apparently the "add emsdk" github action I added cannot be run from a PR.

I'm guessing it's a permission issue in your repo?

crudelios:
I'm adding mp3 support to the emscripten build, which involves fiddling with the build system, so to prevent new notifications each time I try to get the build working, I'll close this for now and reopen when it's ready again.

crudelios:
Reopening as it's now compiling the dependencies and also caching them.

CMakeLists.txt
```diff
@@ -11,9 +11,13 @@ endif()
 
 option(DRAW_FPS "Draw FPS on the top left corner of the window." OFF)
 option(SYSTEM_LIBS "Use system libraries when available." ON)
+option(EMSCRIPTEN_LOAD_SDL_PORTS "Load SDL and SDL_mixer emscripten ports instead of compiling them" OFF)
+option(LINK_MPG123 "Link mpg123 statically to Julius instead of relying on a library." OFF)
+
 cmake_dependent_option(VITA_BUILD "Build for the PlayStation Vita handheld game console." OFF "NOT MSVC" OFF)
 cmake_dependent_option(SWITCH_BUILD "Build for the Nintendo Switch handheld game console." OFF "NOT MSVC; NOT VITA_BUILD" OFF)
 cmake_dependent_option(ANDROID_BUILD "Build for the Android environment." OFF "NOT MSVC; NOT VITA_BUILD; NOT SWITCH_BUILD" OFF)
+cmake_dependent_option(EMESCRIPTEN_BUILD "Build for the Emscripten environment." OFF "NOT VITA_BUILD; NOT SWITCH_BUILD; NOT ANDROID_BUILD" OFF)
```
bvschaik, line 20: Should we start thinking about a TARGET_PLATFORM variable instead of one boolean for each platform?

CMakeLists.txt
```diff
@@ -691,8 +732,12 @@ if(APPLE)
     " BUNDLE DESTINATION ${CMAKE_BINARY_DIR})
 endif()
 
-include_directories(${SDL2_INCLUDE_DIR})
-include_directories(${SDL2_MIXER_INCLUDE_DIR})
+if(SDL2_INCLUDE_DIR)
+    include_directories(${SDL2_INCLUDE_DIR})
```
bvschaik, line 736: In which case would the include dir be empty? Where do the SDL includes come from in that case?

bvschaik (commented):
Couple of questions. Also, do you know if it's possible to split the CMakeLists.txt file? With 800+ lines it's getting too big to my taste.

CMakeLists.txt
crudelios: When the new variable `EMSCRIPTEN_LOAD_SDL_PORTS` (which I added to `CMakeLists.txt`) is set, emscripten will be run with the options `-s  USE_SDL=2 -s USE_SDL_MIXER=2`, which will automatically download and include them without needing to configure their inclusion on cmake.

This is useful if the user wants to compile Julius for emscripten and not have the hassle of compiling libmpg123, SDL and SDL_mixer, which are kinda hard to configure for emscripten. However,  the default ports don't have mp3 support, which our manual compilation does.

crudelios:
> Also, do you know if it's possible to split the CMakeLists.txt file? With 800+ lines it's getting too big to my taste.

What if we move the source files to a `cmake_source_files` file?

CMakeLists.txt
crudelios: Check Discord 😉

bvschaik:
> What if we move the source files to a `cmake_source_files` file?

Actually, I wanted to extract the platform-specific bits, since those distract from the main 'flow' that a build follows. But that's for a later PR :)

bvschaik (approved):

PR #596, opened 2021-04-05
Add Haiku OS compatibility

brownna01:
These small changes are all that's required to be able to build and run natively on Haiku with the Haiku-provided SDL2, SDL2_mixer, and libpng libraries.

The only challenge with these changes in place is the cmake command, which can't natively find the SDL2_INCLUDE_DIR and SDL2_MIXER_INCLUDE_DIR. If specified in the CMake command line as -DSDL2_INCLUDE_DIR=/boot/system/develop/headers/SDL2 -DSDL2_MIXER_INCLUDE_DIR=/boot/system/develop/headers/SDL2, then it builds and runs just fine on Haiku R1 beta2.

### DIFF
--- a/src/platform/haiku/haiku.h
+++ b/src/platform/haiku/haiku.h
@@ -1,0 +1,9 @@
+#ifndef PLATFORM_HAIKU_H
+#define PLATFORM_HAIKU_H
+
+#ifdef __HAIKU__
+
+#define PLATFORM_USE_SOFTWARE_CURSOR
+
+#endif // __HAIKU__
+#endif // PLATFORM_HAIKU_H
--- a/src/platform/screen.c
+++ b/src/platform/screen.c
@@ -9,9 +9,9 @@
 #include "input/cursor.h"
 #include "platform/android/android.h"
 #include "platform/cursor.h"
+#include "platform/haiku/haiku.h"
 #include "platform/switch/switch.h"
 #include "platform/vita/vita.h"
-#include "platform/haiku/haiku.h"
 
 #include "SDL.h"
 
--- a/src/platform/screen.c
+++ b/src/platform/screen.c
@@ -11,6 +11,7 @@
 #include "platform/cursor.h"
 #include "platform/switch/switch.h"
 #include "platform/vita/vita.h"
+#include "platform/haiku/haiku.h"
 
 #include "SDL.h"
 
--- a/cmake/FindSDL2.cmake
+++ b/cmake/FindSDL2.cmake
@@ -106,6 +106,7 @@ ELSE()
         /opt/local # DarwinPorts
         /opt/csw # Blastwave
         /opt
+        /boot/system/develop/headers/SDL2 # Haiku
         ${SDL_EXT_DIR}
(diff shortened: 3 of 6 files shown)


crudelios:
This is great, all new ports are welcome. 👍 

You can probably automate finding the include directories by adding `/boot/system/develop/headers` to `SDL2_SEARCH_PATHS` in both `FindSDL.cmake` and `FindSDL_mixer.cmake`.

brownna01:
Thanks for the hint, I'm not particularly fluent in cmake.

src/platform/haiku/haiku.h
```diff
@@ -0,0 +1,9 @@
+#ifndef PLATFORM_HAIKU_H
+#define PLATFORM_HAIKU_H
+
+#ifdef __HAIKU__
+
+#define PLATFORM_USE_SOFTWARE_CURSOR
```
crudelios, line 6: Just out of curiosity, don't the regular cursors work? Using the software cursor has some drawbacks as the cursor position will only update according to the game's fps.

brownna01: They did not for me. Without software cursor set to true, I had an invisible cursor in both VMware and on a laptop. Mousing over menu items did indicate that the mouse was working, but you couldn't see it unless there was a mouseover change, like a highlight on a button.

Doing a bit of research, I found references to an SDL bug in the Haiku implementation, and indications that hardware cursors weren't fully implemented. So it's most likely one of those two things forcing this.

bvschaik: Do you have any links for the fact that hardware cursors aren't fully implemented?
Which version of SDL does Haiku use? I saw a [commit](https://github.com/libsdl-org/SDL/commit/72bcf546f90423f05eda0fa08510e340683ff6f4) on SDL for Haiku cursor code. Does it make sense to wait for SDL 2.0.16 - or at least re-check this later?

src/platform/screen.c
```diff
@@ -11,6 +11,7 @@
 #include "platform/cursor.h"
 #include "platform/switch/switch.h"
 #include "platform/vita/vita.h"
+#include "platform/haiku/haiku.h"
```
bvschaik, line 14: Same comment about the order of includes.

crudelios:
It seems the latest master of SDL has no way to add a custom cursor.

According to [this information](https://www.haiku-os.org/docs/api/classBCursor.html#details), Haiku only supports 16x16 monochrome cursors, although it apparently [supports loading custom cursor from bitmaps](https://www.haiku-os.org/docs/api/classBCursor.html#afa8336bae1b43af072e1c855960fd5fe), so I'm not sure what to say regarding support.

Still, taking everything into account, I think software cursor is, for now, our best bet for Haiku.

brownna01:
That seems reasonable to me, you found a bit more than I did about it. I compiled SDL2 from the point of the commit you referenced earlier and had the same issues with invisible cursors.

Perhaps I'll try and figure out something in the future with the bitmap cursors.

bvschaik (approved):

PR #617, opened 2021-08-06
Fix earthquake corrupting warehouses. Fixes #616

### LINKED ISSUE #616: Earthquakes hitting warehouses can cause corruption.
Multiple part buildings hit by earthquakes can cause some tiles to become corrupted, making building grid link to wrong buildings and cause eternal fires. Link to a save which displays this behaviour: [earthquake bug.zip](https://github.com/bvschaik/julius/files/6947460/earthquake.bug.zip) While this matches the behaviour from Caesar 3, most of the other earthquake/corruption bugs have been fixed, so I thought it'd be a good idea to fix this one too.

Mechanism of the bug happening:

Earthquake tries to expand. Game calls ```advance_earthquake_to_tile``` from ```scenario/earthquake.c```, causing a warehouse to turn into the rubble. That tile then becomes an earthquake tile and the warehouse/storage bay that existed before has its image removed and its state is changed to ```BUILDING_STATE_DELETED_BY_GAME```. Before ```building_update_state``` is called (which would remove all the data of that warehouse/storage bay), ```advance_earthquake_to_tile``` is called again, advancing on a different rubble tile of that same warehouse. The rubble tile is destroyed via ```building_destroy_by_fire```, which calls ```destroy_linked_parts```. This causes all the tiles of the former warehouse to be destroyed again, included the one marked as ```BUILDING_STATE_DELETED_BY_GAME```, which adds it back to the building grid as the burning ruin. The building is later deleted by the game, which causes the next built building to share its id, linking rubble with that building.

Keriew:
Prevents deleted buildings from leaving rubble on rock. This should never happen in usual circumstances, since player can't built on rock and no buildings are placed on rock, but can happen because of the earthquake bug.

### DIFF
--- a/src/building/destruction.c
+++ b/src/building/destruction.c
@@ -44,7 +44,7 @@ static void destroy_on_fire(building *b, int plagued)
         num_tiles = 0;
     }
     map_building_tiles_remove(b->id, b->x, b->y);
-    if (map_terrain_is(b->grid_offset, TERRAIN_WATER)) {
+    if (map_terrain_is(b->grid_offset, TERRAIN_WATER | TERRAIN_ROCK)) {
         b->state = BUILDING_STATE_DELETED_BY_GAME;
     } else {
         b->type = BUILDING_BURNING_RUIN;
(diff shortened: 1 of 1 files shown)


Keriew:
Caught a few more problems, this isn't a good solution.

crudelios:
Perhaps a better way is to immediately unlink the building parts when the warehouse first gets destroyed?

PR #630, opened 2021-10-17
Improve touch scrolling on windows

crudelios:
Allows holding a finger and dragging inside scrollable elements, like a regular scrollable interface using touch.

Hopefully doesn't break regular scrolling.


crudelios:
Yes it was added to augustus years ago haha! And it doesn't break anything.

bvschaik (approved):

PR #647, opened 2022-03-20
Timelapse: Automatically take screenshots every month, if enabled

nixxquality:
Timelapses are fun, especially in citybuilders.

It's a feature that exists in various games in various ways. Some examples are Rimworld with a timelapse mod that does something similar to this, Europa Universalis IV that uses the game history in the savefile to simulate an in game timelapse and Frostpunk that records screenshots of your city as you play and then plays you a nice "at what cost" video when you win.

I've added a rudimentary but working timelapse feature to Julius. I apologize in advance for any violence I've caused to your code, which is otherwise very pretty.

- The full city screenshot function was copy-pasted and modified for this purpose.
The changes were that I removed the message popup and the unnecessary IS_CITY check, since time can only progress in city view. I also threw in some rotation code so that it would always take screenshots facing north even if the user uses another view momentarily.
- The monthly tick checks if the new timelapse screenshot setting is enabled and if so calls the save screenshot function.
I assume graphics code is not meant to be called from the core logic like this, because the tests fail to build with this setup.
I guess I'd need to add some sort of signaling or maybe an inbetween function that is a noop on test builds to fix this issue.
Due to my unfamiliarity with your code I will have to rely on your judgment.
- Translation needs to be looked at
I don't know how to add a new translation line like this without breaking the game for all other languages. Again, I will rely on you.
- Screenshots are saved in the game folder as `timelapse %Y-%m-%d %H.%M.%S.png`.
If it would be possible to access the map name from the screenshot code that could be an improvement.
- The feature is disabled by default, and available in the configuration menu.
It might need its own category or something, right now it's snuggled up to the gameplay changes.
- Right now input gets a little bit screwed when screenshots are taken.
For example, if your camera is smoothing to a position it gets stopped in its tracks. There might be other issues that me, as someone who has never played C3 before, might not now about. I'm not sure what can be done about it.

I'm very open to feedback so let me know what I'll need to improve to put this in the game.


feiyunw:
It worries me that saving a lot pictures in the game directory, without player interaction and notification. A player can return to the game in a month and forget what he selected before.

How about this idea:
I would like a full city screenshot when mission accomplished, so that I could optimize game playing next time.  I was always too excited to have the screenshot when the winning messages pop up.  A camera-like icon indicating a screenshot in the winning message box will be sufficient for the UI change, without any translation work.

QuakeIV:
i only just noticed this and its really cool looking

PR #649, opened 2022-03-25
Support macOS 10.4

evanmiller:
macOS 10.4 lacks `execinfo.h`. Adding the proposed logic allows the project to be compiled and the game to be played on this very old platform.

(I have read CONTRIBUTING.md and understand I am supposed to raise an issue first, but it's a small change and this seemed like the easiest way to discuss it.)


crudelios:
Thanks for the change!

PR #656, opened 2022-09-19
Disable mouse wheel zoom

hankyuller:
It's easy to unwillingly zoom in or out while dragging the map with right mouse button.


crudelios:
You don't want this PR here, but on Keriew/augustus instead. Please open the PR there. Thanks!

hankyuller:
I understand, please do it for me as my old computer can't even handle this website, thanks !

crudelios:
I believe you didn't add any code to do it, you only requested it, right?

Then no PR is needed.

hankyuller:
_I believe you didn't add any code to do it, you only requested it, right?_

Right

_Then no PR is needed._

Ok, post an issue then ?

_Maybe zoom could be disabled only when right-click scrolling, what do you think?_

It would only help a little as you can still zoom in or out while reaching for the right button, this is all because of the apple magic mouse really. There used to be that exact option "disable mouse wheel zooming" but it was removed in recent versions.

crudelios:
Actually it was "disable zoom" entirely. I think it zooming should be added as a hotkey, but it being a mouse system (not a keyboard press), it would require a major rework of the engine.

I'll move this to the proper place to stop polluting Julius with Augustus stuff.

hankyuller:
Thanks !

PR #660, opened 2022-10-31
Refactor some core code

rsf92:
This is just a small refactor I made while checking the code. Just in case it's deemed a good piece of code

### DIFF
--- a/src/core/buffer.c
+++ b/src/core/buffer.c
@@ -6,8 +6,7 @@ void buffer_init(buffer *buf, void *data, int size)
 {
     buf->data = data;
     buf->size = size;
-    buf->index = 0;
-    buf->overflow = 0;
+    buffer_reset(buf);
 }
 
 void buffer_reset(buffer *buf)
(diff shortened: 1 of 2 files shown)


src/core/buffer.c
```diff
@@ -23,11 +22,8 @@ void buffer_set(buffer *buf, int offset)
 
 static int check_size(buffer *buf, int size)
 {
-    if (buf->index + size > buf->size) {
-        buf->overflow = 1;
-        return 0;
-    }
-    return 1;
+    buf->overflow = (buf->index + size) > buf->size;
```
bvschaik, line 25: To be honest I find the original code cleaner: only setting the overflow variable iff there actually is an overflow.
Also, which value is stored in overflow is compiler dependent, so I try to avoid it.

src/core/calc.c
```diff
@@ -7,12 +7,11 @@ int calc_adjust_with_percentage(int value, int percentage)
 
 int calc_percentage(int value, int total)
 {
+    int ret = 0;
```
bvschaik, line 10: Why introduce a new variable when you can use multiple return statements? Ok, it saves an `else` block, but I find that more readable than setting a variable to return later.

rsf92:
Okay, I'll close the PR, your reasoning is sound enough I won't even try to argue :)

PR #662, opened 2022-12-01
Fix bug where sheep could get an out of bounds destination.

PrettyFlower:
To reproduce, load up Tarraco from campaign (attached), and add a conditional breakpoint here:
image

After a short wait this should get hit, and will result in an out of bounds memory check on the next line. What's happening here is:
- In `formation.c`, the sheep formation gets initialized with a destination x/y of 0,0. For campaign it's loaded from a saved game but the same thing happens from CCK.
- The sheep are unable to find a location to move to, so the formation destination x/y stays at 0.
- In `animal.c`, the sheep figure gets its destination_x/y set based on the formation destination_x/y plus the formation offset.
- Since some formation offset values can be negative and the sheep figure's destination_x/y are unsigned chars, this results in the destination_x/y values rolling over into the 250s (i.e., 0 + -1 = 255), which is out of the bounds of the map.
- The sheep then attempt to find an out of bounds destination, resulting in a bad lookup.

I've gone ahead and fixed the issue for all the animal herds. It wasn't clear to me if this was an issue for soldier and enemy formations as well, but we should probably check that.

[Architect1.zip](https://github.com/bvschaik/julius/files/10129057/Architect1.zip)

### DIFF
--- a/src/figure/formation_layout.c
+++ b/src/figure/formation_layout.c
@@ -1,7 +1,5 @@
 #include "formation_layout.h"
 
-#include "map/grid.h"
-
 static const int FORMATION_LAYOUT_POSITION_X[FORMATION_MAX][MAX_FORMATION_FIGURES] = {
     {0, 1, 0, 1, -1, -1, 0, 1, -1, 2, 2, 2, 0, 1, -1, 2},
     {0, 0, -1, 1, -1, 1, -2, -2, 2, 2, -3, -3, 3, 3, -4, -4},
--- a/src/figure/formation_layout.c
+++ b/src/figure/formation_layout.c
@@ -42,14 +42,3 @@ int formation_layout_position_y(int layout, int index)
 {
     return FORMATION_LAYOUT_POSITION_Y[layout][index];
 }
-
-void formation_layout_position(int layout, int index, formation *m, uint8_t *x, uint8_t *y)
-{
-    int offs_x = formation_layout_position_x(layout, index);
-    int offs_y = formation_layout_position_y(layout, index);
-    int dest_x = m->destination_x + offs_x;
-    int dest_y = m->destination_y + offs_y;
-    map_grid_bound(&dest_x, &dest_y);
-    *x = dest_x;
-    *y = dest_y;
-}
--- a/src/figure/formation_layout.c
+++ b/src/figure/formation_layout.c
@@ -1,5 +1,7 @@
 #include "formation_layout.h"
 
+#include "figure/formation.h"
+
 static const int FORMATION_LAYOUT_POSITION_X[FORMATION_MAX][MAX_FORMATION_FIGURES] = {
     {0, 1, 0, 1, -1, -1, 0, 1, -1, 2, 2, 2, 0, 1, -1, 2},
     {0, 0, -1, 1, -1, 1, -2, -2, 2, 2, -3, -3, 3, 3, -4, -4},
--- a/src/figure/formation_layout.c
+++ b/src/figure/formation_layout.c
@@ -1,6 +1,6 @@
 #include "formation_layout.h"
 
-#include "figure/formation.h"
+#include "map/grid.h"
 
 static const int FORMATION_LAYOUT_POSITION_X[FORMATION_MAX][MAX_FORMATION_FIGURES] = {
     {0, 1, 0, 1, -1, -1, 0, 1, -1, 2, 2, 2, 0, 1, -1, 2},
@@ -41,4 +41,15 @@ int formation_layout_position_x(int layout, int index)
 int formation_layout_position_y(int layout, int index)
 {
     return FORMATION_LAYOUT_POSITION_Y[layout][index];
+}
+
+void formation_layout_position(int layout, int index, formation *m, uint8_t *x, uint8_t *y)
+{
+    int offs_x = formation_layout_position_x(layout, index);
+    int offs_y = formation_layout_position_y(layout, index);
+    int dest_x = m->destination_x + offs_x;
+    int dest_y = m->destination_y + offs_y;
+    map_grid_bound(&dest_x, &dest_y);
+    *x = dest_x;
+    *y = dest_y;
 }
--- a/src/figure/formation_layout.h
+++ b/src/figure/formation_layout.h
@@ -1,8 +1,6 @@
 #ifndef FIGURE_FORMATION_LAYOUT_H
 #define FIGURE_FORMATION_LAYOUT_H
 
-#include "figure/formation.h"
-
 int formation_layout_position_x(int layout, int index);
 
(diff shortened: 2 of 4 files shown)


src/figuretype/animal.c
```diff
@@ -142,10 +142,7 @@ void figure_sheep_action(figure *f)
             if (f->wait_ticks > 400) {
                 f->wait_ticks = f->id & 0x1f;
                 f->action_state = FIGURE_ACTION_197_HERD_ANIMAL_MOVING;
-                f->destination_x = m->destination_x
-                    + formation_layout_position_x(FORMATION_HERD, f->index_in_formation);
-                f->destination_y = m->destination_y
-                    + formation_layout_position_y(FORMATION_HERD, f->index_in_formation);
+                formation_layout_position(FORMATION_HERD, f->index_in_formation, m, &f->destination_x, &f->destination_y);
```
crudelios, line 145: You could just call `map_grid_bound` from here, no?

No need for a whole new function with header includes.

src/figuretype/animal.c
```diff
@@ -197,10 +194,7 @@ void figure_wolf_action(figure *f)
...
```
crudelios, line 197: Same here

src/figuretype/animal.c
```diff
@@ -274,10 +268,7 @@ void figure_zebra_action(figure *f)
             if (f->wait_ticks > 200) {
                 f->wait_ticks = f->id & 0x1f;
                 f->action_state = FIGURE_ACTION_197_HERD_ANIMAL_MOVING;
-                f->destination_x = m->destination_x
-                    + formation_layout_position_x(FORMATION_HERD, f->index_in_formation);
-                f->destination_y = m->destination_y
-                    + formation_layout_position_y(FORMATION_HERD, f->index_in_formation);
+                formation_layout_position(FORMATION_HERD, f->index_in_formation, m, &f->destination_x, &f->destination_y);
```
crudelios, line 271: And here

crudelios:
This is falling the earthquake tests for some reason. We'll need to check if there's a legitimate reason for that or if we should add an exception to the test.

src/figuretype/animal.c
PrettyFlower: @crudelios i didn’t end up changing the type as I didn’t want to muck with the serialization.

crudelios: Fair enough 😉

PrettyFlower: Finally got a chance to look at the earthquake tests. Unsurprisingly, the issue is the destination_x/y for sheep. The new save has a (now bounded) value of 0 and the old destination_x/y is 255.

PrettyFlower: I've gone ahead and added an exception to the test code but let me know if y'all want to go with something else.

src/figure/formation_layout.c
```diff
@@ -42,3 +42,14 @@ int formation_layout_position_y(int layout, int index)
 {
     return FORMATION_LAYOUT_POSITION_Y[layout][index];
 }
+
+void formation_layout_position(int layout, int index, formation *m, uint8_t *x, uint8_t *y)
```
bvschaik, line 46: Using `formation *m` and calling it with a `const formation *` variable gives compiler warnings. Since you don't update the formation, declare the parameter as `const`.

Naming: a formation has multiple coordinates (current, home, standard, destination). When reading code that's calling this function, it's not clear that this function is using the destination, rather than one of the other coordinates. Rename the function name to make that explicit, for example: `formation_layout_get_destination`.

This function is only called from the animal figure code. You mentioned that we might need it for soldiers/enemies as well, but where `formation_layout_position_x/y` are currently used, they're not using destination so we cannot reuse this function anyway. I suggest moving it to animal.c as static function.

src/figure/formation_layout.c
```diff
@@ -42,3 +42,14 @@ int formation_layout_position_y(int layout, int index)
...
+{
+    int offs_x = formation_layout_position_x(layout, index);
+    int offs_y = formation_layout_position_y(layout, index);
+    int dest_x = m->destination_x + offs_x;
+    int dest_y = m->destination_y + offs_y;
```
bvschaik, line 51: The convention in the project is to abbreviate destination to "dst", please stick to that instead of "dest".

src/figure/formation_layout.c
```diff
@@ -42,3 +42,14 @@ int formation_layout_position_y(int layout, int index)
 {
     return FORMATION_LAYOUT_POSITION_Y[layout][index];
 }
+
+void formation_layout_position(int layout, int index, formation *m, uint8_t *x, uint8_t *y)
+{
+    int offs_x = formation_layout_position_x(layout, index);
+    int offs_y = formation_layout_position_y(layout, index);
```
bvschaik, line 49: We're not on a character budget, "offset_x" and "offset_y" are more readable.

src/figuretype/animal.c
bvschaik: The exception looks good to me, let's go ahead with this.

bvschaik, line 145: Line length is a bit too long (> 120 chars), suggestion to break it up like this. Also do the same for the other usages in the zebra and wolf functions.
```suggestion
                formation_layout_position(FORMATION_HERD, f->index_in_formation, m,
                    &f->destination_x, &f->destination_y);
```

bvschaik (changes requested):

src/figure/formation_layout.c
crudelios: I missed these comments!

I try my best but I never think about stuff like this.

src/figuretype/animal.c
PrettyFlower: Since we're only using this for animals now, the FORMATION_HERD argument is unnecessary and the char count goes back down below 120.

src/figure/formation_layout.h
```diff
@@ -1,6 +1,10 @@
 #ifndef FIGURE_FORMATION_LAYOUT_H
 #define FIGURE_FORMATION_LAYOUT_H
 
+#include "figure/formation.h"
+
+#include <stdint.h>
+
```
bvschaik, line 7: These are no longer necessary.

src/figure/formation_layout.c
```diff
@@ -1,6 +1,6 @@
 #include "formation_layout.h"
 
-#include "figure/formation.h"
+#include "map/grid.h"
```
bvschaik, line 3: This can be reverted now the function has been moved to animal.c

bvschaik (commented):
Two more cleanup things, once they're done this can be merged :+1:

bvschaik (approved):

PR #665, opened 2022-12-09
Out of bounds memory read/write bugs during invasions

PrettyFlower:
These can all be reproduced with the -fsanitize=undefined flag on shortly after the invasion starts in the attached save.

[Invasion Test OG.zip](https://github.com/bvschaik/julius/files/10191500/Invasion.Test.OG.zip)

### DIFF
--- a/src/map/figure.c
+++ b/src/map/figure.c
@@ -14,6 +14,13 @@ int map_figure_at(int grid_offset)
     return map_grid_is_valid_offset(grid_offset) ? figures.items[grid_offset] : 0;
 }
 
+static void cap_figures_on_same_tile_index(figure *f)
+{
+    if (f->figures_on_same_tile_index > 20) {
+        f->figures_on_same_tile_index = 20;
+    }
+}
+
 void map_figure_add(figure *f)
 {
     if (!map_grid_is_valid_offset(f->grid_offset)) {
@@ -29,9 +36,7 @@ void map_figure_add(figure *f)
             next = figure_get(next->next_figure_id_on_same_tile);
             f->figures_on_same_tile_index++;
         }
-        if (f->figures_on_same_tile_index > 20) {
-            f->figures_on_same_tile_index = 20;
-        }
+        cap_figures_on_same_tile_index(f);
         next->next_figure_id_on_same_tile = f->id;
     } else {
         figures.items[f->grid_offset] = f->id;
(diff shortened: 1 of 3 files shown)


src/map/figure.c
```diff
@@ -14,6 +14,13 @@ int map_figure_at(int grid_offset)
     return map_grid_is_valid_offset(grid_offset) ? figures.items[grid_offset] : 0;
 }
 
+static void cap_figures_on_same_tile_index(figure *f)
```
crudelios, line 17: Suggestion: instead of capping the value at 20, let it go as high as it wants, then on `adjust_pixel_offset`, for the array index, use `f->figures_on_same_tile_index % 20`.

This has the advantage that it will also work on other eventual spots that also cause the issue.

It will also "centralise" the limit of figures per tile, avoiding the magic number of `20` here.

crudelios (commented):
A small suggestion

src/map/figure.c
PrettyFlower: I was trying to maintain the original logic as much as possible since this is a Julius change. @bvschaik are you ok with @crudelios suggestion?

bvschaik (commented):
I see weird behaviour (also without this change): if I run without the bounds checks compiler flag, the enemies first "stage" in formation near the invasion point, and then move off to destroy the city. If I run with the bounds checks enabled, the first formation does the stage bit, but the others make a beeline for the city along the edge of the map. I think the first behaviour is the correct one; it looks more logical at least.

Also, the horn for announcing another enemy formation coming onto the map is played 91 times, which is quite annoying. I'll look into both issues tomorrow.

Edit: C3 crashes when the invasion starts so we have no reference here...

PrettyFlower:
I see the "beeline for the city" behavior when I run the game without the code changes I made (doesn't seem to matter if I have the compiler flag enabled or not, but I've only been testing with -fsantize=undefined). When I added the code changes, they started using the staging formations. Maybe -fsanitize=address inadvertently fixes the issue somehow.

bvschaik (approved):
The beeline is caused by the out-of-bounds in formation_layout.c: it tries to find a spot on the map where the formation can line up without any obstacles. Since the >16 soldiers are getting an invalid index for the next formation type, and the >32 soldiers index out of the formation_layout array entirely, the game cannot find a spot, causing the beeline. With bounds checking (-fsantize=undefined), there's most likely a different bit of memory beyond 32 than without the bounds checking.

Since all things now have a logical explanation, this looks good to me 👍

PR #667, opened 2022-12-16
Use -fsanitize=undefined when compiling with Clang.

PrettyFlower:
I tried upgrading to VS 2022 and it stopped reporting every single for loop as an error, so I propose we go ahead and set this flag by default when in Debug mode. Might want to do it for GCC as well.


crudelios:
Just a question as I haven't tested with the option on yet: if there's an out of bounds with this enabled, does the game immediately crash?

PrettyFlower:
Not with this setup (tho you can configure it to work that way).

crudelios:
Okay nice. Then I guess it's OK to add this option.

bvschaik (approved):

PR #669, opened 2023-01-05
Fix bathhouse water preview

b483:
Fix image offset when drawing bathhouse preview (ghost) version filled with water (near reservoir)


bvschaik (approved):
Nice one! I just checked and this bug is also present in C3.

PR #670, opened 2023-01-06
Enable display scaling on Android tablets

bvschaik:
@crudelios Could you perhaps check if this breaks anything on Vita/Switch?

### DIFF
--- a/src/platform/screen.c
+++ b/src/platform/screen.c
@@ -243,10 +243,11 @@ int system_can_scale_display(int *min_scale, int *max_scale)
     int width, height;
     SDL_GetWindowSize(SDL.window, &width, &height);
 #ifdef __ANDROID__
-    int maxOrientation1 = get_max_scale_percentage(width, height);
-    int maxOrientation2 = get_max_scale_percentage(height, width);
-    if (maxOrientation1 < 100 && maxOrientation2 < 100) {
-        SDL_Log("Not allowing scale on Android: %d x %d = max scale %d or %d", width, height, maxOrientation1, maxOrientation2);
+    int max_scale_current_orientation = get_max_scale_percentage(width, height);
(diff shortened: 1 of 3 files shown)


src/platform/screen.c
```diff
@@ -230,14 +230,33 @@ int system_scale_display(int display_scale_percentage)
...
+#ifndef __ANDROID__
+    if (system_is_fullscreen_only()) {
+        return 0;
+    }
+#endif
     int width, height;
     SDL_GetWindowSize(SDL.window, &width, &height);
-    return get_max_scale_percentage(width, height);
+#ifdef __ANDROID__
+    int maxOrientation1 = get_max_scale_percentage(width, height);
```
crudelios, line 246:
```suggestion
    int max_orientation_portrait = get_max_scale_percentage(width, height);
```

src/platform/screen.c
```diff
@@ -230,14 +230,33 @@ int system_scale_display(int display_scale_percentage)
...
+    int maxOrientation2 = get_max_scale_percentage(height, width);
```
crudelios, line 247:
```suggestion
    int max_orientation_landscape = get_max_scale_percentage(height, width);
```

crudelios (commented):
It should work but I'll check,

Anyway, here's a small oversight I noticed.

src/platform/screen.c
bvschaik: Since we don't know whether the current orientation is portrait or landscape I opted for simply "1" and "2".

crudelios: Makes sense. What about `max_orientation_current` and `max_orientation_alternative` though?

bvschaik: Good suggestion. I opted for `max_scale_current_orientation` and `max_scale_alternative_orientation` to make it clear we're dealing with a maximum scale.

PR #675, opened 2023-01-22
Windows: use directsound by default instead of wasapi

crudelios:
Fixes many audio issues still present in sdl


crudelios:
Using directsound still has some drawbacks though, like increased latency and less compatibility with multiple outputs.

Another possible change would be to allow the user to select their preferred audio engine in the config window.

PR #677, opened 2023-02-18
Fix military academy understaffed text

b483:
Fix building info text for military academy when below 100% employment


bvschaik (approved):
Thanks!

Issue #689, opened 2023-05-15
hard to improve some Chinese translation


feiyunw:
It's hard to improve some Chinese translation.

Per the source:
```c
// src/core/encoding_simp_chinese.c:
static const chinese_entry codepage_to_utf8[IMAGE_FONT_MULTIBYTE_SIMP_CHINESE_MAX_CHARS] = {
    {0x8080, {0xe6, 0xa1, 0xa3}},
...
    {0x90d1, {0xe5, 0x87, 0xb8}},
};

// src/core/encoding_trad_chinese.c:
static const chinese_entry codepage_to_utf8[IMAGE_FONT_MULTIBYTE_TRAD_CHINESE_MAX_CHARS] = {
    {0x8080, {0xef, 0xbc, 0x81}},
...
    {0x918b, {0xe5, 0xbe, 0xb9}}
};
```
It uses a very customized encoding system for Chinese ideograph, which starts coded characters from 0x8080.
None of the popular Chinese encoding systems uses such encoding scheme.
Thus this game requires a special tool to subset and reorganize the official standard (as GB 18030 for Simplified Chinese) coded characters, and a customized font system is used to display such characters.

For Simplified Chinese, this customized encoding system is used in the files loaded by /src/core/lang.c: load_text().
It prevents muggles from translation improvement work.
To rectify this problem, we should abandon this encoding system for translation text at least, where UTF-8 (or GB 18030 for the Simplified Chinese case) will be good.
It'd be better to put translation in a plain text file, or some script which supports UTF-8 like Lua.

A modern font system will be helpful then, and it will be another topic.

FYI:
The double bytes Simplified Chinese ideograph defined by GB 18030 begins from 0x8140 and has 21008 coded characters.

[GB 18030](https://openstd.samr.gov.cn/bzgk/gb/std_list?p.p1=0&p.p90=circulation_date&p.p91=desc&p.p2=18030), GB 18030-2022 Page 5 Fig.3
[Output GB 18030-2022 repertoire by Lua](https://blog.csdn.net/feiyunw/article/details/83551768)
[pyftsubset](https://fonttools.readthedocs.io/en/latest/subset)
[Region Specific Subset OTFs Simplified Chinese](https://github.com/adobe-fonts/source-han-sans/releases)

crudelios:
The problem with abandoning the encodig system is that the original translation files rely on it to work.

We can't distribute the original translation files, even as a new format, as that would violate copyright.

[...]

PR #690, opened 2023-05-16
Add dump_lang_data() to log translated texts and messages

feiyunw:
This PR add a debug helper function: dump_lang_data.
It logs src/core/lang.c data content to help checking texts and messages validity.
It's called only when you change the language setting in the Options menu, thus the log won't appear normally.

It helps #689 to identify corrupted/duplicate/empty translation text and messages.

### DIFF
--- a/src/core/lang.c
+++ b/src/core/lang.c
@@ -1,12 +1,14 @@
 #include "core/lang.h"
 
 #include "core/buffer.h"
+#include "core/encoding.h"
 #include "core/file.h"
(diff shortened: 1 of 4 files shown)


bvschaik:
You should use [EngConverter](https://github.com/bvschaik/citybuilding-tools/releases/tag/engconverter-v0.5) for this purpose.

Julius is a production game. Having a bit of code that logs several megabytes of data to a file is not acceptable.

PR #695, opened 2023-06-16
Haiku OS: Remove no longer needed Haiku-specific header

brownna01:
Previously I had created a Haiku-specific header file that only specified that SDL2 should use a software cursor.

Patches to the Haiku port of SDL2 have since rendered this unnecessary, so remove the header and all references to it. Hardware cursors works now, as on other platforms.

See https://discourse.libsdl.org/t/sdl-sdl-updated-patches-for-haiku/30888 for the SDL2 changes.


bvschaik (approved):
Thanks!

PR #700, opened 2023-08-04
Support Greek fan translation

bvschaik:


crudelios (approved):
This is looking good to me!

PR #707, opened 2023-11-01
Add current difficulty to sidebar infos

Gheeksy:
**Context:**
We're trying to set up a speedrun community, in order to make run submission / validation easier, we would love to have more informations available to the screen, such as the current game difficulty.

I didn't really know where to display that, so I added it to the sidebar, but I could change that if you think there's a better place for it. Also, if that's needed, I could try to make it a setting.

**Changes:**
- Add current difficulty level to the sidebar

**Screenshots:**
julius_IV9y2CuIjp


crudelios:
Looks fine to me, except for two comments:

There should be a little higher vertical margin between the win criteria and the difficulty, the distance should be the same as in between "Unemployment" and "Culture".

Also, there are native translations for the "difficulty" and each difficulty level string, you can use them directly. That way it will automatically have a translation to every existing language. I can help you out with that if you want.

Gheeksy:
Hey, thanks for the review!

I just fixed to vertical margin issue, I hope that's what you were asking for.

julius_H1zqL4MwaY

Concerning the translations, I did not manage to understand how to use the original game's strings.
I would love your help on that (maybe telling me where to search could be enough).

bvschaik:
Before going any further: generally we only accept PRs with features that are useful for the majority of the users. For this PR, I don't see the added value for a regular player: difficulty is a global setting (not per saved game), that you tend to set once and then forget about.

That said, how do those speedrun contests work? Speed as in game-time or wallclock-time? If you want to quickly check a .sav file, this PR doesn't make sense since it's a global setting. Or does this mean the whole play is videotaped and you want to confirm that the user doesn't secretly change the difficulty halfway through the play?
Also be aware that the info sidebar doesn't show the difficulty if the screen is not tall enough, and of course it doesn't show if the sidebar is collapsed, so another solution may be better.

As for the translations: you can use the `lang_draw_text(group, number, ...)` function to draw game text. You can look up the `group` and `number` to use in `src/window/difficulty_options.c`, which is the difficulty options dialog. Group for difficulty-related strings is `153`.

QuakeIV:
generally speedrunners produce a recording, yes

the game ideally will produce some diagnostic information to see they are not cheating, and you can do noise analysis comparisons to try to detect edits to the recording

its not perfect or anything (arguably if you can figure out how to add it to the game you can also falsify them and recompile) but thats what the guy is asking for

Gheeksy:
> Before going any further: generally we only accept PRs with features that are useful for the majority of the users. For this PR, I don't see the added value for a regular player: difficulty is a global setting (not per saved game), that you tend to set once and then forget about.
> 
> That said, how do those speedrun contests work? Speed as in game-time or wallclock-time? If you want to quickly check a .sav file, this PR doesn't make sense since it's a global setting. Or does this mean the whole play is videotaped and you want to confirm that the user doesn't secretly change the difficulty halfway through the play? Also be aware that the info sidebar doesn't show the difficulty if the screen is not tall enough, and of course it doesn't show if the sidebar is collapsed, so another solution may be better.
> 
> As for the translations: you can use the `lang_draw_text(group, number, ...)` function to draw game text. You can look up the `group` and `number` to use in `src/window/difficulty_options.c`, which is the difficulty options dialog. Group for difficulty-related strings is `153`.

Yes, I totally understand your concern about the majority of players.
Could disabling this info by default be a solution? Making it a setting like "speedrun infos" to check somewhere?

Yes the speedrun has to be video recorded, but requiring the players to show their settings at the start of each run would not be acceptable. The time we count is real time spent from the start of the run, until the victory screen.
Asking speedrunners to run the game at a low enough resolution to see those informations wouldn't be a problem.

Also we know it's not 100% sure the player didn't edit its files, but that would still help us for the majority of the runs to have this information on screen.
I guess we could probably check the real difficulty based on certain in-game events if we have a doubt on a specific player / run?

Thanks for the infos about translations, I'll take a look.

QuakeIV:
what if you made it a compilation option rather than a checkbox, so you can then distribute your approved speedrunning build?

Gheeksy:
> what if you made it a compilation option rather than a checkbox, so you can then distribute your approved speedrunning build?

Wouldn't that require each runner to install the development dependencies ?
That surely could be a solution, but I'm afraid it would make it much less accessible to the community.

But I can totally understand your choice if you think this shouldn't be included in the standard build.

QuakeIV:
by distribute, i mean you guys could compile your approved speed running build and then provide the files, and tell the speedrunners thats the version they need to run with.  so it gets added to master, its just not there by default

i dont know if that works for you or not its just an idea that might make it a little more inoffensive to the dev team.  im a bit on the sidelines to be clear, im not really active in this project just a fan of it

bvschaik:
A compiler flag is an option, disadvantage of that is that you'd need to create a speedrun build, for all operating systems, and for something like macOS that's not trivial to do.
Another option is to make it a "hidden" setting that can only be enabled by editing the `julius.ini` file.

As for showing the information, how about a small label in the bottom left corner of the city?
Something like this perhaps?

difficulty_overlay

Gheeksy:
Thanks for your answers, the `julius.ini` suggestion seems to be the best for everyone.
Maybe we should close this, and I'll tag this pull request when it's ready?

bvschaik:
> Thanks for your answers, the `julius.ini` suggestion seems to be the best for everyone. Maybe we should close this, and I'll tag this pull request when it's ready?

Sounds good to me. Just for clarity: are you going to work on this?

Gheeksy:
> > Thanks for your answers, the `julius.ini` suggestion seems to be the best for everyone. Maybe we should close this, and I'll tag this pull request when it's ready?
> 
> Sounds good to me. Just for clarity: are you going to work on this?

Yes I am, but maybe it will take a few days because I'm often going in the wrong direction at first.
I'm not used at all to this kind of project nor this language.

I'll close this PR for now.

PR #708, opened 2023-11-08
use devkitPro cmakery

WinterMute:
Use devkitPro cmakery to build project for switch.


bvschaik:
Please also update the github actions scripts in `.ci_scripts`.

WinterMute:
Sorry, I hadn't noticed the old docker tag which doesn't contain the full cmakery scripts.

The advantage over the current approach is it's using the [extensive cmake support](https://github.com/devkitPro/pacman-packages/tree/master/cmake) we maintain for the platforms we target. This hopefully provides more examples of using our infrastructure which more homebrewers can take advantage of which hopefully means we find standard processes propagating and people experiencing less issues building homebrew projects.

Standardisation and flattening of build environments helps everyone in the end.

There are some more tweaks I'd like to make - for instance using pkg-config to find libraries rather than specifying them manually should avoid issues if & when we update dependencies for SDL.

WinterMute:
What's up with the last two workflows? What's triggered those?

image

bvschaik:
> What's up with the last two workflows? What's triggered those?

Both are triggered by you pushing to your branch. The workflows have to be approved by one of the maintainers before they're run, since this is your first contribution to Julius, hence the warning icon.

Thanks for the explanation, I'll take a closer look this evening.

CMakeLists.txt
```diff
@@ -39,6 +34,10 @@ set(SHORT_NAME julius)
 set(USER_FRIENDLY_NAME Julius)
 project(${SHORT_NAME} C)
 
+if (NINTENDO_SWITCH)
```
bvschaik, line 37: Where does this `NINTENDO_SWITCH` come from?

CMakeLists.txt
```diff
@@ -740,7 +739,7 @@ if(SYSTEM_LIBS)
 endif()
 if(PNG_FOUND)
     include_directories(SYSTEM ${PNG_INCLUDE_DIRS})
-    target_link_libraries(${SHORT_NAME} ${PNG_LIBRARIES})
+    target_link_libraries(${SHORT_NAME} PRIVATE ${PNG_LIBRARIES})
```
bvschaik, line 742: Apparently the Vita build and CodeQL don't like this. Any particular reason you added it?

CMakeLists.txt
WinterMute: This comes from our cmake platform configuration https://github.com/devkitPro/pacman-packages/blob/master/cmake/switch/NintendoSwitch.cmake#L11

CMakeLists.txt
WinterMute: When I added the target_link_libraries for Switch I set it to PRIVATE then got an error about mixing signatures as per https://cmake.org/cmake/help/latest/policy/CMP0023.html

We can either set all of these to PRIVATE or just leave them unlabelled. I think the latter is probably the least intrusive here but it might be worth thinking about.

bvschaik: Since Julius is an executable and not a library, there's nothing to be propagated. But we should indeed consider making everything `PRIVATE` at some point. For now, leaving it unlabelled is indeed fine.

bvschaik (approved):

WinterMute:
Thanks for merging.

crudelios:
@WinterMute This seems like a good addition.

Do you mind also doing it for [Augustus](https://github.com/Keriew/augustus), please?

It has the added challenge that it needs to add files to romfs on build.

PR #712, opened 2023-12-20
czech.c: Fix overlapping text and remove some translation bloat

gryffus:
- Fix some text overlapping buttons.
- Remove bloat


gryffus:
Should fix things like this:
obrazek
obrazek

gryffus:
I will just submit all in 1 commit.

Closing

PR #714, opened 2023-12-28
Improve metainfo

Alexander-Wilms:
* Update SPDX license identifier
* Sort tags with jdAppStreamEdit
* Use new `<developer>` syntax
* Reword description
* Add German translation of description
* Change help URL to a link to the Github wiki

Metainfo file validated using
`flatpak run org.freedesktop.appstream.cli validate --pedantic  res/julius.metainfo.xml`

### DIFF
--- a/res/julius.metainfo.xml
+++ b/res/julius.metainfo.xml
@@ -1,24 +1,30 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<component type="desktop-application">
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+	<!--Created with jdAppStreamEdit 8.0-->
 	<id>com.github.bvschaik.julius</id>
-	<metadata_license>CC0-1.0</metadata_license>
-	<project_license>AGPL-3.0</project_license>
 	<name>Julius</name>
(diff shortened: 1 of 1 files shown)


bvschaik:
Thanks!

Question regarding the German translation: I see that every element (p, li) now has an English version without language and one directly below for German. How does that work in practice? Don't you get the two languages mixed since the one without language is always displayed, also for German? And isn't there a better way, for example having multiple `<description>` tags with different languages, like you did for the summary? If we add 10 languages, the description will become an unmaintainable mix of languages with the current setup...

Alexander-Wilms:
No, if you use e.g. KDE Discover or Gnome Software on a system with German locale, only the elements with the xml:lang="de" property will be shown.

Cf. the Metainfo spec:

> In metainfo files, this tag should be translated by-paragraph, meaning that in a translated file, each translated <p/> child has an xml:lang property. 

https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-description

I think the idea is that not the entire translation has to be re-checked when the English description changes.

bvschaik:
Thanks for the explanation, looks good to me then.

bvschaik (approved):

PR #718, opened 2024-01-31
Enhancement for sidebar unemployment

feiyunw:
This is a UI enhancement suggestion.
It is very timing critical to build structures.
A safe way to do so is to build the structure when there are enough jobless workers for it.
However, the unemployment amount, shown in the sidebar, does not fit this need very well.
The labor seekers need some time covering his service to the houses; thus, the unemployment amount is not a good indicator for consecutively building new structures.
To solve this problem, I present this modification: to show the unemployment rate title in different color, indicating whether all buildings that should have workers allocated do have them.

### DIFF
--- a/src/city/data_private.h
+++ b/src/city/data_private.h
@@ -183,6 +183,7 @@ extern struct city_data_t {
         int32_t graph_order;
     } population;
     struct {
+        int32_t all_have_workers;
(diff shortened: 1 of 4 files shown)


bvschaik:
Closing as won't fix, see related issue #719 for rationale.

PR #725, opened 2024-03-17
Add Flatpak support for Linux

tintou:
Allows to use Flatpak as the way to play the game on Linux.

### DIFF
--- a/.ci_scripts/run_build.sh
+++ b/.ci_scripts/run_build.sh
@@ -20,6 +20,10 @@ case "$BUILD_TARGET" in
 	echo "Creating disk image"
 	hdiutil create -volname Julius -srcfolder julius.app -ov -format UDZO julius.dmg
 	;;
+"flatpak")
+	flatpak-builder repo com.github.bvschaik.julius.json --install-deps-from=flathub --keep-build-dirs
+	cp .flatpak-builder/build/julius/res/version.txt res/version.txt
+	;;
 "appimage")
 	cd build && make -j4 && make test
 	make DESTDIR=AppDir install
--- a/.ci_scripts/run_cmake.sh
+++ b/.ci_scripts/run_cmake.sh
@@ -10,6 +10,8 @@ case "$BUILD_TARGET" in
 "mac")
 	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DSYSTEM_LIBS=OFF -DCMAKE_OSX_ARCHITECTURES="x86_64;arm64" ..
 	;;
+"flatpak")
+	;;
 "appimage")
 	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DSYSTEM_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/usr ..
 	;;
--- a/com.github.bvschaik.julius.json
+++ b/com.github.bvschaik.julius.json
@@ -1,0 +1,28 @@
+{
+    "id" : "com.github.bvschaik.julius",
+    "runtime" : "org.freedesktop.Platform",
+    "runtime-version" : "23.08",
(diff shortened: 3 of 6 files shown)


.ci_scripts/run_build.sh
```diff
@@ -20,6 +20,10 @@ case "$BUILD_TARGET" in
 	echo "Creating disk image"
 	hdiutil create -volname Julius -srcfolder julius.app -ov -format UDZO julius.dmg
 	;;
+"flatpak")
+	flatpak-builder repo com.github.bvschaik.julius.json --install-deps-from=flathub --keep-build-dirs
+	cp .flatpak-builder/build/julius/res/version.txt res/version.txt
```
bvschaik, line 25: Why do we need this copy action? `res/version.txt` should be generated by CMake -- I'm guessing that flatpack-builder uses cmake under the hood?

.ci_scripts/run_cmake.sh
```diff
@@ -10,6 +10,8 @@ case "$BUILD_TARGET" in
 "mac")
 	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DSYSTEM_LIBS=OFF -DCMAKE_OSX_ARCHITECTURES="x86_64;arm64" ..
 	;;
+"flatpak")
```
bvschaik, line 13: How is cmake invoked for Flatpak?

bvschaik (commented):
As a total noob when it comes to these kind of packages for Linux, could you explain how adding this will work together or conflict with [this](https://github.com/flathub/com.github.bvschaik.julius)?

.ci_scripts/run_build.sh
tintou: Yes it is using cmake under the hood but it is building it out of tree (in `./.flatpak-builder/build/julius`) and thus isn't being found afterwards by the `build_upload.sh` script

.ci_scripts/run_cmake.sh
tintou: `flatpak-builder repo com.github.bvschaik.julius.json` in `run_build.sh` will call cmake (with ninja instead of make) as it is defined as `"buildsystem" : "cmake-ninja"`

tintou:
> As a total noob when it comes to these kind of packages for Linux, could you explain how adding this will work together or conflict with [this](https://github.com/flathub/com.github.bvschaik.julius)?

The manifest within Flathub is just for it to be published in the Flathub store. Including a manifest in-tree allows anyone to build the software as a flatpak (here I have even IDE integration that will natively allow me to build it in one click).
There won't be a conflict with the one from Flathub as only one manifest is used so the manifest at Flathub will download the source but ignore the one in-tree.

If you compare both manifests, they are really the same, Flathub just requires to specify a version for reproducibility while the manifest in-tree points to the latest commit from the "master" branch :)

bvschaik (approved):
Thanks for the explanation, and the PR!

com.github.bvschaik.julius.json
```diff
@@ -0,0 +1,28 @@
...
+        {
+            "name" : "julius",
+            "buildsystem" : "cmake-ninja",
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/bvschaik/julius.git",
+                    "branch" : "master"
+                }
+            ]
```
bvschaik, line 25: Looking closer at the build output, I see this happening:
1. Git repo is checked out for the build as usual
2. flatpak-builder is invoked
3. Git repo is checked out _again_ by the builder

This means that if you do some changes locally, then build a .flatpak, your local changes are not used at all because the sources it builds are the latest master from github... how can we make it use the local sources?

tintou: You're right, I've opened https://github.com/bvschaik/julius/pull/726 to fix this

PR #729, opened 2024-04-20
Improve tooltip looks

crudelios:
Before:

tooltip_before

After:

tooltip_after

### DIFF
--- a/src/graphics/text.c
+++ b/src/graphics/text.c
@@ -421,7 +421,9 @@ int text_draw_multiline(const uint8_t *str, int x_offset, int y_offset, int box_
 
 int text_measure_multiline(const uint8_t *str, int box_width, font_t font, int *largest_width)
 {
-    *largest_width = 0;
+    if (largest_width) {
+        *largest_width = 0;
+    }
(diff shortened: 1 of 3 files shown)


src/graphics/text.c
```diff
@@ -419,8 +419,9 @@ int text_draw_multiline(const uint8_t *str, int x_offset, int y_offset, int box_
     return y - y_offset;
 }
 
-int text_measure_multiline(const uint8_t *str, int box_width, font_t font)
+int text_measure_multiline(const uint8_t *str, int box_width, font_t font, int *largest_width)
 {
+    *largest_width = 0;
```
bvschaik, line 424: Strictly speaking `largest_width` is an optional parameter so it can be null. I know we only use this in one place (the tooltips), but let's be safe and add a null-check around accessing it.

bvschaik (changes requested):
Somehow I missed this PR... Code looks good in general, one safety comment.

crudelios:
I finally updated this! It's ready to merge.

bvschaik (approved):

Issue #643, opened 2022-02-27
iOS Support planned?


eduo:
Ask your question here.
I was wondering if iOS or iPadOS is in the plans at all. With Mac compilation done, it may be closer to possible and with the recently improved support for keyboard and mouse it's not even necessary to support touch 100%.

crudelios:
There are no plans to support iOS.

Supporting iOS not only requires a developer license which costs about 100 dollars a year, but the Apple Store's terms of use are not compatible with Julius' license (GPL v3).

eduo:
> There are no plans to support iOS.
> 
> Supporting iOS not only requires a developer license which costs about 100 dollars a year, but the Apple Store's terms of use are not compatible with Julius' license (GPL v3).

Just to clarify: Supporting iOS doesn't require a paid license nor does it require violating the GPL. I think you're referring to publishing in the App Store, which I agree with.

I skipped it because I thought it was obvious, my mistake. I meant for sideloading in the same way as spiritually similar projects like Exult or DevilutionX (or ScummVM or RetroArch or ppsspp or many others) currently support compilation for iOS/iPad and in some cases even produce a ready-made IPA that can be sideloaded without ever paying a developer license or publishing in the app store.

I'm sorry I wasn't clear enough. I understand why projects like these wouldn't want to get anywhere near Apple's App Store and I fully support that decision. I meant something else.

eduo:
I appreciate the time taken to give a response. I asked precisely because I saw Android support and guessed most of the UX part could be translated (I understand this is but a small part of the work that would have to be done).

I also asked because I have friends interested that might have the chops necessary to help with such a port. I'll see about them creating a fork and if successful will tell them to make a PR.

Thanks for your efforts and work.

bvschaik:
Before any work is done, I'd like to have a few points clarified. Keep in mind: I did a bit of iOS programming more than 10 years ago (iOS 4/5), so my knowledge is limited and quite outdated...

1. How does this sideloading stuff / install apps without the app store work? How should users install the app and how should we provide it?
2. What about the assets from the original game? Android provides a real filesystem outside the app data, which the user can access and copy things onto from a computer. Within the app you then grant permission to this directory so the game can use it. I don't remember iOS having a similar concept: all data is stored in the app itself: how does the user get the assets there, and the savegames/maps out?
[...]

PR #743, opened 2024-12-27
Add iOS build support

axmo:
This adds support for running on (and cross-compiling to) iOS devices, as described in https://github.com/bvschaik/julius/issues/643#issuecomment-2542390491

See `ios/README.md` for build instructions.

### DIFF
--- a/ios/README.md
+++ b/ios/README.md
@@ -1,0 +1,23 @@
+# iOS instructions
+
+Download SDL2 and SDL2_mixer source releases and put them in `ios/SDL2` and `ios/SDL2_mixer` respectively.
+
+run the following commands:
+
+```
+	$ mkdir build && cd build
+	$ cmake .. -DTARGET_PLATFORM=ios -G Xcode
+```
+
+Open the resulting Xcode Project file `julius.xcodeproj`
+
+## Build for Simulator
+Choose `julius` in the Build Schemes dropdown, choose a Simulator from the Run Destinations dropdown, and click Run.
+
+## Build for Device
+With the Augustus project and target selected, set a Bundle Id, and choose a Development Team, then choose the `julius` in the Build Schemes dropdown, choose a device from the Run Destinations dropdown, and click Run.
+
+## Running
+AirDrop or otherwise download a C3 install folder to your iOS device.
+
+Launch Julius. Immediately after the first launch, Julius will display a file-picker dialog.  Navigate to your C3 folder, and choose 'Open'. The C3 files will be *copied* to Julius's application documents directory, and the game will use those files from then on. Use the default location for saved games.
--- a/src/platform/ios/GameDataPickerController.h
+++ b/src/platform/ios/GameDataPickerController.h
@@ -1,0 +1,11 @@
+#import <UIKit/UIKit.h>
+#import "SDL.h"
+#import "SDL_syswm.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GameDataPickerController : NSObject<UIDocumentPickerDelegate>
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/src/platform/ios/GameDataPickerController.m
+++ b/src/platform/ios/GameDataPickerController.m
@@ -1,0 +1,47 @@
+#import "GameDataPickerController.h"
+
+#import "ios.h"
+#import "SDL.h"
+#import "SDL_syswm.h"
+#import "SDL_video.h"
+#import "SDL_system.h"
+
+#import "platform/screen.h"
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>xw
+
+@implementation GameDataPickerController
+
+NSURL *url;
+
+- (void)documentPicker:(UIDocumentPickerViewController *)controller didPickDocumentsAtURLs:(NSArray<NSURL *> *)urls {
+    
+    NSLog(@"urls: %@", urls);
+    NSURL *importDirectory = urls.firstObject;
+    [importDirectory startAccessingSecurityScopedResource];
+    
+    NSURL *gameDataDirectory = [[[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject] URLByAppendingPathComponent:@"C3" isDirectory:YES];
+    
+    NSError *error = nil;
+    BOOL success = false;
+    success = [[NSFileManager defaultManager] copyItemAtPath:importDirectory.path toPath:gameDataDirectory.path error:&error];
+    
+    if(success) {
+        NSLog(@"Files copied to: %@", gameDataDirectory);
+    } else {
+        NSLog(@"Error copying data files: %@", error);
+        return;
+    }
+    
+    [importDirectory stopAccessingSecurityScopedResource];
+    
+    char path[300];
+    [gameDataDirectory.path getFileSystemRepresentation:path maxLength:300];
+    c3_path_chosen(path);
+}
+
+- (void) documentPickerWasCancelled:(UIDocumentPickerViewController *)controller {
+    
+}
+
+@end
+
--- a/src/platform/ios/ios.h
+++ b/src/platform/ios/ios.h
@@ -1,0 +1,6 @@
+#include <TargetConditionals.h>
+
+const char* ios_show_c3_path_dialog(int again);
+void c3_path_chosen(char *new_path);
+const char* ios_get_base_path(void);
+int ios_set_base_path(const char *path);
--- a/src/platform/ios/ios.m
+++ b/src/platform/ios/ios.m
@@ -1,0 +1,75 @@
+#import "ios.h"
+#import "SDL.h"
+#import "SDL_syswm.h"
+#import "SDL_video.h"
+#import "SDL_system.h"
+#import "SDL_messagebox.h"
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+
+#import "GameDataPickerController.h"
+
+#define GAME_PATH_MAX 300
+static int has_directory;
+static char path[GAME_PATH_MAX];
+
+void c3_path_chosen(char *new_path) {
+    unsigned long len = strlen(new_path);
+    strncpy(path, new_path, len);
+    has_directory = true;
+}
+
+const char* ios_get_base_path(void) {
+    NSURL *gameDataDirectory = [[[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject] URLByAppendingPathComponent:@"C3" isDirectory:YES];
+    
+    [gameDataDirectory.path getFileSystemRepresentation:path maxLength:300];
+    
+    return path;
+}
+
+int ios_set_base_path(const char *path) {
+    CFURLRef url = (__bridge CFURLRef)[NSURL fileURLWithPath: [NSString stringWithCString:path]];
+    
+    return 1;
+}
+
+
+const char *ios_show_c3_path_dialog(int again)
+{
+    UIWindow *window = nil;
+    UIWindow *alertwindow = nil;
+
+    GameDataPickerController *delegate = [[GameDataPickerController alloc] init];
+    UIDocumentPickerViewController *picker = [[UIDocumentPickerViewController alloc] initForOpeningContentTypes: @[UTTypeFolder]];
+    
+    picker.delegate = delegate;
+
+    NSError *error = nil;
+    picker.directoryURL = [[NSFileManager defaultManager] URLForDirectory:NSDocumentDirectory inDomain:NSUserDomainMask appropriateForURL:nil create:YES error:nil];
+
+    if (error) {
+        // Handle the error.
+    }
+
+    if (window == nil || window.rootViewController == nil) {
+        alertwindow = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+        alertwindow.rootViewController = [UIViewController new];
+        alertwindow.windowLevel = UIWindowLevelAlert;
+
+        window = alertwindow;
+
+        [alertwindow makeKeyAndVisible];
+        
+    }
+
+    [window.rootViewController presentViewController:picker animated:YES completion:nil];
+    
+    while (!has_directory) {
+        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate distantFuture]];
+    }
+
+    if (alertwindow) {
+        alertwindow.hidden = YES;
+    }
+    
+    return path;
+}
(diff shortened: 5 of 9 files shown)


ios/README.md
```diff
@@ -0,0 +1,23 @@
...
+	$ cmake .. -DTARGET_PLATFORM=ios -G Xcode
+```
+
+Open the resulting Xcode Project file `julius.xcodeproj`
+
+## Build for Simulator
+Choose `julius` in the Build Schemes dropdown, choose a Simulator from the Run Destinations dropdown, and click Run.
+
+## Build for Device
+With the Augustus project and target selected, set a Bundle Id, and choose a Development Team, then choose the `julius` in the Build Schemes dropdown, choose a device from the Run Destinations dropdown, and click Run.
```
crudelios, line 18:
```suggestion
With the Julius project and target selected, set a Bundle Id, and choose a Development Team, then choose the `julius` in the Build Schemes dropdown, choose a device from the Run Destinations dropdown, and click Run.
```

src/platform/ios/ios.m
```diff
@@ -0,0 +1,75 @@
...
+
+#import "GameDataPickerController.h"
+
+#define GAME_PATH_MAX 300
+static int has_directory;
+static char path[GAME_PATH_MAX];
+
+void c3_path_chosen(char *new_path) {
+    unsigned long len = strlen(new_path);
+    strncpy(path, new_path, len);
```
crudelios, line 17: I don't think I've ever done this in Julius, but only in Augustus, but still, `snprintf` Is safer.

Also, if `len` is larger than `GAME_PATH_MAX`, you got an issue.

```suggestion
    snprintf(path, GAME_PATH_MAX, "%s", new_path);
```

crudelios (commented):
Thanks for this!

Here are my small suggestions.

src/platform/ios/ios.m
axmo: good point, thanks!

crudelios:
I'm also thinking: instead of adding the sources to `ios/SDL2`, can't we use the existing `ext/SDL2`?

That way the directory convention would be consistent in all builds.

axmo:
> I'm also thinking: instead of adding the sources to `ios/SDL2`, can't we use the existing `ext/SDL2`?
> 
> That way the directory convention would be consistent in all builds.

We could!  For iOS I'm building SDL and mixer from source, and I wasn't sure if the other platforms using `ext/SDL2` were expecting the source or binary releases there, which is why I put it under `ios/SDL2`. (Where are other platforms finding SDL2_mixer?)

crudelios:
What we usually do is use `ext/SDL2/SDL2-2.X.X` and `ext/SDL2/SDL2_mixer-2.X.X`.

Things get a bit messy if you're trying to configure for different platforms from the same base directory, but with some effort it works and, more importantly, it works just fine for github actions.

Speaking of which, the test build is failing. I think it's an include in `julius.c` which should be inside some`#ifdef`s but isn't. Can you check please?

Also: do you think you can add a github actions workflow to automate ios builds? If not, can you at least provide command line instructions here for building the app so we later create the workflow ourselves?

Thanks!

axmo:
Ok. for iOS I'm adding the SDL2 and SDL2_mixer CMake projects with `add_subdirectory`, as suggested in this [libSDL forum post](https://discourse.libsdl.org/t/state-of-cmake-ios-for-sdl2/34509/4), so I'll use `ext/SDL2/SDL2/...` and `ext/SDL2/SDL2_mixer/...`

Automated iOS builds for distribution would require putting a signing certificate and provisioning profile on GitHub, which I'm not prepared to do myself, but if you just want to check that it builds correctly, or run tests in a simulator I guess that's doable. A warning: GitHub actions running on macOS cost 10x the price of GitHub actions on Linux, and in my experience builds on GitHub take ages. (that was a few years ago, maybe it has gotten better?)

crudelios:
Yeah for now we just want to check that it builds.

The macOS action is free, our mac build is done using it. I think we use a self signing certificate.

axmo:
Ignore previous stuff about the ext/SDL/... location, I've taken a look at /.ci_scripts/ and am adding an `install_sdl_ios`, which will follow the same directory structure as the others.

axmo:
Hm. Debug builds for simulator and for device work, but creating a release archive (used for App Store or Ad-Hoc distribution) causes an error:

`Wrong SDL_config.h, check your include path?`, which seems to be because `USING_GENERATED_CONFIG_H` is set, but I'm not sure why it would only be set for release builds.

crudelios:
Did you try deleting the SDL2 folder (and adding it brand new) and all intermediate build caches on your PC?

Android studio also has a lot of problems rebuilding, and deleting everything to start fully clean usually fixes them.

axmo:
Aha. looks like I had multiple copies of SDL2 and SDL2_mixer in `ext/SDL2` after running `install_dependencies`, which were all being copied to `build/ext` and then due to header search path precedence, it was loading the wrong SDL_config.h as a result.

All seems to be working now, and I've added the necessary `run_cmake` and `run_build` scripts

axmo:
Please let me know if anything else is needed for this to be considered ready to merge.

src/platform/julius.c
```diff
@@ -28,6 +28,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "platform/ios/ios.h"
```
crudelios, line 31: Includes in alphabetical order please.

Other than that,  @bvschaik what do you think?

src/platform/ios/ios.h
```diff
@@ -0,0 +1,12 @@
+#ifndef PLATFORM_IOS_H
+#define PLATFORM_IOS_H
+
+#ifdef __IPHONEOS__
+
+const char* ios_show_c3_path_dialog(int again);
+void c3_path_chosen(char *new_path);
+const char* ios_get_base_path(void);
+int ios_set_base_path(const char *path);
```
bvschaik, line 9: Be consistent with where you put the `*`. Convention in the project is to add a space between the type and the `*`, also for return values: (also check the other files)
```suggestion
const char *ios_show_c3_path_dialog(int again);
void c3_path_chosen(char *new_path);
const char *ios_get_base_path(void);
int ios_set_base_path(const char *path);
```

src/platform/ios/ios.m
```diff
@@ -0,0 +1,76 @@
...
+    GameDataPickerController *delegate = [[GameDataPickerController alloc] init];
+    UIDocumentPickerViewController *picker = [[UIDocumentPickerViewController alloc] initForOpeningContentTypes: @[UTTypeFolder]];
+    
+    picker.delegate = delegate;
+
+    NSError *error = nil;
+    picker.directoryURL = [[NSFileManager defaultManager] URLForDirectory:NSDocumentDirectory inDomain:NSUserDomainMask appropriateForURL:nil create:YES error:nil];
+
+    if (error) {
+        // Handle the error.
```
bvschaik, line 51: What should we do to handle the error?

src/platform/julius.c
```diff
@@ -438,6 +448,28 @@ static const char *ask_for_data_dir(int again)
...
+            SDL_arraysize(buttons), buttons, NULL
+        };
+        int result;
+        SDL_ShowMessageBox(&messageboxdata, &result);
+        if (!result) {
+            return NULL;
+        }
+    }
+    
+    printf("ask for data dir -- ios \n");
```
bvschaik, line 470: Remove debug print statement.

src/platform/julius.c
```diff
@@ -70,6 +71,11 @@ static void exit_with_status(int status)
     exit(status);
 }
 
+#ifdef __IPHONEOS__
+julius_args args;
```
bvschaik, line 75: Can be static?

src/platform/julius.c
```diff
@@ -356,6 +362,10 @@ static void teardown(void)
     platform_screen_destroy();
     SDL_Quit();
     teardown_logging();
+    
+#ifdef __IPHONEOS__
+    setup(&args);
```
bvschaik, line 367: Why do we need to run setup in `teardown()`?

src/platform/ios/GameDataPickerController.h
```diff
@@ -0,0 +1,12 @@
+#import <UIKit/UIKit.h>
+
+#import "SDL.h"
+#import "SDL_syswm.h"
```
bvschaik, line 4: Do we need these imports in the header?

src/platform/ios/GameDataPickerController.m
```diff
@@ -0,0 +1,47 @@
...
+    
+    [importDirectory stopAccessingSecurityScopedResource];
+    
+    char path[300];
+    [gameDataDirectory.path getFileSystemRepresentation:path maxLength:300];
+    c3_path_chosen(path);
+}
+
+- (void) documentPickerWasCancelled:(UIDocumentPickerViewController *)controller {
+
```
bvschaik, line 43: Do we need to do anything on cancel?

bvschaik (commented):
Looking good in general 👍 I mainly have some small questions due to not being familiar with the inner workings of iOS.

res/ios/Assets.xcassets/ColorMap.textureset/Universal.mipmapset/ColorMap.png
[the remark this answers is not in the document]
axmo: This was a holdover from the Xcode template SDL's iOS instructions suggested.  Turns out it was only used for a texture for the 3d spinning cube the template sets up.  I've now removed it.

res/ios/Assets.xcassets/AppIcon.appiconset/julius_1024.png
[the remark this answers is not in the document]
axmo: Xcode puts up a big warning if you don't have a 1024x1024 icon.

image

src/platform/ios/ios.m
axmo: `NSDocumentDirectory` should always exist and be writeable by the app. If there's an error here, something has gone very wrong with the user's iOS installation and I don't know that we can move forward at all. 

I guess we should alert that something has gone terribly wrong and wait for the user to force quit the app (iOS apps may not self-terminate)

src/platform/julius.c
axmo: iOS apps are not allowed to self-terminate.  App lifecycles are managed by the OS in response to user input and device resources available.

Since we can't quit ourselves, If the user picks Exit from the menu, they are left with a blank screen. Left nothing else to do, they will then tap the home button, which will suspend the app and bring the user back to the Home Screen, but the app is still 'open' and if the user taps on the Julius icon again, they will be brought back to the same blank screen. My quick solution to avoid this annoyance was to re-start the game in `teardown()`.

The right answer is probably to also remove the 'Exit' option from the menu on iOS.

axmo: Since listing in the App Store is out of the question, Apple can't really argue about it if we `exit(0)`, But it is definitely outside of the norm for iOS apps to do so.

src/platform/ios/GameDataPickerController.h
axmo: No, they are not.  (But `#import "SDL_platform.h"` *is* needed in `ios.h` now that I've removed SDL.h from here, for the `__IPHONEOS__` definition.)

src/platform/julius.c
crudelios: Can't the window even be minimised?

Android is more versatile, it closes the window and restarts cleanly, even though the app stays in memory!

src/platform/ios/GameDataPickerController.m
axmo: We should probably alert the user that we cannot continue without the game data folder and then re-open the game data picker window.

src/platform/julius.c
axmo: Lots of things are probably possible using undocumented API that would get you rejected from App Review... I've never done any not-for-app-store iOS apps before though, so I have not explored the options there.  `exit(0)` would *definitely* get an app rejected.

As for running minimized: Julius on iOS does currently support running in Split View and Slide-Over window modes, but once off-screen it is suspended.  Assuming the device has enough memory, when the app is brought back to the foreground again, it will keep running as though nothing had happened. If the user opened a memory-hungry app in the meantime, causing Julius to be terminated, it will launch to the title screen, and the user will lose any unsaved progress.

Ideally we would respond to the notification that we're about to be suspended by doing a suspend-specific autosave, and then on launch, checking to see if it exists and reloading it, skipping right past the menus, but that is a feature for some future update.

.ci_scripts/run_cmake.sh
```diff
@@ -10,6 +10,9 @@ case "$BUILD_TARGET" in
 "mac")
 	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DSYSTEM_LIBS=OFF -DCMAKE_OSX_ARCHITECTURES="x86_64;arm64" ..
 	;;
+"ios")
+	mkdir build && cd build && cmake cmake .. -DTARGET_PLATFORM=ios -G Xcode
```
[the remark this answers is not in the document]
axmo: This appears to have been a copy and paste error. Oddly enough it does work, but should not be there.

src/platform/ios/ios.h
axmo: done!

src/platform/ios/ios.m
axmo: Actually, never mind, there's no need to set the document picker directory on open. `error` here was meant to hold any file-copying error later in the chain, but that was moved to `GameDataPickerController.m` instead.  The document picker will default to the last location opened by the user or their user document directory.

crudelios:
The Visual Studio build is failing with this error:

```
D:\a\julius\julius\src\platform\ios\ios.h(4,1): error C2773: #import and #using available only in C++ compiler [D:\a\julius\julius\build\julius.vcxproj]
  (compiling source file '../src/platform/julius.c')
```

Maybe only include `ios.h` from `julius.c` if the target build is iOS? It's not ideal, but if that's the only way to prevent the issue, then so be it.

Or, instead of `#import`, is it possible to use `#include`?

src/platform/ios/GameDataPickerController.m
axmo: Alert and re-request game data folder added in recent commits

src/platform/ios/ios.m
axmo: indeed, it was left over from before I moved some logic out, and has now been removed.

axmo:
> Or, instead of `#import`, is it possible to use `#include`?

Ah, yes, it should indeed be `#include` (and now it is)

axmo:
Once again ready for consideration.

crudelios:
Ah nice, the builds are working on all platforms now.

Thinking a bit ahead, you're still planning to add support for augustus, right?

In that case, is it okay if, when this is merged to Julius, I then merge the latest Julius changes to augustus and you work from there?

bvschaik (approved):
Looks good to me, let's merge 👍

PR #744, opened 2025-01-15
Feature/condition fire ruins

UberWaffe:
Added event trigger for building caught fire
Added event trigger for building collapse poor maintenance

This allows for events to trigger their checks on events other than the start of a new game month.
Currently added trigger types include: Building catching fire, building being constructed, building deleted by player, etc.

Also adds the concept of 'context'.
For these new types of triggers, this is the type of building that is involved.

I.e. You can do thinkgs like fire off an event when the player places a specific type of building.
Or when a specific type of building (or even any type) catches fire.

Should support redoing the tutorials completely with events only.


UberWaffe:
Apologies, wrong base branch used as merge.
Meant to be for Augustus.
PEBKAC error.

Issue #749, opened 2025-03-05
Windowed Fullscreen


felipecrs:
Hi,

I was wondering, wouldn't it be nice to have Windowed Fullscreen support?

This way I can move my cursor across monitors to other applications without having to minimize the game.

bvschaik:
Constraining the mouse to the window (using `SDL_WindowGrab`) was a deliberate choice: https://github.com/bvschaik/julius/pull/51

felipecrs:
That's when the game is in _Exclusive Fullscreen_ mode. The same applies to most other games. I'm not asking to change that.

What I'm asking for is a _Windowed Fullscreen_ mode, where the window decorations are hidden, and the window size matches the resolution of the display.

Then, in _Windowed_ or _Windowed Fullscreen_ it would not make sense to constrain the cursor.

felipecrs:
Oh. _Windowed Fullscreen_ is also known as _Borderless Fullscreen_.

bvschaik:
We're using SDL's `SDL_WINDOW_FULLSCREEN_DESKTOP` option: https://github.com/bvschaik/julius/blob/db676ad52d8f9b26ba2db52d6273c9c4fd788828/src/platform/screen.c#L287

I'm not sure if that also counts as windowed fullscreen. SDL does not offer anything else besides true fullscreen at a specific resolution (causing desktop resizes), and this option. It does seem to have a [borderless option](https://wiki.libsdl.org/SDL2/SDL_SetWindowBordered) that may do what you ask.

But technical details aside: how would the user tell the difference between true fullscreen and fullscreen without borders? And how would she switch between modes? Or move the window from one screen to another?

felipecrs:
I just realized that Augustus has this "feature". All you need to do is to NOT have _Fullscreen_ ticked and then set _Windowed resolution_ to your display resolution:

![Image](https://github.com/user-attachments/assets/ac608d71-7cd5-4fd2-827e-2aab783a2f0e)

So, maybe the cleanest and simplest way to solve this would be to add more resolution options (including display's) in the _Video settings_:

![Image](https://github.com/user-attachments/assets/175756ea-73bc-467a-bcb6-7516e40276d5)

bvschaik:
I tried some things out yesterday and today. Findings:

First, a borderless fully screen-covering window is ridiculously tricky to implement correctly for all platforms, especially considering things like notches.

[...]

Issue #172, opened 2019-11-12
Windows 10 mouse scroll issue


dcnieho:
When trying to scroll rightward (or downward), it does not work (view doesn't change) when i place my cursor as far right (or down) as it can go. If i put it one pixel inward, scrolling happens. No problem with the left and upper screen edges.

I don't seem to have a log file to send you?

This is on fully updated Windows 10 Enterprise, using the nightly build from yesterday. Happen both on my laptop's built in screen, and on an external screen (both single monitor setups, so internal screen disabled when on external screen).

Let me also say that it is amazing to be able to play the game on my big monitor, super cool! How did you manage to achieve this, if i may ask? Partial decompile and make your own new graphics and sound back end? I guess there is large overlap in engine between these and the Pharaoh, Zeus and ROTMK games, so perhaps in some distant future we may get those to work too. I am not in a position to offer help with that now, but perhaps. Thanks for bringing this back to me, its impressive, really impressive how well it all works!

bvschaik:
Are you playing full screen or windowed? Since you're using display scaling, Windows might not consider the very last pixel as part of the game's window due to rounding. I'll have a look.
The log file `julius-log.txt` _should_ be in the same directory as your julius.exe (which works best if it's in the same directory as your C3 installation). If it's not there I will have a look at that as well.

crudelios:
@bvschaik long time no see! I've been away for a while and slowing inching to get back to julius, but the amount of work I have isn't helping...

Anyway, both this issue and #173 are due to DPI scaling.

I had both issues fixed in my DPI PR, but that, in turn, created many other issues so I removed the PR.

Still, the only solution I can think of is to take care of DPI manually, because SDL doesn't do that (at least it didn't in v2.0.9).

rsn8887:
Couldn't you just check for last or second-to last pixel, e.g. use a two-pixel active boundary?

dcnieho:
If you point me to the relevant bit in the code for windows, i'm happy to have a look and try out some possible fixes if i see any.

@bvschaik: I play fullscreen (so no window decorations) everything else is normal size, just the cursor is so small its hard to see.
@crudelios: in that sense, i guess changes to DPI awareness aren't necessary other than for the cursors? Or is there other things i might notice if i look carefully?

regarding log: I guess it is a permissions issue. Admin rights (or a UAC accept) are needed to create files in the program files folder. If i run the exe from my desktop, a log appears. here's its contents:

```
[...]

PR #750, opened 2025-03-12
Improve fullscreen mouse handling

bvschaik:
Based on investigation done for #749:

- Always grab the mouse in fullscreen mode, also when there's only 1 display, and also on macOS, which seems to work properly now
- Remove workaround for #172 due to always grabbing the mouse now: when the mouse cannot leave the window, it won't ever be in the 1-pixel limbo at the edge of the screen


crudelios:
The code looks good, I'll test on my surface and I'll let you know!

crudelios:
I just checked and unfortunately the scrolling workaround is still needed. Without it scrolling doesn't work if the mouse cursor is at exactly the right or bottom edge of the screen.

bvschaik:
Bugger. The scrolling workaround causes the screen to continue scrolling after the mouse leaves the window, something I want to fix.

crudelios:
Yeah that's also not good.

I guess we can file a bug report on SDL? I'm just not sure how long it would take to fix, especially now that SDL3 is a thing.

Either that or we only apply the workaround when there's only one monitor.

I'm not sure if it's possible to detect whether the new monitor is to the right or below the window, as it's only on those cases that we'd need to prevent scrolling.

bvschaik:
Indeed with SDL3 being stable they probably won't do much about SDL2 bugs anymore. Unless the bug is also in SDL3, perhaps.

New try: keeping the win10 fix in place and additionally tracking the window's focus. Could you give it a go?

crudelios (approved):
Working fine now!

PR #754, opened 2025-05-15
UI enhance

xxoommd:
1. Add 'return_to_main_menu' and 'Julius_config_options' on top menu bar;
2. Add 'play_pause_button' on sidebar extra panel;
3. Add 'inverse_dragging_map_with_right_mouse' config.


bvschaik:
General comments (I didn't look closely at the code yet):

> 1. Add 'return_to_main_menu' and 'Julius_config_options' on top menu bar;

Why? We have "New game" for this exact feature. The Julius config options might be nice but I'm afraid of possible side effects of allowing users to change settings mid-game.

> 2. Add 'play_pause_button' on sidebar extra panel;

Again, why? On devices with a keyboard this is a hotkey, on devices without keyboard we have a play/pause button in the top left corner. Also, it will not fit in all translations, especially "Resume" is a problem. Also also: if these are UI enhancement from Augustus (as mentioned in the other identical PR that you closed), why didn't you copy the translations for all languages?

> 3. Add 'inverse_dragging_map_with_right_mouse' config.

No one has asked for this before, dragging the map the other way feels unnatural?

xxoommd:
Sry for duplicated ft, ignore it.

In general, Augustus has smooth UI features, like middle mouse scrolling, roamers' path previewing, etc. I hope Julius can apply these handy features (However, it has no custom assets as Augustus does, may find other ways to achieve) and embrace all players with all kinds of different habits. For example, right mouse dragging, my friends and I prefer to drag the MAP rather than the CAMERA VIEW. Julius config in game could only show UI configs, hide 2 bug options, this won't affect gameplay.

In China thousands of players are playing C3 (some of them doing live stream on TikTok) everyday, but they can't access outside GFW, their needs and experiences cannot be shared with others.

This PR is not perfect, consider it as an idea. If it goes against your first intention, and you want to keep it original, I will undo this.

trackxx:
> > 3. Add 'inverse_dragging_map_with_right_mouse' config.
> 
> No one has asked for this before, dragging the map the other way feels unnatural?

This has been a feature in Augustus for a long time and is a very common way of navigating the map in other strategy games.
I use it exclusively, it's very natural, since in games where the mouse doesn't disappear during rightclick, the world moves, but mouse remains stationary in the gameworld, clearly 'dragging' the map.

bvschaik:
Sorry for the delayed response, I was on holiday.

> In general, Augustus has smooth UI features, like middle mouse scrolling, roamers' path previewing, etc. I hope Julius can apply these handy features

Path preview and zooming are both features that I consider not in line with the project goals, not to mention that they will take a lot of time to adapt from Augustus.

> right mouse dragging, my friends and I prefer to drag the MAP rather than the CAMERA VIEW.

I'm on board with this one. Since Augustus probably has translations for all of the languages rather than one Chinese only, I'll cherry-pick this change from Augustus myself if you don't mind.

PR #759, opened 2025-07-28
Allow players to continue playing 2 years after winning the second mission

HyperJeanJean:
The second mission in the campaign ("Brindisium") is the last tutorial mission and ends very quickly, players may not have had the time to explore all the game mechanics. This change should allow the players to experiment with the game a bit more before playing the "real" missions.

From my own testing, save compatibility does not seem to be impacted. I can still load and play saves in C3 after winning the mission, and it will correctly trigger the victory screen after two years.

image


bvschaik (approved):
Looks good to me. And thanks for testing thoroughly with original C3!

PR #762, opened 2025-08-30
Fix tooltips sometimes having a trailing new line

HyperJeanJean:
From my understanding, the algorithm that calculated the required width for those tooltips updated the `current_width` variable before checking if it has enough space to do so, leading to a `current_width` (and a `largest_width`) larger than the `box_width`. So tooltips that had a longer text than `box_width` had a chance to fit their text in a single line even though two lines should be required.

With that fix, `current_width` is updated after checking that there is enough space.

Before:
image<br/>
image

After:
image<br/>
image

As as side not, the original Caesar 3 tooltips do not have this issue (but they appear to be a bit more longer?) so I guess this is a Julius issue.

### DIFF
--- a/src/graphics/text.c
+++ b/src/graphics/text.c
@@ -389,15 +389,16 @@ int text_draw_multiline(const uint8_t *str, int x_offset, int y_offset, int box_
         }
         int current_width = 0;
         int line_index = 0;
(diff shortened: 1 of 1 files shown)


bvschaik (approved):
Good catch! I like that the measure and draw methods are now identical in structure again.

PR #764, opened 2025-08-30
Fix slightly misplaced button hitbox for build menu

HyperJeanJean:
There is a slight hitbox misplacement for the build menu buttons, it seems to be 8 pixels to the right. 

Caesar 3 has the same issue but the build menu in the editor is fine.


bvschaik (approved):
Nice one.

PR #767, opened 2025-08-30
Display the number of active wharfs in resource settings

HyperJeanJean:
This is a proposal to show the number of active wharfs to the resource settings of meat when the wharf is enabled. The other resources already show detailed information about the working and non-working industries but I could not find a way to retrieve all wharfs (including the non-working ones) so this is a quite limited addition. This means that if there are some non-active wharfs, they are completely ignored.

image

For now, a wharf is considered active when it has both some workforce and a boat.


bvschaik (approved):
Useful addition!

PR #768, opened 2025-08-30
Hide resources that cannot be traded in empire screen

HyperJeanJean:
The empire screen show the trade information based on the empire objects, but everything else in the game uses the city objects data, e.g. to determine if a resource can be traded or if it should be shown in the trade advisor. In almost all cases, this is perfectly fine as the city objects data is initialized from the empire objects data.

In Damascus though (and potentially other cases), there is an inconsistency between the two: the empire screen shows that Tarsus could export marble even though it actually can't because the city data says the opposite.

This change is a proposal to actually use the city data instead of the empire data in the empire screen. IMHO given that the empire data is only used to initialize the city data, it makes little to no sense to continue showing the empire data the whole game when the city data is available. This should ensure that the data shown by the empire screen is actually relevant to the player.

Before:
image<br/>
image

After:
image<br/>
image

As far as I can tell, this does not affect the gameplay but this is a change from the original Caesar 3 anyway so feel free to close it if does not fit into the scope of Julius.

### DIFF
--- a/src/window/empire.c
+++ b/src/window/empire.c
@@ -160,7 +160,7 @@ static void draw_trade_city_info(const empire_object *object, const empire_city 
         lang_text_draw(47, 10, x_offset + 44, y_offset + 40, FONT_NORMAL_GREEN);
         int index = 0;
         for (int resource = RESOURCE_MIN; resource < RESOURCE_MAX; resource++) {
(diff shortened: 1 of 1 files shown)


QuakeIV:
I'm in favor of this personally

bvschaik (approved):

Issue #789, opened 2026-05-04
Crash when clicking on a city to open a trade route


axelfo1ey:
I'm playing julius-1.8.0.21, at 2nd mission Clerk-Brundisium.
When opening the empire map and clicking on another city, the game crashes to desktop.
Tried loading other map, and the error also occurs. 
It also happens with 1.8.0.22-016d5254, but it's fine in 1.8.0.

Log says:
`ERROR: Oops, crashed with signal 11 :(`

[save.zip](https://github.com/user-attachments/files/27328274/save.zip)

crudelios:
It's not crashing for me, at least on android.

This most likely means there's an issue with your original C3 installation.

Which language are you using?

Also,  how did you install the original C3?

axelfo1ey:
@crudelios , here it only crashes on dev-releases. 
It works fine on android, but I'm also using the stable release.

I'm using the default language, and I installed C3 from GOG.

brunocabete:
Happened to me as well and is reproducible.
I'm on linux and using the appimage build version 1.8.0.22-016d5254

I loaded [this save](https://caesar3.heavengames.com/downloads/showfile.php?fileid=14) and just clicked on a red city on the world map. it crashes with the error:
```
ERROR: Oops, crashed with signal 11 :(
INFO:   ./julius.appimage(+0x1e422) [0x556898cff422]
INFO:   ./julius.appimage(+0x1bae1) [0x556898cfcae1]
INFO:   /lib64/libc.so.6(+0x1a0b0) [0x7f85406b40b0]
INFO:   ./julius.appimage(+0xcef6f) [0x556898daff6f]
INFO:   ./julius.appimage(+0xa37e2) [0x556898d847e2]
INFO:   ./julius.appimage(+0xa4b88) [0x556898d85b88]
INFO:   ./julius.appimage(+0x77c5f) [0x556898d58c5f]
INFO:   ./julius.appimage(main+0x430) [0x556898cfb930]
INFO:   /lib64/libc.so.6(+0x3681) [0x7f854069d681]
INFO:   /lib64/libc.so.6(__libc_start_main+0x88) [0x7f854069d798]
INFO:   ./julius.appimage(+0x1aee5) [0x556898cfbee5]
```

I'm playing the game in Portuguese.

Edit: Downloaded the game in english from GOG, extracted with innoextract and pointed julius to it.
Same error

crudelios:
Fixed, thanks @brunocabete!

PR #791, opened 2026-06-22
Fix issue #789: game crashes when clicking on a city on empire map

brunocabete:
it looks like a bug that only crashes depending on the build/platform/compiler/whatever, as it is a simple array out of bounds error. It crashes for some people and grabs random garbage from OOB to others


crudelios:
Good catch!

bvschaik:
Wow, nice catch!

Issue #775, opened 2025-12-03
Houses or tiles seem to be linked


zwentendorf:
First of all thank you for all of your work! I'm so happy that I can play Caesar III on Linux.

I think I have two houses (or tiles on the map?) that seem to be linked.

When I open the building info of one of my large tents, I get a info declaring it as a medium villa:

<img width="845" height="543" alt="Image" src="https://github.com/user-attachments/assets/35e81839-685e-48a3-8133-cfd613b9cf84" />

<img width="811" height="557" alt="Image" src="https://github.com/user-attachments/assets/5f198ab5-2fab-4eb9-804e-817519b8cebd" />

The only actual medium villa on my map shows the same building info:

<img width="632" height="428" alt="Image" src="https://github.com/user-attachments/assets/161a44a0-3d0c-4b37-929b-eb14489ebf14" />

<img width="811" height="557" alt="Image" src="https://github.com/user-attachments/assets/a4ab396a-02c8-46af-948a-c20bd46f85c4" />

The medium villa shouldn't have evolved to a medium villa, because it doesn't have access to a hospital. The health/hospital coverage overlay says it does have access, although all its neighbours don't:

<img width="1913" height="743" alt="Image" src="https://github.com/user-attachments/assets/33ba7d48-819e-470b-89bd-74c651228c24" />

I guess that's because the medium villa is linked to the large tent in the first picture (that clearly has access to a hospital).

Bulldozing the large tent will actually destroy my medium villa (and leave the large tent allone). Bulldozing either of them (or both) doesn't fix the issue – I can build a new vacant lot at the place my medium villa is currently standing and it will once again gain hospital access and evolve to a medium villa.

Here's my save file:

[savefile.zip](https://github.com/user-attachments/files/23913566/savefile.zip)

bvschaik:
Thanks for the detailed report, this is an interesting issue. I looked into it and they are indeed "linked". How it works: the game keeps a list of buildings with all their properties, and for convenience it also has a 2d grid indicating which building ID is at which x/y location.

This medium villa is on the 2d grid twice: once in its normal location, and once in the large tent location. At the position where the large tent is, there is no building in the building list. Once they got linked, the game state became corrupt and deleting the houses doesn't work because of the duplication in the 2d grid.

Some questions:
- Do you have any idea how this could have happened?
- Which version of Julius are you running? 1.8.0?
- Where did you get Julius from? Flatpak, appimage, distro repository, built yourself?

[...]

PR #795, opened 2026-09-04
Fix undo corruption when merged house becomes vacant

csfreitas:
Hi Bianca, one thing I should have made clearer: this builds on [Crudelios's earlier fix](https://github.com/bvschaik/julius/commit/d8ba882a5ff53ac237f9f551c7c069c6ac053c0c), which you linked in #775. It adds the same Undo protection for the case where a merged 2x2 house becomes four vacant lots.

You were right about the ID protection. The problem I reproduced was Undo bringing back the old house's map data after it had split. Later deletion and ID reuse then linked the old tile to a distant building.

I've removed the artificial test and corrected the description and code comment. The [steps and saves are in my previous comment](https://github.com/bvschaik/julius/pull/795#issuecomment-5577528792).

The fix prevented this sequence in my master test, but I can't confirm it covers every case of #775 or the reporter's exact sequence. It won't repair an already corrupted save.

### DIFF
--- a/src/building/house.c
+++ b/src/building/house.c
@@ -77,11 +77,14 @@ static void create_vacant_lot(int x, int y, int image_id)
 
 void building_house_change_to_vacant_lot(building *house)
 {
+    if (house->house_is_merged) {
+        // New vacant lots may reuse building IDs still referenced by undo.
+        game_undo_disable();
+    }
     house->type = BUILDING_HOUSE_VACANT_LOT;
     house->subtype.house_level = house->type - BUILDING_HOUSE_VACANT_LOT;
     int image_id = image_group(GROUP_BUILDING_HOUSE_VACANT_LOT);
     if (house->house_is_merged) {
-        game_undo_disable();
         map_building_tiles_remove(house->id, house->x, house->y);
         house->house_is_merged = 0;
         house->size = house->house_size = 1;
--- a/src/building/house.c
+++ b/src/building/house.c
@@ -77,14 +77,12 @@ static void create_vacant_lot(int x, int y, int image_id)
 
 void building_house_change_to_vacant_lot(building *house)
 {
-    if (house->house_is_merged) {
-        // New vacant lots may reuse building IDs still referenced by undo.
-        game_undo_disable();
-    }
     house->type = BUILDING_HOUSE_VACANT_LOT;
     house->subtype.house_level = house->type - BUILDING_HOUSE_VACANT_LOT;
     int image_id = image_group(GROUP_BUILDING_HOUSE_VACANT_LOT);
     if (house->house_is_merged) {
+        // New vacant lots may reuse building IDs still referenced by undo.
+        game_undo_disable();
         map_building_tiles_remove(house->id, house->x, house->y);
         house->house_is_merged = 0;
         house->size = house->house_size = 1;
--- a/src/building/house.c
+++ b/src/building/house.c
@@ -81,6 +81,7 @@ void building_house_change_to_vacant_lot(building *house)
     house->subtype.house_level = house->type - BUILDING_HOUSE_VACANT_LOT;
     int image_id = image_group(GROUP_BUILDING_HOUSE_VACANT_LOT);
     if (house->house_is_merged) {
+        game_undo_disable();
         map_building_tiles_remove(house->id, house->x, house->y);
         house->house_is_merged = 0;
         house->size = house->house_size = 1;
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -127,8 +127,8 @@ add_integration_test(sav_valentia1 valentia57.sav valentia57-after.sav 1026)
 add_integration_test(sav_lugdunum1 brugle-lugdunum.sav brugle-lugdunum-after.sav 1176)
 
 add_integration_test(sav_native1 brugle-lugdunum-native.sav brugle-lugdunum-native-after.sav 1678)
-add_test(NAME undo_vacant_house_split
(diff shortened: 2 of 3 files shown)


crudelios:
Good catch! I did disable undo on house evolution merge/split, but missed this one.

bvschaik:
Could you detail the steps required to reproduce the bug so I can check whether the fix works?

I tried to reproduce but I could not. I built some statues (to be deleted later), then some houses. I caused the people to emigrate with high taxes and 0 wages. I deleted the statues just before a 2x2 merged house got vacated. The 3 vacant lots chose building ids outside the deleted statue range due to this bit of code in src/building/building.c:

https://github.com/bvschaik/julius/blob/3c7d865d8a509a7e7714d2a17971cf2c85a139df/src/building/building.c#L54-L62

The code for creating a building explicitly checks the undo list, so they are not reused.

Also, please don't bother with writing a test for this - the tests are only for integration testing the game logic, not these kind of bugs.

csfreitas:
Thanks for checking this, Bianca. You are right about the building-ID protection, and I need to correct my earlier explanation.

My original diagnostic forced the house to become vacant between `game_undo_start_build()` and `game_undo_finish_build()`. During that interval, undo is not ready, so `game_undo_contains_building()` returns false. However, normal gameplay does not advance the simulation while `building_construction_in_progress()` is true. The diagnostic bypassed that restriction, so it was not a valid gameplay reproduction.

I then ran a separate local, headless diagnostic without this PR's fix, using the normal construction APIs and `game_run()`, with maximum taxes and zero wages. Across three existing saves, I observed 38 merged-to-vacant transitions. Four occurred while undo was available. In all four, the new lots avoided the protected statue ID, and undo restored the statue without duplicate map references to that ID. This was not a manual GUI reproduction and does not rule out other causes of #775.

The test in this PR only checks that undo gets disabled; it does not demonstrate the reported corruption. Understood about the integration suite: I will not pursue adding this diagnostic there.

I have moved this PR to draft. I do not currently have valid reproduction steps for the proposed cause, and the existing fix should not be treated as established. Sorry for presenting the earlier result too confidently, and thank you for catching this. The next step is to identify a sequence reachable through normal gameplay before proposing a correction.

csfreitas:
Hi Bianca,

You were right about `building_create()` checking the undo list. It isn't reusing IDs while they're protected by Undo, as I originally thought.

This took me quite a while to pin down. I kept hitting cases where Undo was correctly disabled, so demolishing something and waiting for a house to downgrade wasn't enough. The case I eventually reproduced was a merged 2×2 house becoming four vacant lots while Undo was still available.

I prepared a test save just before that happens. I reproduced the problem on Julius 1.7 and master (`5992838`), then repeated it on the same master with the proposed guard. The master comparison was on Windows x64 with SDL 2.32.8.

The sequence is:

1. Load `01-start.sav` and pause immediately. Demolish the distant small villa, ID `37`, at `(70,38)`.
2. Resume until merged house `41` at `(71,88)` becomes vacant, then pause and use Undo.
3. Demolish the adjacent theater and lots `41` and `476`. Let those removals finish, then pause again.
4. Place a medium statue at `(70,88)`, demolish lot `477` at `(72,89)`, and place a vacant house at `(59,71)`, all while paused.
5. Resume without further construction or Undo. In this run, the corruption was visible by May.

The important part happens at step 2. Undo restores the villa, but it also restores the old merged-house image and tile metadata. The house itself isn't merged again: its four separate vacant-lot records remain.

The distant link appears later, after deletion and normal ID reuse. ID `477` became a burning ruin at `(32,7)`, but the old tile at `(72,89)` still referenced it. Clicking the old location opened information about the distant ruin. The saved file confirms that both locations reference the same ID.

With the guard, Undo becomes unavailable as soon as the merged house becomes vacant. The villa isn't restored, and neither is the stale house-map state. After continuing the remaining steps, `477` was reused at the same ruin location, but this time the old tile was clear and there was only one map reference to that ID.

So the change prevents an Undo operation that is no longer safe; it doesn't repair a save that's already corrupted.

The ZIP includes the starting save, the before/after Undo checkpoints, the corrupted master save, and the fixed-run comparison. The README has the remaining coordinates and checkpoint details. Start from `01` to repeat the sequence: the intermediate saves are for inspection, since loading them doesn't restore the live Undo history.

This reproduces a vacant-lot/burning-ruin link rather than the reporter's exact villa/tent case. The guard prevents the corruption in this sequence, but I can't say it covers every possible cause of #775. Without a save from before the reporter's corruption or the exact steps that led to it, I can't confirm that this is the same path they hit.

[julius-775-review-files-20260907.zip](https://github.com/user-attachments/files/31932020/julius-775-review-files-20260907.zip)

bvschaik:
Thanks. I tried your reproduction steps and was able to reproduce. I still don't fully understand how exactly it corrupts the data - I tried to reproduce on my own emigration save but failed. However, I think the fix is a good one to have anyway, since it prevents an inconsistent state.

Could you mark the PR as ready so I can merge it? I think we should close the linked issue, since it's highly likely that this inconsistent state lead to the linked houses: all the other cases of houses merging/splitting are covered by disabling undo.

csfreitas:
Thanks for taking the time to reproduce it! Apparently the planets have to align just right to corrupt this save. I spent quite a while getting Rome into exactly the wrong state 😄

The key seems to be that Undo restores the old map state, but leaves the four separate vacant-lot records in place. Later demolition and ID reuse make that mismatch visible as a link to a distant building. That would explain why simply making a house lose its residents usually isn't enough.

I've marked the PR as ready. Thanks again!

bvschaik:
Thanks for figuring out how to reproduce in the first place!

PR #796, opened 2026-09-04
Fix Emscripten 6 linking

csfreitas:
## Summary

Remove the unused `--bind` option from the Emscripten linker flags.

## Background

The Emscripten workflow installs `emsdk latest`. The latest SDK now resolves to Emscripten 6.0.9, where linking the Julius C target with `--bind` pulls in the Embind C++ RTTI library and fails with undefined `typeinfo` and `vtable` symbols.

Julius does not contain any `EMSCRIPTEN_BINDINGS`, include `emscripten/bind.h`, or otherwise use Embind. The option has therefore been unnecessary since the initial web build and adding `-sDEFAULT_TO_CXX` would only link an unused C++ runtime.

`--bind` is also deprecated by Emscripten in favor of `-lembind` for projects that actually use Embind.

## Validation

The Emscripten job builds successfully with Emscripten 6.0.9 after removing the flag:

https://github.com/csfreitas/augustus/actions/runs/33912603431

This change only affects the Emscripten linker command.


bvschaik:
Thanks for the fix.

